PR #30, opened 2020-10-17
Fix snakeyaml version

finos-admin:
Running `mvn dependency:tree` I see snakeyaml 1.26 , so I expect it fixes the issue.

@gs-bracej - is this a valid solution, although we haven't quite figured out why this version conflict only happens on this module and not on others?

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -337,6 +337,12 @@
 
     <dependencyManagement>
         <dependencies>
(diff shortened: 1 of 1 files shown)


gs-bracej (approved):

maoo (approved):
LGTM!

PR #152, opened 2021-03-10
cleanups in extraProtocolToClassiferPath

MauricioUyaguari:


legend-engine-protocol-pure/src/main/java/org/finos/legend/engine/protocol/pure/v1/extension/PureProtocolExtension.java
```diff
@@ -1,4 +1,4 @@
-// Copyright 2020 Goldman Sachs
+// Copyright 2021 Goldman Sachs
```
kevin-m-knight-gs, line 1: This should probably stay 2020, since the file was added in 2020.

PR #182, opened 2021-04-09
fix the bug for date propagation and inheritance

YannanGao-gs:
fix the bug for date propagation and inheritance


davidharte-gs (changes requested):
As discussed, please investigate why the owning milestoned class doesn't appear to have the correct number of qualified properties, this code most likely needs to be updated to handle processing of superTypes.

PR #197, opened 2021-04-17
Fix type inference for fold

rajamony:


pierredebelen (changes requested):
Could we please have a test in the compiler section instead of Query?

rajamony:
Done. Test is simpler and now entirely in grammar

legend-engine-language-pure-compiler/src/main/java/org/finos/legend/engine/language/pure/compiler/toPureGraph/handlers/Handlers.java
```diff
@@ -96,6 +96,21 @@ private static void updateTwoParamsLambda(Object lambda, GenericType newGenericT
...
+    {
+        if (lambda instanceof Lambda)
+        {
+            Variable variable = ((Lambda) lambda).parameters.get(0);
+            variable._class = PackageableElement.getUserPathForPackageableElement(newGenericType._rawType());
+            variable.multiplicity = m;
+
+            Variable variable2 = ((Lambda) lambda).parameters.get(1);
+            variable2._class = PackageableElement.getUserPathForPackageableElement(newGenericType2._rawType());
+            variable2.multiplicity = m;
```
pierredebelen, line 109: The multiplicity of the lambda parameters should not be equal. The iterator will most likely be 1 but the accumulator should be equal to the one of the Lambda return.

pierredebelen (changes requested):
Thanks for fixing the tests! Please look at the comment for the lambda parameters multiplicity inference.

legend-engine-language-pure-compiler/src/main/java/org/finos/legend/engine/language/pure/compiler/toPureGraph/handlers/Handlers.java
rajamony: Have to keep in mind it is type AND multiplicity! The only two-parameter lambda we now have with different types is fold, for which iterator is mult-1. Conversion from org.finos.legend.pure.m3.coreinstance.meta.pure.metamodel.multiplicity.Multiplicity to org.finos.legend.engine.protocol.pure.v1.model.packageableElement.domain.Multiplicity is inline. Not sure if there is a better way to do it. Please review.

pierredebelen: Looks great now! Thanks for the update!

pierredebelen (approved):

PR #200, opened 2021-04-21
Add support for storeSubPath in included mapping

emilia-sokol-gs:
When compiling included mapping, substitute store "[dbInc->db]" syntax was not parsed to graph

### DIFF
--- a/legend-engine-language-pure-grammar/src/test/java/org/finos/legend/engine/language/pure/grammar/test/roundtrip/TestMappingGrammarRoundtrip.java
+++ b/legend-engine-language-pure-grammar/src/test/java/org/finos/legend/engine/language/pure/grammar/test/roundtrip/TestMappingGrammarRoundtrip.java
@@ -442,4 +442,28 @@ public class TestMappingGrammarRoundtrip extends TestGrammarRoundtrip.TestGramma
                 "  }\n" +
                 ")\n");
     }
+
+    @Test
+    public void testMappingWithIncludedMapping()
+    {
+        test("###Mapping\n" +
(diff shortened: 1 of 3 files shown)


legend-engine-language-pure-grammar/src/test/java/org/finos/legend/engine/language/pure/grammar/test/roundtrip/TestMappingGrammarRoundtrip.java
```diff
@@ -442,4 +442,28 @@ public void testAggregationAware()
                 "  }\n" +
                 ")\n");
     }
+
+    @Test
+    public void testMappingWithIncludedMapping()
+    {
+        test("###Mapping\n" +
+                "Mapping test::decomposeMapping\n" +
+                "(\n" +
```
pierredebelen, line 451: For this example to make sense, shouldn't the source mapping refer to dbInc? And shouldn't the 'target' mapping have a reference to db?

pierredebelen (changes requested):

legend-engine-language-pure-grammar/src/test/java/org/finos/legend/engine/language/pure/grammar/test/roundtrip/TestMappingGrammarRoundtrip.java
emilia-sokol-gs: you're right! Thanks for pointing that out, I updated and moved this test to TestRelationalGrammarRoundtrip

pierredebelen (approved):

PR #221, opened 2021-05-06
make relational grammar parser more consistent with protocol

akphi:
For `PropertyPointer` instead of outputting `null` for `class` when the property is of embedded property mapping, we will output `empty string`. This behaviour will make this:
- comply with the protocol
- in sync with Studio
- compiler more resilient


akphi:
Closing this as after discussing with @pierredebelen, we might eventually decide to change the Pure protocol instead. Also setting the unset field to `empty string` seems like a code-smell.

PR #229, opened 2021-05-14
Fix relational test with Pure Client Test Suite

emilia-sokol-gs:


pierredebelen:
Makes sense, but we need to make sure we start executing the 'grammar based test' in Github

PR #237, opened 2021-05-25
Ensure properties are always added to RelationalDBConnection

aziemchawdhary-gs:
If HelperRelationalBuilder.resolveDatabase fails, we still want to add
properties to the connection - common use case is if the connection is using a
place holder element for which database resolution will fail but we still want
to construct a connection


legend-engine-language-pure-store-relational/src/main/java/org/finos/legend/engine/language/pure/compiler/toPureGraph/HelperRelationalDatabaseConnectionBuilder.java
```diff
@@ -50,13 +50,16 @@ public static void addDatabaseConnectionProperties(org.finos.legend.pure.m3.core
...
         {
             connection._element(HelperRelationalBuilder.resolveDatabase(element, elementSourceInformation, context));
-            connection._timeZone(timeZone);
-            connection._quoteIdentifiers(quoteIdentifiers);
         }
         catch (RuntimeException e)
         {
             connection._element(new Root_meta_relational_metamodel_Database_Impl(element)._name(element));
         }
+        finally
```
akphi, line 58: I guess this is fine too, but why don't we just move all of these to before the `try ... catch` block instead of introducing this `finally` block?

pierredebelen: indeed!

aziemchawdhary-gs: yes this makes sense

PR #249, opened 2021-05-31
Add graphFetch cross key cache tests

gs-ssh16:


pierredebelen (commented):
Shouldn't we have the test in ASCII instead of JSON?

gs-ssh16:
Hi @pierredebelen, I added the test in _legend-engine-executionPlan-execution-store-relational_ module which does not have access to parser and plan generator. Should I move it to the server module (doesn't seem to be the correct place)?

pierredebelen:
As much as we can't have dependency to the parser & compiler in Execution's runtime & compile scope, I think it's ok to have dependencies to (legend-engine-language-pure-*) in the test scope as it will make the tests far more readable. I did the same in legend-engine-test-runner-mapping

gs-ssh16:
Added the tests in ASCII

PR #311, opened 2021-07-12
Integrating relational model with Delta Lake

aamend:
- adding PURE model for delta
- create pure2sql DSL for delta lake
- create authentication and datasource strategy
- adding doc


aamend:
Ignore pull request, I will submit a new one with CLA signed.

PR #322, opened 2021-07-22
improving performance of template functions

gs-gunjan:


gs-gunjan (commented):
This would go green once there is legend-pure release

PR #341, opened 2021-08-07
Use new MongoClient and add tests for Query APIs

akphi:
Continuation of #340 and #333 

### Fixes and improvements for Query APIs:
- [x] Add tests
- [x] Use `mongodb.client.MongoClient` instead of the `mongodb.MongoClient` so we can instantiate the connection using connection string
- [x] Remove `projectId` in `Query`
- [x] Rename  `QueryAPI` to `ApplicationQuery`

### Other small fixes:
- [x] Fix the way `Vault` retrieve the value, it should filter out `nullish` values and return the last value so we can do overriding of vaults


legend-engine-application-query/src/main/java/org/finos/legend/engine/application/query/api/QueryStoreManager.java
```diff
@@ -0,0 +1,169 @@
...
+    public Query updateQuery(String queryId, Query query, String currentUser) throws JsonProcessingException
+    {
+        List<Query> matchingQueries = LazyIterate.collect(this.getQueryCollection().find(Filters.eq("id", queryId)), QueryStoreManager::documentToQuery).toList();
+        if (!queryId.equals(query.id))
+        {
+            throw new ApplicationQueryException("Updating query ID is not supported", Response.Status.BAD_REQUEST);
+        }
+        if (matchingQueries.size() > 1)
+        {
+            throw new IllegalStateException("Found multiple query with ID '" + queryId + "'");
```
davidharte-gs, line 130: 'multiple query' -> 'multiple queries' :)

akphi: Great catch! Fixed!

akphi:
The fix for grammar roundtrip flakiness has been moved out to https://github.com/finos/legend-engine/pull/346

PR #345, opened 2021-08-18
M2M: Test with large decimal values

aziemchawdhary-gs:
Add test for decimal value fix in legend-pure (https://github.com/finos/legend-pure/pull/299)
Test needs to be in legend-engine, as the legend-pure JSON deserializer cannot handle the decimal value used in this test.

Note: this test will fail until [this](https://github.com/finos/legend-pure/pull/299) is merged and released in legend-pure


dave-wathen (approved):
Good test

PR #346, opened 2021-08-18
Make Grammar composer print elements in order when there is no section index

akphi:
- [x] Make sure grammar composer respect the order of elements in `PureModelContextData` when there is no `SectionIndex`

After doing some investigation, the flakiness is caused by us passing a `Set` to `getExtraFreeSectionComposers` in `PureGrammarComposerExtension`. The fix is to make it a `List`


akphi:
@davidharte-gs let's hold this off till I confirm again with @MauricioUyaguari

PR #350, opened 2021-08-23
Support at/indexOf function in the execution plan function library

YannanGao-gs:
Support at/indexOf function in the execution plan function library


YannanGao-gs:
The code review is approved by Dave Wathen. @dave-wathen

PR #356, opened 2021-09-01
Make AbstractServicePlanExecutor implement ServiceRunner and remove async execution

gs-ssh16:
Changes to accommodate suggestions in this SDLC PR:

- https://github.com/finos/legend-sdlc/pull/259#discussion_r699480684
- https://github.com/finos/legend-sdlc/pull/259#discussion_r699599628

### DIFF
--- a/legend-engine-language-pure-dsl-service-execution/src/main/java/org/finos/legend/engine/language/pure/dsl/service/execution/ServiceRunner.java
+++ b/legend-engine-language-pure-dsl-service-execution/src/main/java/org/finos/legend/engine/language/pure/dsl/service/execution/ServiceRunner.java
@@ -14,15 +14,11 @@
 
 package org.finos.legend.engine.language.pure.dsl.service.execution;
 
-import org.apache.commons.io.IOUtils;
 import org.finos.legend.engine.plan.execution.PlanExecutorInfo;
 import org.finos.legend.engine.plan.execution.cache.graphFetch.GraphFetchCrossAssociationKeys;
 
-import java.io.IOException;
+import java.io.ByteArrayOutputStream;
 import java.io.OutputStream;
-import java.io.PipedInputStream;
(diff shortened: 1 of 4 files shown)


legend-engine-language-pure-dsl-service-execution/src/main/java/org/finos/legend/engine/language/pure/dsl/service/execution/ServiceRunner.java
```diff
@@ -62,21 +58,13 @@ default String getCanonicalClassName()
...
-            PipedInputStream inputStream = new PipedInputStream(outputStream);
-            return IOUtils.toString(inputStream, Charset.defaultCharset());
-        }
-        catch (IOException e)
-        {
-            throw new RuntimeException(e);
-        }
+        ByteArrayOutputStream outputStream = new ByteArrayOutputStream();
+        this.run(serviceRunnerInput, outputStream);
+        return outputStream.toString();
```
kevin-m-knight-gs, line 63: We should specify the charset when converting the bytes to a string. Using the default charset is always risky.

kevin-m-knight-gs (changes requested):
This looks like a good improvement. The only comment I have is about specifying the charset when converting from bytes to a string.

legend-engine-language-pure-dsl-service-execution/src/main/java/org/finos/legend/engine/language/pure/dsl/service/execution/ServiceRunner.java
gs-ssh16: Currently serialization is being done using the default charset and hence I am using the same conventions here. But what you said is a valid point. Definitely going forward, we need to agree on a standard charset and use it all places.

kevin-m-knight-gs (approved):

PR #368, opened 2021-09-13
Prevent generating valid json when error occurs while streaming results

rafaelbey:
Jackson JsonGenerator has a feature to auto close JSON arrays and object when the generator gets closed.  Since this feature is enable by default, when there is an exception during streaming, the  try-with-resource will close the generator, and the  JSON arrays/objects with it.  

This means that even when we encounter an exception while streaming, it will generate proper JSON that will have incomplete data (given the streaming failure).

This feature also could hide developer bugs since it could hide places the developer forgot to properly close JSON arrays and objects.


davidharte-gs (commented):
- were you able to confirm that there is no/quantify the risk with the general ObjectMapper/withStandardConfiguration changes?
- good time to add a UserTest class in Engine, first test case be for this use case?

rafaelbey:
> * were you able to confirm that there is no/quantify the risk with the general ObjectMapper/withStandardConfiguration changes?

Yes.  The _ExecutionResultObjectMapperFactory_ ObjectMapper is only use on 5 serializers.  None of them reference a JsonGenerator directly, hence the fix should have no negative impact.  Even further, given how the serializers leverage the ObjectMapper, any error while streaming will end on malformed output already.

  
> * good time to add a UserTest class in Engine, first test case be for this use case?

I look at this, and that test leverage features that are not on the open source project.  We should discuss what other options we might have to have similar coverage.

rafaelbey:
@davidharte-gs - discussed with the team, and the docket image checks were failing given some new vulnerabilities, and that is not related to my change.  This should be fixed by now.

davidharte-gs (approved):
Looks good, thanks.

PR #378, opened 2021-09-21
Fix lambda with biTemporal class in property expression

emilia-sokol-gs:
Without this change the second %latest parameter will be lost, which is causing execution to fail


davidharte-gs (approved):
Looks good, thanks.

PR #380, opened 2021-09-21
Add maven coordinates to AlloySDLC

Yasirmod17:
Add maven coordinates to AlloySDLC

### DIFF
--- a/legend-engine-protocol-pure/src/main/java/org/finos/legend/engine/protocol/pure/v1/model/context/AlloySDLC.java
+++ b/legend-engine-protocol-pure/src/main/java/org/finos/legend/engine/protocol/pure/v1/model/context/AlloySDLC.java
@@ -20,8 +20,14 @@ import java.util.Objects;
 
 public class AlloySDLC extends SDLC
 {
-    @JsonProperty(value = "project", required = true)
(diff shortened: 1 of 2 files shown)


legend-engine-protocol-pure/src/main/java/org/finos/legend/engine/protocol/pure/v1/model/context/AlloySDLC.java
```diff
@@ -20,9 +20,15 @@
 
 public class AlloySDLC extends SDLC
 {
-    @JsonProperty(value = "project", required = true)
+    @JsonProperty(value = "project")
```
pierredebelen, line 23: Could you please deprecate the project variable?

pierredebelen (changes requested):

legend-engine-protocol-pure/src/main/java/org/finos/legend/engine/protocol/pure/v1/model/context/AlloySDLC.java
Yasirmod17: Done!

pierredebelen (approved):

PR #382, opened 2021-09-27
Fix a bug related to calling meta::pure::new with only one single parameter

YannanGao-gs:
Fix a bug related to calling meta::pure::new with only one single parameter


YannanGao-gs:
Code review was approved by Adam Fine.

PR #384, opened 2021-09-27
Engine supports temp table creation for snowflake at runtime

janeenyamak1:
Engine supports temp table creation for snowflake at runtime


davidharte-gs (changes requested):
As discussed, please let's not hardcode databases/schemas, let's first identify the minimum set of info required to get creation of the temporary table and the put command working. Thanks!

PR #385, opened 2021-09-28
Service Tag Changes

aayush81198:
Service tag grammar support for studio added and relevant roundtrips added


gs-ssh16 (approved):
Looks good

pierredebelen:
Why not leveraging the usual extension mechanism with Stereotypes & Tagged Values (and even more importantly Profile).

pierredebelen (changes requested):
Why not leveraging the usual extension mechanism with Stereotypes & Tagged Values (and even more importantly Profile).

PR #387, opened 2021-09-29
Fix the relational grammar parser to support joinOperation in mapping

YannanGao-gs:
Fix the relational grammar parser to support joinOperation in mapping


akphi:
@YannanGao-gs could you ask @beekemarie  to add you to `legend-contributors`?

PR #425, opened 2021-10-28
Bug fix: Handle CR character in relational mapping tester grammar composer

akphi:
See https://github.com/finos/legend-studio/issues/608


davidharte-gs (approved):
Approved

PR #426, opened 2021-10-28
Make AlloySDLCLoader to account for more generic metadata server URLs

akphi:
This is to make `AlloySDLCLoader` work with https://github.com/finos/legend-depot as right now the metadata server calls are hardcode with prefix `/metadata/api`


davidharte-gs (changes requested):
No tests?

akphi:
> No tests?

Yeah, this change has more to do with us modifying the config during deployment. I don't think we have an existing test for this anyway.

PR #428, opened 2021-10-28
Support Integers as Enum Values

YannanGao-gs:
Support Integers as Enum Values
resolves finos/legend-studio#563


YannanGao-gs:
link this issue with https://github.com/finos/legend-studio/issues/563

YannanGao-gs:
resolves  finos/legend-studio#563

aziemchawdhary-gs:
We should make the corresponding change in legend-pure as well I think. 

If we don't, we end up in a situation where a user has defined a model using this syntax, but in Pure IDE Light we can't use the same syntax as it isn't recognized. 
This causes a huge pain when debugging things like execution plan generation. 

Ideally we should be able to take model definitions from Studio and paste them into Pure IDE Light and be able to generate execution plans in interpreted mode for queries on top of this model. Without making the corresponding change in legend-pure, we then have to modify the users model syntax to be valid in Pure IDE Light.

aziemchawdhary-gs:
Can we also add tests to test the use of the enum and not just the definitions of enum with integer values?

YannanGao-gs:
close this pr for now

YannanGao-gs:
Close this ticket due to the reasons below.

Able to create an enum whose values are Integer, but this kind of enum can not be referenced. e.g. myEnum.1
When parsing the code myEnum.1, .1 will be treated as a Float since Float/Decimal has a higher priority than our customed token. Therefore, there will be an 'unexpected token' error when referring to this kind of enum.
Able to create an enum whose values are single-quoted Integer, but single quotes will be removed during the roundtrip.
Enum model::Something {'1'} . -> Enum model::Something {1}
There is a special logic to get rid of the single quote as long as it matches a regex pattern.

PR #432, opened 2021-11-02
Support inner join in the filter mapping in the Engine

YannanGao-gs:
Support inner join in the filter mapping in the Engine
The grammar is: ~filter [mapping::db] (INNER) @Firm_Person | [mapping::db] FirmFilter


AFine-gs:
add a test to throw an error with an invalid  join of "inner"

PR #442, opened 2021-11-15
Remove Duplicate Property Validation

mrudula-gs:
Remove Duplicate Property Validation for Class & Class-Association


davidharte-gs (commented):
Looks good, thanks

PR #446, opened 2021-11-17
generate test Runtime for runtime pointer provided in service spec

PrateekGarg-gs:
Currently If Runtime path is provided is service spec, while generating H2 runtime for testing , it returns the original runtime instead of generating a new H2 runtime. Handled this case with this commit


dave-wathen (approved):

davidharte-gs (approved):
Approved

PR #447, opened 2021-11-18
corrected SchemaTo field being dropped in postprocessing mappers in toGrammar Api

PrateekGarg-gs:
SchemaTo field in postprocessing mappers was being dropped if SchemaTo=SchemaFrom. It was causing failures in Studio when switching to text mode . As Studio tried to fetch grammar from Legend-sdlc but this 'schemaTo' field was being dropped , causing failure in compilation. Added Round trip test for same case


dave-wathen (approved):

davidharte-gs (approved):
Approved

PR #458, opened 2021-11-27
Allow StreamingResult in service tests

dave-wathen:


aziemchawdhary-gs (approved):

dave-wathen:
Requires SDLC release that depends on engine containing this change.

PR #459, opened 2021-11-30
Service Test Parameters - multiple parameters fix

aayush81198:
The service registration fix was added for parameters of list type, which was breaking in the earlier release.


aziemchawdhary-gs:
Can we have a test added to this codebase to test out the bug this is fixing, as well as the tests in downstream projects?

PR #460, opened 2021-11-30
fixing the grammar with left outer join, throwing parser issue for 'inner'

gs-gunjan:
This PR contains a fix for left_outer join type when it is being composed to grammar.
The change also contains making the joinType case sensitive, it would throw an error otherwise


gs-gunjan:
This PR also includes tests

davidharte-gs (approved):
Approved. Thx

PR #465, opened 2021-12-03
Fix batching flow for inMemory XStore node

hardikmaheshwari:
Issue -
Currently when we handle InMemory XStore operations we expect every child object has an associated parent object. If that's not the case we throw below error
```
No parent was found for a child object
```
Proposed Resolution - 
I believe we shouldn't expect every child to have associated parent & neglect any child that does not have a corresponding parent as dictated by XStore mapping. To resolve the issue removing the check we had

Tests - 
Updated test data for existing tests to include such a case where a child object does not have associated parent. Test fails without these changes and works with these changes


gs-ssh16 (approved):
LGTM

PR #467, opened 2021-12-07
set fetch size to stream result data

PrateekGarg-gs:


legend-engine-executionPlan-execution-store-relational/src/main/java/org/finos/legend/engine/plan/execution/stores/relational/result/SQLExecutionResult.java
```diff
@@ -80,6 +80,7 @@ public SQLExecutionResult(List<ExecutionActivity> activities, SQLExecutionNode S
         {
             this.connection = connection;
             this.statement = connection.createStatement();
+            this.statement.setFetchSize(100);
```
aziemchawdhary-gs, line 83: This is setting fetchSize to a specific value for every DB connection we make - I don't think we want to do this

davidharte-gs (approved):
Approved

PR #489, opened 2021-12-21
Allow build on Mac OSX

dave-wathen:
Also tested these changes on Windows and Linux


akphi:
If this becomes a concern, why dont we use CI matrix for Github action to setup parallel CI jobs running on various OSes? 

https://docs.github.com/en/actions/learn-github-actions/workflow-syntax-for-github-actions#example-running-with-multiple-operating-systems

dave-wathen:
That's a nice option.  I'm not sure the fixes here warrant the complexity:

- Temporary file handling which could probably be written in an entirely platform dependent fashion.  I didn't do so because the current approach allows some configurability.
- Test port discovery which is not really OS specific.  It would have failed on any multihomed box regardless of OS.  It just so happened to cause a port clash on my machine.

aziemchawdhary-gs (approved):

PR #506, opened 2022-01-11
Enhance service grammar to support execution options

rafaelbey:
This expose the service feature that legend-pure already supports.  Execution options are implemented as extensions that different DSL modules can implement, hence the grammar is done thru islands (#{...}#).


davidharte-gs (approved):
Approved

rafaelbey:
Discussed with @pierredebelen and we will approach our use case thru a different mechanims.

PR #507, opened 2022-01-11
Fix local mapping property parsing

hardikmaheshwari:
It seems there is a bug in parsing local mapping property for relational mapping. Instead of populating the owner of the property we are using local property type. 
This bug impacts studio compilation phase

### DIFF
--- a/legend-engine-language-pure-store-relational/src/main/java/org/finos/legend/engine/language/pure/grammar/from/RelationalParseTreeWalker.java
+++ b/legend-engine-language-pure-store-relational/src/main/java/org/finos/legend/engine/language/pure/grammar/from/RelationalParseTreeWalker.java
@@ -1098,7 +1098,7 @@ public class RelationalParseTreeWalker
         multiplicity.setUpperBound("*".equals(toMultiplicityContext.getText()) ? null : Integer.parseInt(toMultiplicityContext.getText()));
         PropertyPointer propertyPointer = new PropertyPointer();
         propertyPointer.property = PureGrammarParserUtility.fromIdentifier(ctx.identifier());
-        propertyPointer._class = localMappingPropertyInfo.type;
+        propertyPointer._class = null;
         propertyPointer.sourceInformation = this.walkerSourceInformation.getSourceInformation(ctx.identifier());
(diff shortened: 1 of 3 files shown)


legend-engine-language-pure-store-relational/src/main/java/org/finos/legend/engine/language/pure/grammar/from/RelationalParseTreeWalker.java
```diff
@@ -1098,7 +1098,7 @@ private PropertyMapping visitPropertyMappingWithLocalMappingProperty(RelationalP
         multiplicity.setUpperBound("*".equals(toMultiplicityContext.getText()) ? null : Integer.parseInt(toMultiplicityContext.getText()));
         PropertyPointer propertyPointer = new PropertyPointer();
         propertyPointer.property = PureGrammarParserUtility.fromIdentifier(ctx.identifier());
-        propertyPointer._class = localMappingPropertyInfo.type;
+        propertyPointer._class = _class;
```
MauricioUyaguari, line 1101: What does this fix? Ideally, we don't want to infer anything in the grammar as it should be 1 to 1 with the protocol. Could we fix the compiler instead?

hardikmaheshwari: As of now we are assigning type or property as its owner which is wrong. We can do this at parser stage because looking at the class mapping we know the name of property owner. We will resolve the owner at the compilation stage

gs-ssh16, line 1101: Don't think setting it to _class is correct. The class being mapped will not have this property. The mapping class which gets created at the compile time will be the owner. Hence, leaving it as null makes more sense.

gs-ssh16 (changes requested):

legend-engine-language-pure-store-relational/src/main/java/org/finos/legend/engine/language/pure/grammar/from/RelationalParseTreeWalker.java
hardikmaheshwari: @gs-ssh16 updated the changes. Please take a look

gs-ssh16 (approved):

PR #513, opened 2022-01-16
Fixes to allow usage of meta::pure::mapping::from on functions

rafaelbey:
To properly support meta::pure::mapping::from, the functions second pass needs to happen after connections and runtime to ensure these can be properly accessed while resolving function arguments.  

Also, given runtime are not really packaged, the stub needs to be converted to an actual runtime.

### DIFF
--- a/legend-engine-language-pure-compiler/src/main/java/org/finos/legend/engine/language/pure/compiler/toPureGraph/PureModel.java
+++ b/legend-engine-language-pure-compiler/src/main/java/org/finos/legend/engine/language/pure/compiler/toPureGraph/PureModel.java
@@ -403,7 +403,6 @@ public class PureModel implements IPureModel
         // Fourth pass - qualifiers
         pure.classes.forEach(el -> visitWithErrorHandling(el, new PackageableElementFourthPassBuilder(this.getContext(el))));
         pure.associations.forEach(el -> visitWithErrorHandling(el, new PackageableElementThirdPassBuilder(this.getContext(el))));
-        pure.functions.forEach(el -> visitWithErrorHandling(el, new PackageableElementSecondPassBuilder(this.getContext(el))));
     }
 
     private void loadStores(PureModelContextDataIndex pure)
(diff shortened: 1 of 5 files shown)


legend-engine-language-pure-compiler/src/main/java/org/finos/legend/engine/language/pure/compiler/toPureGraph/PureModel.java
```diff
@@ -439,6 +438,8 @@ private void loadOtherElementsPreStores(PureModelContextDataIndex pure)
 
     private void loadOtherElementsPostConnectionsAndRuntimes(PureModelContextDataIndex pure)
     {
+        // functions might reference runtime and connections
+        pure.functions.forEach(el -> visitWithErrorHandling(el, new PackageableElementSecondPassBuilder(this.getContext(el))));
```
rafaelbey, line 442: Can there be a negative side effect on moving this after connections and runtime?

legend-engine-language-pure-compiler/src/main/java/org/finos/legend/engine/language/pure/compiler/toPureGraph/ValueSpecificationBuilder.java
```diff
@@ -132,7 +134,18 @@ public ValueSpecification visit(org.finos.legend.engine.protocol.pure.v1.model.v
     @Override
     public ValueSpecification visit(org.finos.legend.engine.protocol.pure.v1.model.valueSpecification.raw.PackageableElementPtr packageableElementPtr)
     {
-        org.finos.legend.pure.m3.coreinstance.meta.pure.metamodel.PackageableElement packageableElement = this.context.resolvePackageableElement(packageableElementPtr.fullPath, packageableElementPtr.sourceInformation);
+        PackageableElement packageableElement = this.context.resolvePackageableElement(packageableElementPtr.fullPath, packageableElementPtr.sourceInformation);
+        if (packageableElement instanceof Root_meta_pure_runtime_PackageableRuntime)
```
rafaelbey, line 138: Could there be a better place to convert PackageableRuntime to an actual Runtime?

rafaelbey:
@pierredebelen - These are the initial set of changes to ensure the  _meta::pure::mapping::from_ function can be leveraged.  The main challenge is that Runtime are not really packaged, and we need to go from the stub to the real runtime.

PR #518, opened 2022-01-20
Add new APIs to ServiceRunner to allow better programmatic executions

rafaelbey:
This PR adds new APIs to ServiceRunner to be able to:

- Get the service result type - useful to know how to parse the results of the execution (tds, graph, constant value, etc)
- Get the list of variables the service expects and details of these variables - useful when operating on the ServicerRunner interface rather than on the SDLC generated instance.  

With the changes, the PR also ensure all the run/execution APIs support the same features, reducing confusion among users.

This also moves execution logic from legend-sdlc, letting SDLC define the service and its arguments, and engine to define how these will be use and executed.

### DIFF
--- a/legend-engine-language-pure-dsl-service-execution/src/main/java/org/finos/legend/engine/language/pure/dsl/service/execution/AbstractServicePlanExecutor.java
+++ b/legend-engine-language-pure-dsl-service-execution/src/main/java/org/finos/legend/engine/language/pure/dsl/service/execution/AbstractServicePlanExecutor.java
@@ -291,21 +291,29 @@ public abstract class AbstractServicePlanExecutor implements ServiceRunner
             result.accept(new ResultVisitor<Void>()
             {
                 @Override
-                public Void visit(ErrorResult errorResult)
+                public Void visit(StreamingResult streamingResult)
                 {
-                    throw new RuntimeException("Error: " + ((ErrorResult) result).getMessage());
+                    try
+                    {
+                        streamingResult.stream(outputStream, serializationFormat);
+                        return null;
+                    }
+                    catch (IOException e)
+                    {
+                        throw new UncheckedIOException("Error serializing result", e);
+                    }
                 }
 
                 @Override
                 public Void visit(StreamingObjectResult tStreamingObjectResult)
                 {
-                    return stream(tStreamingObjectResult);
+                    return visit((StreamingResult) tStreamingObjectResult);
                 }
 
                 @Override
                 public Void visit(JsonStreamingResult jsonStreamingResult)
                 {
-                    return stream(jsonStreamingResult);
+                    return visit((StreamingResult) jsonStreamingResult);
                 }
 
                 @Override
@@ -324,22 +332,15 @@ public abstract class AbstractServicePlanExecutor implements ServiceRunner
                 }
 
                 @Override
+                public Void visit(ErrorResult errorResult)
+                {
+                    throw new RuntimeException("Error: " + ((ErrorResult) result).getMessage());
+                }
+
+                @Override
                 public Void visit(MultiResult multiResult)
                 {
                     throw new RuntimeException("Unknown result type: " + result.getClass().getCanonicalName());
-                }
-
-                private Void stream(StreamingResult streamingResult)
-                {
-                    try
-                    {
-                        streamingResult.stream(outputStream, serializationFormat);
-                        return null;
-                    }
-                    catch (IOException e)
-                    {
-                        throw new UncheckedIOException("Error serializing result", e);
-                    }
                 }
             });
         }
--- a/legend-engine-language-pure-dsl-service-execution/src/main/java/org/finos/legend/engine/language/pure/dsl/service/execution/AbstractServicePlanExecutor.java
+++ b/legend-engine-language-pure-dsl-service-execution/src/main/java/org/finos/legend/engine/language/pure/dsl/service/execution/AbstractServicePlanExecutor.java
@@ -24,14 +24,22 @@ import org.finos.legend.engine.plan.execution.cache.graphFetch.GraphFetchCache;
 import org.finos.legend.engine.plan.execution.cache.graphFetch.GraphFetchCrossAssociationKeys;
 import org.finos.legend.engine.plan.execution.result.ConstantResult;
 import org.finos.legend.engine.plan.execution.result.ErrorResult;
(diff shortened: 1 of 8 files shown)


rafaelbey:
**THIS BREAKS legned-sdlc**

legend-engine-language-pure-dsl-service-execution/src/main/java/org/finos/legend/engine/language/pure/dsl/service/execution/AbstractServicePlanExecutor.java
```diff
@@ -122,19 +136,82 @@ protected AbstractServicePlanExecutor(String servicePath, String executionPlanRe
...
+        else if(this.plan instanceof CompositeExecutionPlan)
+        {
+            List<ResultType> resultTypes = ((CompositeExecutionPlan) this.plan).executionPlans.values()
+                    .stream()
+                    .map(x -> x.rootExecutionNode.resultType)
+                    .collect(Collectors.toList());
+
+            Assert.assertFalse(resultTypes.isEmpty(), () -> "Composite plan without any actual plan?");
+            Assert.assertTrue(resultTypes.stream().allMatch(x -> x.getClass().equals(resultTypes.get(0).getClass())), () -> "Composite with different result types among plans?");
+            return resultTypes.get(0);
```
rafaelbey, line 167: @kevin-m-knight-gs : I added the validation here.  While every plan should yield the same result type, I'm hesitant to add the validation to plan  generation given the negative impact that can have if is not the case.   If we face issues with this API, we can then understand the reasoning and adapt better.  Thoughts?

kevin-m-knight-gs: This validation is a little strange. We're only validating that all the result types are of the same class, not that they're all fully equal. I think full equality is the validation we really want to do. And I think we really should do it at plan generation time, not in this method. Since it's not the validation we actually want and it's not in the place we actually want it, let's just leave it off for now. You can revert this bit to what it was before.

However, we genuinely do need to understand whether the design is that the result type should be equal across all execution plans within a composite plan (even if the validations aren't fully in place to ensure it). If that is the case, then this API is fundamentally correct. But if it's possible to have different result types, then this API is fundamentally not correct. So we need to determine that before going forward.

legend-engine-language-pure-dsl-service-execution/src/main/java/org/finos/legend/engine/language/pure/dsl/service/execution/AbstractServicePlanExecutor.java
```diff
@@ -122,19 +136,82 @@ protected AbstractServicePlanExecutor(String servicePath, String executionPlanRe
...
+        }
+        else
+        {
+            throw new UnsupportedOperationException("Not supported: " + this.plan.getClass());
+        }
+    }
+
+    protected ExecutionBuilder executionBuilder(ServiceRunnerInput serviceRunnerInput, StreamProvider streamProvider)
+    {
+        List<ServiceParameter> parameters = this.getParameters();
```
kevin-m-knight-gs, line 177: Since the default implementation of this throws UnsupportedOperationException, we probably should handle that case.

legend-engine-language-pure-dsl-service-execution/src/main/java/org/finos/legend/engine/language/pure/dsl/service/execution/AbstractServicePlanExecutor.java
```diff
@@ -122,19 +136,82 @@ protected AbstractServicePlanExecutor(String servicePath, String executionPlanRe
...
+            Assert.assertTrue(resultTypes.stream().allMatch(x -> x.getClass().equals(resultTypes.get(0).getClass())), () -> "Composite with different result types among plans?");
+            return resultTypes.get(0);
+        }
+        else
+        {
+            throw new UnsupportedOperationException("Not supported: " + this.plan.getClass());
+        }
+    }
+
+    protected ExecutionBuilder executionBuilder(ServiceRunnerInput serviceRunnerInput, StreamProvider streamProvider)
```
kevin-m-knight-gs, line 175: I think newExecutionBuilder would be a better name, so that it's symmetric with the other factory methods for constructing an ExecutionBuilder.

legend-engine-language-pure-dsl-service-execution/src/main/java/org/finos/legend/engine/language/pure/dsl/service/execution/AbstractServicePlanExecutor.java
rafaelbey: As discussed, after research, for relational the result type can be different on the tds relational column type.  Also, result type have editable fields that could impact the internal plan.  

For now, I will remove the getResultType() method until I can give more proper thoughts to how expose it.

legend-engine-language-pure-dsl-service-execution/src/main/java/org/finos/legend/engine/language/pure/dsl/service/execution/AbstractServicePlanExecutor.java
```diff
@@ -194,74 +271,105 @@ protected ExecutionBuilder newExecutionBuilder()
...
-                    throw new RuntimeException("Error serializing result", e);
+                    throw new RuntimeException("Unknown result type: " + result.getClass().getCanonicalName());
                 }
-            }
-            else
-            {
-                throw new RuntimeException("Unknown result type: " + result.getClass().getCanonicalName());
-            }
+
+                private Void stream(StreamingResult streamingResult)
```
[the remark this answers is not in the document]
rafaelbey: Haa! Thanks for catching up this huge miss.

legend-engine-language-pure-dsl-service-execution/src/main/java/org/finos/legend/engine/language/pure/dsl/service/execution/AbstractServicePlanExecutor.java
rafaelbey: How I would go about handling this?

kevin-m-knight-gs (approved):

kevin-m-knight-gs (approved):

PR #532, opened 2022-02-01
fix potential NullPointerException

gs-gunjan:
adding null checks to fix potential NullPointerException


erwin-jeanlouis-gs (approved):
Looks good

PR #534, opened 2022-02-01
Improved existing test cases to allow new template functions without failing.

vj2050gs:


vj2050gs:
fixing permission issues

PR #548, opened 2022-02-11
add support for merge operation

AFine-gs:
add support for merge operation


AFine-gs:
This has a dependency on   https://github.com/finos/legend-pure/pull/452

PR #568, opened 2022-02-25
Update Docker Image version of mssqlServer and version update for Test containers

PrateekGarg-gs:


pom.xml
```diff
@@ -1638,6 +1638,20 @@
...
+            <dependency>
+                <groupId>org.testcontainers</groupId>
+                <artifactId>testcontainers</artifactId>
+                <version>1.16.3</version>
+            </dependency>
+            <dependency>
+                <groupId>org.testcontainers</groupId>
+                <artifactId>mssqlserver</artifactId>
+                <version>1.16.3</version>
+                <scope>test</scope>
```
vikask-gs, line 1651: remove test scope. In base pom, no scope should be mentioned. It is upto other poms to specify scope as per need.

PR #582, opened 2022-03-07
Add ability to build service runner using pure element path

rafaelbey:
Add the ability to use pure element path in conjunction with studio project coordinates to find and load the ServiceRunner class.  This allow users to focus on first class concepts Studio present rather than expecting the generated class name.

### DIFF
--- a/legend-engine-shared-core/src/main/java/org/finos/legend/engine/plan/platform/java/JavaSourceHelper.java
+++ b/legend-engine-shared-core/src/main/java/org/finos/legend/engine/plan/platform/java/JavaSourceHelper.java
@@ -34,6 +34,7 @@ import java.util.Map;
 import java.util.Set;
 import java.util.TreeMap;
 import java.util.function.BiConsumer;
+import java.util.stream.Collectors;
 import java.util.stream.Stream;
 
 public class JavaSourceHelper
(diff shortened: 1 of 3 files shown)


legend-engine-shared-core/src/main/java/org/finos/legend/engine/plan/platform/java/JavaSourceHelper.java
```diff
@@ -277,4 +278,20 @@ private static boolean fileHasContent(Path path, byte[] content) throws IOExcept
     {
         return (Files.size(path) == content.length) && Arrays.equals(content, Files.readAllBytes(path));
     }
+
+    public static String serviceElementPathToJavaPath(String packagePrefix, String servicePath)
```
rafaelbey, line 282: This is similar as the logic [SDLC uses to generate the class name](https://github.com/finos/legend-sdlc/blob/19fcd19f51e2ae9baa7bff53152658e740abbf8f/legend-sdlc-generation-service/src/main/java/org/finos/legend/sdlc/generation/service/ServiceExecutionClassGenerator.java#L116)

If we go with this direction, will update SDLC to use this method rather than having duplicated logic.

PR #597, opened 2022-03-18
fix kerberos ticket iterator

elopezcastro:
fix kerberos ticket iterator

### DIFF
--- a/legend-engine-shared-core/src/main/java/org/finos/legend/engine/shared/core/identity/credential/LegendKerberosCredential.java
+++ b/legend-engine-shared-core/src/main/java/org/finos/legend/engine/shared/core/identity/credential/LegendKerberosCredential.java
@@ -16,10 +16,7 @@ package org.finos.legend.engine.shared.core.identity.credential;
 
 import javax.security.auth.Subject;
 import javax.security.auth.kerberos.KerberosTicket;
-import javax.swing.text.html.HTMLDocument;
-
 import org.finos.legend.engine.shared.core.identity.Credential;
-
 import java.util.Iterator;
 
 /*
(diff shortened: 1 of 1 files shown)


legend-engine-shared-core/src/main/java/org/finos/legend/engine/shared/core/identity/credential/LegendKerberosCredential.java
```diff
@@ -45,7 +45,11 @@ public void setSubject(Subject subject)
     @Override
     public boolean isValid()
     {
-        KerberosTicket kerberosTicket = subject.getPrivateCredentials(KerberosTicket.class).iterator().next();
-        return kerberosTicket!=null && kerberosTicket.isCurrent();
+        if (subject.getPrivateCredentials(KerberosTicket.class) != null)
```
aormerod-gs, line 48: should subject.getPrivateCredentials(KerberosTicket.class) be assigned to a local variable to match for the check and the usage? (Or is there no chance of it being modified between these two lines?)

elopezcastro: done

aormerod-gs (approved):

epsstan (approved):

epsstan:
@elopezcastro - Should we also change other code paths that look for the ticket ? Like this one in legend-shared ? https://github.com/finos/legend-shared/blob/master/legend-shared-pac4j-kerberos/src/main/java/org/finos/legend/server/pac4j/kerberos/KerberosProfile.java

PR #648, opened 2022-04-20
create relational validation API

AFine-gs:
create relational validation API


AFine-gs:
Depends on  https://github.com/finos/legend-pure/pull/537

PR #655, opened 2022-04-26
Support current user id in engine.

CptTeddy:
- Support current user id in engine.
- Add UserExtension to allow potential support for non-kerberos profile types in Alloy.


legend-engine-executionPlan-execution/src/main/java/org/finos/legend/engine/plan/execution/nodes/state/ExecutionState.java
```diff
@@ -126,9 +126,17 @@ public ExecutionState varName(String var)
...
     public ExecutionState setAuthUser(String user)
+    {
+        return setAuthUser(user, true);
+    }
+
+    public ExecutionState setAuthUser(String user, boolean setTransformAllocation)
     {
         this.authId = user;
-        this.transformAllocation = true;
+        if (setTransformAllocation)
```
epsstan, line 136: This if check is redundant.  this.transformAllocation  = setTransformAllocation is sufficient

legend-engine-executionPlan-execution/src/main/java/org/finos/legend/engine/plan/execution/PlanExecutor.java
```diff
@@ -169,6 +170,9 @@ public Result execute(SingleExecutionPlan singleExecutionPlan, ExecutionState st
             if (singleExecutionPlan.authDependent) {
                 state.setAuthUser((singleExecutionPlan.kerberos == null) ? user : singleExecutionPlan.kerberos);
             }
+            if (state.authId == null) {
+                state.setAuthUser(IdentityFactoryProvider.getInstance().makeIdentity(profiles).getName(), false);
+            }
             singleExecutionPlan.getExecutionStateParams(org.eclipse.collections.api.factory.Maps.mutable.empty()).forEach(state::addParameterValue);
```
epsstan, line 176: Can line 1761 be the same as 174 ? atleast wrto computing the current user ?

epsstan (changes requested):

legend-engine-executionPlan-execution/src/main/java/org/finos/legend/engine/plan/execution/nodes/state/ExecutionState.java
CptTeddy: Fixed. Thanks!

legend-engine-executionPlan-execution/src/main/java/org/finos/legend/engine/plan/execution/PlanExecutor.java
CptTeddy: Yes, this will evaluate to the same value wrto. evaluating current user. In the case _kerberos_ or _user_ is non null, this will evaluate to the Kerberos value we expect.

CptTeddy: If authDependent, the kerberos is bound to the plan while IdentityFactory evaluates the profile id at runtime. For current_user, we don't care about this as much since semantically we could support "current" as being extracted from the plan or the profile. We didn't have a plan previously since we didn't support this in Engine, so this is up to us to define if reasonable.

PR #658, opened 2022-04-28
add relational testable tests

MauricioUyaguari:
image

### DIFF
--- a/legend-engine-test-runner-service/src/main/java/org/finos/legend/engine/testable/service/assertion/ServiceTestAssertionEvaluator.java
+++ b/legend-engine-test-runner-service/src/main/java/org/finos/legend/engine/testable/service/assertion/ServiceTestAssertionEvaluator.java
@@ -1,22 +1,21 @@
-//  Copyright 2022 Goldman Sachs
+// Copyright 2022 Goldman Sachs
 //
-//  Licensed under the Apache License, Version 2.0 (the "License");
-//  you may not use this file except in compliance with the License.
-//  You may obtain a copy of the License at
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
 //
-//       http://www.apache.org/licenses/LICENSE-2.0
+//      http://www.apache.org/licenses/LICENSE-2.0
 //
-//  Unless required by applicable law or agreed to in writing, software
-//  distributed under the License is distributed on an "AS IS" BASIS,
-//  WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
-//  See the License for the specific language governing permissions and
-//  limitations under the License.
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
 
 package org.finos.legend.engine.testable.service.assertion;
 
-import com.fasterxml.jackson.databind.JsonNode;
-import com.fasterxml.jackson.databind.ObjectMapper;
-import com.fasterxml.jackson.databind.SerializationFeature;
+import java.util.stream.Collectors;
+import java.util.stream.Stream;
 import org.finos.legend.engine.plan.execution.result.ConstantResult;
 import org.finos.legend.engine.plan.execution.result.Result;
 import org.finos.legend.engine.plan.execution.result.StreamingResult;
@@ -28,22 +27,48 @@ import org.finos.legend.engine.protocol.pure.v1.model.test.assertion.TestAsserti
 import org.finos.legend.engine.protocol.pure.v1.model.test.assertion.status.AssertFail;
 import org.finos.legend.engine.protocol.pure.v1.model.test.assertion.status.AssertPass;
 import org.finos.legend.engine.protocol.pure.v1.model.test.assertion.status.AssertionStatus;
-import org.finos.legend.engine.protocol.pure.v1.model.test.assertion.status.EqualToJsonAssertFail;
-import org.finos.legend.engine.test.runner.shared.JsonNodeComparator;
 import org.finos.legend.engine.testable.service.helper.PrimitiveValueSpecificationToObjectVisitor;
 
 public class ServiceTestAssertionEvaluator implements TestAssertionEvaluator
 {
-    private Result result;
+    private final String serializationFormat;
+    private final Result result;
+
+    public ServiceTestAssertionEvaluator(Result result, String serializationFormat)
+    {
+        this.result = result;
+        this.serializationFormat = serializationFormat;
+
+    }
 
     public ServiceTestAssertionEvaluator(Result result)
     {
-        this.result = result;
+        this(result, null);
     }
+
+
+    private SerializationFormat getSerializationFormat()
+    {
+        if (this.serializationFormat == null)
+        {
+            return SerializationFormat.defaultFormat;
+        }
+        try
+        {
+            return SerializationFormat.valueOf(this.serializationFormat);
+        }
+        catch (IllegalArgumentException exception)
+        {
+            throw new    UnsupportedOperationException("Unsupported serialization format '" + this.serializationFormat
+                + "'." + "Supported formats are:" + Stream.of(SerializationFormat.values()).map(SerializationFormat::name).collect(Collectors.joining(",")));
+        }
+    }
+
 
     @Override
     public AssertionStatus visit(TestAssertion testAssertion)
     {
+        SerializationFormat serializationFormat = this.getSerializationFormat();
         if (testAssertion instanceof EqualTo)
         {
             if (!(result instanceof ConstantResult))
@@ -62,7 +87,7 @@ public class ServiceTestAssertionEvaluator implements TestAssertionEvaluator
             else
             {
                 AssertFail assertFail = new AssertFail();
-                assertFail.message = "Expected : " + expected.toString() + ", Found : " + actual.toString();
+                assertFail.message = "Expected : " + expected + ", Found : " + actual.toString();
 
                 assertionStatus = assertFail;
             }
@@ -79,44 +104,21 @@ public class ServiceTestAssertionEvaluator implements TestAssertionEvaluator
             }
             else if (result instanceof StreamingResult)
             {
-                actualJson = ((StreamingResult) result).flush(((StreamingResult) result).getSerializer(SerializationFormat.PURE));
+                actualJson = ((StreamingResult) result).flush(((StreamingResult) result).getSerializer(serializationFormat));
             }
             else
             {
                 throw new UnsupportedOperationException("Result type - " + result.getClass().getSimpleName() + " not supported with EqualToJson Assert !!");
             }
-
             try
             {
-                ObjectMapper objectMapper = new ObjectMapper();
-                JsonNode expected = objectMapper.readTree(((EqualToJson) testAssertion).expected.data.getBytes());
-                JsonNode actual = objectMapper.readTree(actualJson.getBytes());
-
-                AssertionStatus assertionStatus;
-                if (JsonNodeComparator.NULL_MISSING_EQUIVALENT.compare(expected, actual) == 0)
-                {
-                    assertionStatus = new AssertPass();
-                }
-                else
-                {
-                    EqualToJsonAssertFail fail = new EqualToJsonAssertFail();
-                    fail.expected = objectMapper.writer(SerializationFeature.INDENT_OUTPUT, SerializationFeature.ORDER_MAP_ENTRIES_BY_KEYS).writeValueAsString(expected);
-                    fail.actual = objectMapper.writer(SerializationFeature.INDENT_OUTPUT, SerializationFeature.ORDER_MAP_ENTRIES_BY_KEYS).writeValueAsString(actual);
-                    fail.message = "Actual result does not match Expected result";
-
-                    assertionStatus = fail;
-                }
-                assertionStatus.id = testAssertion.id;
-                return assertionStatus;
+                return ServiceTestAssertionHelper.compareAssertionJSON(testAssertion, ((EqualToJson) testAssertion).expected.data, actualJson);
             }
             catch (Exception e)
             {
                 throw new RuntimeException(e);
             }
         }
-        else
-        {
-            return null;
-        }
+        return null;
     }
 }
--- a/legend-engine-test-runner-service/src/main/java/org/finos/legend/engine/testable/service/extension/ServiceTestRunner.java
+++ b/legend-engine-test-runner-service/src/main/java/org/finos/legend/engine/testable/service/extension/ServiceTestRunner.java
@@ -12,8 +12,17 @@
 //  See the License for the specific language governing permissions and
 //  limitations under the License.
 
+
 package org.finos.legend.engine.testable.service.extension;
 
+import static org.finos.legend.engine.language.pure.compiler.toPureGraph.HelperModelBuilder.getElementFullPath;
+
+import java.io.Closeable;
+import java.io.IOException;
+import java.util.ArrayList;
+import java.util.List;
+import java.util.Map;
+import java.util.ServiceLoader;
 import org.eclipse.collections.api.RichIterable;
 import org.eclipse.collections.api.factory.Lists;
 import org.eclipse.collections.api.factory.Maps;
@@ -78,15 +87,6 @@ import org.finos.legend.pure.generated.Root_meta_pure_test_AtomicTest;
 import org.finos.legend.pure.generated.Root_meta_pure_test_TestSuite;
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
-
-import java.io.Closeable;
-import java.io.IOException;
-import java.util.ArrayList;
-import java.util.List;
-import java.util.Map;
-import java.util.ServiceLoader;
-
-import static org.finos.legend.engine.language.pure.compiler.toPureGraph.HelperModelBuilder.getElementFullPath;
 
 public class ServiceTestRunner implements TestRunner
 {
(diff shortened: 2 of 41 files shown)


legend-engine-testable/src/main/java/org/finos/legend/engine/testable/api/Testable.java
```diff
@@ -48,24 +53,27 @@
 
     private final TestableRunner testableRunner;
 
+    private static final ObjectMapper objectMapper = ObjectMapperFactory.getNewStandardObjectMapperWithPureProtocolExtensionSupports();
```
hardikmaheshwari, line 56: Should we load this while initialization.
If we load in this fashion it would be susceptible to thread context this class gets loaded in.
What do you think?

legend-engine-test-runner-service/src/main/java/org/finos/legend/engine/testable/service/assertion/RelationalTestAssertionHelper.java
```diff
@@ -0,0 +1,40 @@
...
+    }
+
+    public String flushToJSONString()
+    {
+        try
+        {
+            return relationalResult.flush(relationalResult.getSerializer(SerializationFormat.PURE_TDSOBJECT));
+        }
+        catch (RuntimeException error)
+        {
```
hardikmaheshwari, line 36: why do we use pure_tds object as default serialization. Seems a bit off from naming, should we sue default?

legend-engine-test-runner-service/src/main/java/org/finos/legend/engine/testable/service/assertion/ServiceTestAssertionEvaluator.java
```diff
@@ -66,6 +66,10 @@ else if (testAssertion instanceof EqualToJson)
             {
                 actualJson = (String) ((ConstantResult) result).getValue();
             }
+            else if (result instanceof RelationalResult)
```
hardikmaheshwari, line 69: I think store specific logic should not come to this module & do we need this, should we just default to streaming result with default serialization format
and have an option in test specification to specify. What are your thoughts

legend-engine-test-runner-service/src/main/java/org/finos/legend/engine/testable/service/assertion/ServiceTestAssertionEvaluator.java
```diff
@@ -103,9 +107,10 @@ else if (result instanceof StreamingResult)
                 throw new RuntimeException(e);
             }
         }
+        // TODO: EqualToTDS
         else
         {
-            return null;
+            throw new UnsupportedOperationException("Unsupported assertion '" + testAssertion.id + "' of type '" + testAssertion.getClass().getName() + "'");
```
hardikmaheshwari, line 113: I don't think we should throw as we expect to have multiple visitor and we should pick non-null values. So we should not throw exception here

legend-engine-test-runner-service/src/main/java/org/finos/legend/engine/testable/service/extension/ServiceTestRunner.java
```diff
@@ -16,6 +16,7 @@
 import org.finos.legend.engine.plan.execution.result.ConstantResult;
 import org.finos.legend.engine.plan.execution.result.Result;
 import org.finos.legend.engine.plan.execution.result.StreamingResult;
+import org.finos.legend.engine.plan.execution.result.json.JsonStreamingResult;
```
hardikmaheshwari, line 19: do we need this?

legend-engine-test-runner-service/src/test/resources/testable/legend-testable-service-M2M.pure
```diff
@@ -0,0 +1,137 @@
...
+              url: 'data:application/json,%7B%22employees%22%3A%5B%7B%22firstName%22%3A%22firstName%2059%22%2C%22lastName%22%3A%22lastName%2050%22%7D%5D%2C%22legalName%22%3A%22legalName%2011%22%2C%22type%22%3A%22llc%22%7D';
+            }
+          }#
+        ]
+      ];
+    }#;
+  }
+  testSuites:
+  [
+    testSuite1:
```
hardikmaheshwari, line 46: do we not already have this test?
Sorry I may have missed it

legend-engine-test-runner-service/src/test/resources/testable/legend-testable-service-relational.pure
```diff
@@ -0,0 +1,265 @@
+###Data
+Data data::RelationalData
+{
+  Relational
+  #{
+    PersonTable:
+      {
+        columns:
```
hardikmaheshwari, line 8: Should we just allow CSV format, that feels more convenient. Possibly I am missing the reason

legend-engine-xt-relationalStore-pure/src/main/resources/core_relational/relational/metamodel/test/relationalTest.pure
```diff
@@ -0,0 +1,53 @@
...
+
+// ------------------------------------------------------------------------------ DATA -----------------------------------------------
+Class meta::relational::metamodel::data::RelationalData extends  meta::pure::data::EmbeddedData
+{
+  tables: meta::relational::metamodel::data::RelationalDataTable[*];
+}
+
+Class meta::relational::metamodel::data::RelationalDataTable
+{
+  schemaName: String[0..1];
```
hardikmaheshwari, line 37: should schema name be optional?

legend-engine-xt-relationalStore-pure/src/main/resources/core_relational/relational/metamodel/test/relationalTest.pure
```diff
@@ -0,0 +1,53 @@
...
+  tableName: String[1];
+  columns: meta::relational::metamodel::data::RelationalDataTableColumn[*];
+  rows: meta::relational::metamodel::data::RelationalDataTableRow[*];
+}
+
+Class meta::relational::metamodel::data::RelationalDataTableRow
+{
+  values: String[1];
```
hardikmaheshwari, line 45: should we do the parsing of row values in the parser step itself too so that we have a list of values of rows, this might be helpful as parsing in java would be easier

legend-engine-xt-relationalStore-protocol/src/main/java/org/finos/legend/engine/protocol/pure/v1/model/packageableElement/store/relational/data/RelationalDataTableRow.java
```diff
@@ -0,0 +1,23 @@
...
+// limitations under the License.
+
+package org.finos.legend.engine.protocol.pure.v1.model.packageableElement.store.relational.data;
+
+import org.finos.legend.engine.protocol.pure.v1.model.SourceInformation;
+
+public class RelationalDataTableRow
+{
+    public SourceInformation sourceInformation;
+    public String values;
```
hardikmaheshwari, line 22: should we have this as List<String>?

legend-engine-xt-relationalStore-grammar/pom.xml
```diff
@@ -216,11 +216,16 @@
             <artifactId>legend-engine-xt-json-model</artifactId>
             <scope>test</scope>
         </dependency>
-<!--        <dependency>-->
-<!--            <groupId>platform.alloy</groupId>-->
-<!--            <artifactId>legend-engine-language-pure-dsl-service</artifactId>-->
-<!--            <scope>test</scope>-->
-<!--        </dependency>-->
+        <dependency>
```
hardikmaheshwari, line 219: This seems like something which can give rise to a cyclic dependency structure. I had to add such dependencies for tests myself, maybe we should have a module for these tests? What do you think?

legend-engine-xt-relationalStore-grammar/src/main/antlr4/org/finos/legend/engine/language/pure/grammar/from/antlr4/data/RelationalEmbeddedDataParserGrammar.g4
```diff
@@ -0,0 +1,57 @@
...
+;
+
+// -------------------------------------- DEFINITION -------------------------------------
+definition:                                 table*
+                                            EOF
+;
+table:                                      identifier COLON
+                                                BRACE_OPEN
+                                                (
+                                                    schemaName
```
hardikmaheshwari, line 38: should the grouping be by schema as its a higher level concept?

legend-engine-xt-relationalStore-grammar/src/test/java/org/finos/legend/engine/language/pure/compiler/test/TestRelationalDataCompilationFromGrammar.java
```diff
@@ -0,0 +1,371 @@
...
+import org.finos.legend.engine.language.pure.compiler.toPureGraph.PureModel;
+import org.finos.legend.engine.protocol.pure.v1.model.context.PureModelContextData;
+import org.finos.legend.pure.generated.*;
+import org.finos.legend.pure.m3.coreinstance.meta.pure.metamodel.PackageableElement;
+import org.junit.Assert;
+import org.junit.Test;
+
+import static org.finos.legend.engine.language.pure.compiler.test.TestCompilationFromGrammar.TestCompilationFromGrammarTestSuite.test;
+
+public class TestRelationalDataCompilationFromGrammar
```
hardikmaheshwari, line 28: Maybe we should have some tests with special characters like quotes, semi-colon, etc

legend-engine-xt-relationalStore-grammar/src/test/java/org/finos/legend/engine/language/pure/grammar/test/TestRelationalEmbeddedDataGrammarRoundtrip.java
```diff
@@ -0,0 +1,130 @@
...
+                        "        ];\n" +
+                        "      }\n" +
+                        "  }#\n" +
+                        "}\n"
+        );
+    }
+
+
+    @Test
+    public void testAssertToTDS()
```
hardikmaheshwari, line 67: maybe a test with relational data embedded into service as well?

legend-engine-test-runner-service/src/test/java/org/finos/legend/engine/testable/service/TestServiceTestSuite.java
```diff
@@ -2246,4 +2252,106 @@ public void testMultiExecutionServiceTestSuite()
...
+            return null;
+        }
+       catch (Exception e)
+       {
+           throw new RuntimeException(e);
+       }
+    }
+
+    @Test
+    public void testRelationalServiceTestSuite()
```
hardikmaheshwari, line 2275: I think we should have few tests with parameters in query

legend-engine-xt-relationalStore-executionPlan/src/main/java/org/finos/legend/engine/plan/execution/stores/relational/test/RelationalConnectionFactory.java
```diff
@@ -50,4 +57,25 @@ public Optional<Connection> tryBuildFromConnection(Connection connection, String
     {
         return ConnectionFactoryExtension.super.tryBuildFromConnection(connection, testData, element);
     }
+
+    @Override
+    public Optional<Pair<Connection, List<Closeable>>> tryBuildTestConnection(Connection sourceConnection, EmbeddedData data)
+    {
+        if(data instanceof RelationalData  && sourceConnection instanceof RelationalDatabaseConnection)
```
hardikmaheshwari, line 64: way we designed this, we expect that we will be spinning up H2 server here with data populated here itself and pass that as closable. I didn't see H2 server spawned anywhere, where we spawning that?

legend-engine-testable/src/main/java/org/finos/legend/engine/testable/api/Testable.java
```diff
@@ -48,24 +53,27 @@
...
+
+
     public Testable(ModelManager modelManager)
     {
         this.testableRunner = new TestableRunner(modelManager);
     }
 
     @POST
-    @Path("doTests")
+    @Path("runTests")
```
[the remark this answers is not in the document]
MauricioUyaguari: I think runTests make more sense. Using `execute` might imply we are always going through the `execute` query path which is not always the case with this pass

legend-engine-test-runner-service/src/main/java/org/finos/legend/engine/testable/service/assertion/RelationalTestAssertionHelper.java
MauricioUyaguari: How about we use the assert type for this. 
If it is `AssertToJSON` we use the default
Moving to assertToTDS we use the `TDS` serialization for this. 
what are your thoughts for doing that. 
I think it might be an overkill to model another assert for TDS.

legend-engine-test-runner-service/src/test/resources/testable/legend-testable-service-M2M.pure
MauricioUyaguari: yeah right. we do. removing it thanks.

legend-engine-test-runner-service/src/main/java/org/finos/legend/engine/testable/service/assertion/ServiceTestAssertionEvaluator.java
MauricioUyaguari: The issue with this is that we give a false positive and the test result comes back as `TestPass` if there is no implementation for that assertion type yet. We could have multiple visitors but after going through the core assertion types and all the extra ones if its still null then I think we should throw an error and treat it as a `TestError`.  

Faced this issue when running the test through studio and hadn't finished writing assertToTDS.

legend-engine-xt-relationalStore-pure/src/main/resources/core_relational/relational/metamodel/test/relationalTest.pure
MauricioUyaguari: was going to have a default schema as `default` if they don't provide one kind of how we do for the store modeling in pure

legend-engine-xt-relationalStore-executionPlan/src/main/java/org/finos/legend/engine/plan/execution/stores/relational/test/RelationalConnectionFactory.java
MauricioUyaguari: I think we get this for free when we use the type `LocalH2DatasourceSpecification` for the relational database connection. This is currently how we run the relational mapping tests.

legend-engine-test-runner-service/src/main/java/org/finos/legend/engine/testable/service/assertion/RelationalTestAssertionHelper.java
MauricioUyaguari: @hardikmaheshwari Rethinking it, I dont think we should expose this to test. I think the asserts should take care of this. 
If use is using EqualToTDS then we can use the TDS serialization. For now we will support EqualToTDS with json input. We can further improve this by having user type out CSV but for now I think it is fairly clear for the user this way. 
Thoughts?

legend-engine-xt-relationalStore-grammar/pom.xml
MauricioUyaguari: The dependencies are scoped for only tests. If they do lead to cyclic dependency we can move them out to its own module but for now I think its fine.

legend-engine-xt-relationalStore-pure/src/main/resources/core_relational/relational/metamodel/test/relationalTest.pure
MauricioUyaguari: moved this just to csv.

legend-engine-xt-relationalStore-protocol/src/main/java/org/finos/legend/engine/protocol/pure/v1/model/packageableElement/store/relational/data/RelationalDataTableRow.java
MauricioUyaguari: moved this just to csv value

legend-engine-test-runner-service/src/main/java/org/finos/legend/engine/testable/service/assertion/ServiceTestAssertionEvaluator.java
hardikmaheshwari: Lets throw this error where we are filtering nuls

legend-engine-xt-relationalStore-executionPlan/src/main/java/org/finos/legend/engine/plan/execution/stores/relational/test/RelationalConnectionFactory.java
```diff
@@ -50,4 +57,25 @@ public Optional<Connection> tryBuildFromConnection(Connection connection, String
...
+    {
+        if(data instanceof RelationalCSVData && sourceConnection instanceof RelationalDatabaseConnection)
+        {
+            RelationalDatabaseConnection connection = new RelationalDatabaseConnection();
+            connection.databaseType = DatabaseType.H2;
+            connection.type = DatabaseType.H2;
+            connection.element = sourceConnection.element;
+            connection.authenticationStrategy = new TestDatabaseAuthenticationStrategy();
+            LocalH2DatasourceSpecification localH2DatasourceSpecification = new LocalH2DatasourceSpecification();
+            // TODO generate sql with pure helper function
```
hardikmaheshwari, line 72: Lets boot up H2 here itself

MauricioUyaguari: Looking at the code for relational, the difference is we already have a dedicated H2LocalServer up. 
See: https://github.com/finos/legend-engine/blob/master/legend-engine-xt-relationalStore-executionPlan-connection/src/main/java/org/finos/legend/engine/plan/execution/stores/relational/H2LocalServer.java
There is no need to boot up a specific special server for relational.

MauricioUyaguari:
@hardikmaheshwari updated with the service test changes we discussed.

pierredebelen (approved):

PR #674, opened 2022-05-10
fix protocols transformation of function Class parameters

davidharte-gs:
Harmonize Engine compilation with Pure for f: ::new(..). Included test for roundtrip from Pure Code->PMCD->Pure Code, using Pure transformation.


davidharte-gs:
depends on: https://github.com/finos/legend-pure/pull/560

PR #697, opened 2022-05-25
Stricter enforce rule on execution and generation collection modules

rafaelbey:
This prevents logback and dropwizard as a dependency on the generation and execution modules that gitlab projects will depend on.  This also slightly improve how enforcer required java version is defined to ensure all definitions are using the same value.


rafaelbey:
@kevin-m-knight-gs - As discussed last week, this is part of the dependency cleanup of DropWizard and Logback.  Let me know if there are other dependency we need to ban, any other place we should do similar uplift of enforcer.

PR #731, opened 2022-06-09
Fix JSON to Grammar transformation when Applied functions are passed as paramters.

vj2050gs:
Fix JSON to grammar transformation when the first parameter to a function is an instance of Applied Function. Ensured correct grammar generation when using eq (equals) applied function by avoiding the `->` syntax.


rafaelbey (approved):

davidharte-gs (changes requested):
As discussed, please review the testing strategy, it seems that the tests in TestJsonToGrapparAPi.java may be a subset of those in TestDomainGrammarRoundTrip. Thx

PR #738, opened 2022-06-14
create module for sqlServer db extension

vikask-gs:


vikask-gs:
> Shouldn't the new module be: legend-engine-xt-relationalStore-pure-sqlServer

We may want to add connection acquisition code in module for each db. Hence I skipped the 'pure' part.

PR #756, opened 2022-06-21
Mastery spec creation of xt-mastery-pure

christian-gallagher:


christian-gallagher:
Not intended to pull onto Finos master yet.

PR #759, opened 2022-06-21
Support function overloading

YannanGao-gs:
**What type of PR is this?**
Feature Request

**What does this PR do / why is it needed ?**
update the path of Function from package::name to package::function signature

1. Update Engine parser so that the function signature will be used for function.name
2. SDLC identifies the old style (use functionName as the file name) functions, then creates a new entity that has functionName as its path  
3. Studio will create a ‘Delete’ action to delete this kind of entity above, then do a ‘Create’ operation to correct the gitlab file name
 

**Which issue(s) this PR fixes:**
Fixes #

**Other notes for reviewers:
Does this PR introduce a user-facing change?**
Related to https://github.com/finos/legend-sdlc/pull/515 and https://github.com/finos/legend-studio/pull/1339
Require Alloy changes

### DIFF
--- a/legend-engine-language-pure-compiler/src/main/java/org/finos/legend/engine/language/pure/compiler/toPureGraph/HelperModelBuilder.java
+++ b/legend-engine-language-pure-compiler/src/main/java/org/finos/legend/engine/language/pure/compiler/toPureGraph/HelperModelBuilder.java
@@ -252,15 +252,7 @@ public class HelperModelBuilder
      */
     public static String getSignature(Function function)
     {
-        return getFunctionNameWithoutSignature(function) + getFunctionSignatureSuffix(function);
-    }
-
-
-    private static String getFunctionSignatureSuffix(Function function)
-    {
-        return "_" + LazyIterate.collect(function.parameters, p -> p._class != null ? p._class + "_" + getMultiplicitySignature(p.multiplicity) : null).select(Objects::nonNull).makeString("__")
-                // TODO: do we have to take care of void return type ~ Nil?
-                + "__" + function.returnType + "_" + getMultiplicitySignature(function.returnMultiplicity) + "_";
+        return getFunctionNameWithoutSignature(function) + terseSignatureSuffix(function);
     }
 
     public static String getFunctionNameWithoutSignature(Function function)
@@ -279,10 +271,10 @@ public class HelperModelBuilder
 
     private static String terseSignatureSuffix(Function function)
     {
-        return "_" + LazyIterate.collect(function.parameters, HelperModelBuilder::getParameterSignature).select(Objects::nonNull).makeString("__")
+        String functionSignature =  LazyIterate.collect(function.parameters, HelperModelBuilder::getParameterSignature).select(Objects::nonNull).makeString("__")
                 // TODO: do we have to take care of void return type ~ Nil?
                 + "__" + getClassSignature(function.returnType) + "_" + getMultiplicitySignature(function.returnMultiplicity) + "_";
-
+        return function.parameters.size() > 0 ? "_" + functionSignature : functionSignature;
     }
 
 
--- a/legend-engine-language-pure-grammar/src/main/java/org/finos/legend/engine/language/pure/grammar/from/domain/DomainParseTreeWalker.java
+++ b/legend-engine-language-pure-grammar/src/main/java/org/finos/legend/engine/language/pure/grammar/from/domain/DomainParseTreeWalker.java
@@ -24,6 +24,7 @@ import org.eclipse.collections.api.list.MutableList;
 import org.eclipse.collections.impl.factory.Lists;
 import org.eclipse.collections.impl.list.mutable.FastList;
 import org.eclipse.collections.impl.utility.Iterate;
+import org.eclipse.collections.impl.utility.LazyIterate;
 import org.eclipse.collections.impl.utility.ListIterate;
 import org.finos.legend.engine.language.pure.grammar.from.ParseTreeWalkerSourceInformation;
 import org.finos.legend.engine.language.pure.grammar.from.ParserErrorListener;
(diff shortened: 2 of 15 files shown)


legend-engine-language-pure-grammar/src/test/java/org/finos/legend/engine/language/pure/grammar/test/roundtrip/TestDomainGrammarRoundtrip.java
```diff
@@ -596,7 +596,7 @@ public void testMetaNewFunctionWithSingleParameter()
     @Test
     public void testMultiIFExpressions()
     {
-        test("function test::multiExpressions(): String[0..1]\n" +
```
MauricioUyaguari, line 599: 💀 ☠️

MauricioUyaguari (changes requested):

YannanGao-gs:
Need to fix broken tests

legend-engine-language-pure-grammar/src/main/java/org/finos/legend/engine/language/pure/grammar/from/domain/DomainParseTreeWalker.java
```diff
@@ -395,6 +397,10 @@ private org.finos.legend.engine.protocol.pure.v1.model.packageableElement.domain
         func.returnType = ctx.functionTypeSignature().type().getText();
         func.returnMultiplicity = this.buildMultiplicity(ctx.functionTypeSignature().multiplicity().multiplicityArgument());
         func.sourceInformation = this.walkerSourceInformation.getSourceInformation(ctx);
+        String functionSignature =   LazyIterate.collect(func.parameters, HelperValueSpecificationGrammarComposer::getParameterSignature).select(Objects::nonNull).makeString("__")
```
MauricioUyaguari, line 400: Is this change needed because when going from text to pmcd we can't have 2 elements with the same package and name?
Do you know where that constraint comes from?
My thinking is that the grammar/protocol should not have to handle this. 
It is when we compile (protocol -> metamodel)  the function we need to provide and generate the full function name.

legend-engine-language-pure-compiler/src/main/java/org/finos/legend/engine/language/pure/compiler/toPureGraph/HelperModelBuilder.java
```diff
@@ -279,10 +271,10 @@ public static String getTerseSignature(Function function)
 
     private static String terseSignatureSuffix(Function function)
     {
-        return "_" + LazyIterate.collect(function.parameters, HelperModelBuilder::getParameterSignature).select(Objects::nonNull).makeString("__")
+        String functionSignature =  LazyIterate.collect(function.parameters, HelperModelBuilder::getParameterSignature).select(Objects::nonNull).makeString("__")
```
MauricioUyaguari, line 274: Could we make the logic more clear here. Maybe use string builders. Also lets add tests for these functions to make it clear what the expectation that each returns.

legend-engine-language-pure-grammar/src/main/java/org/finos/legend/engine/language/pure/grammar/from/domain/DomainParseTreeWalker.java
```diff
@@ -395,6 +397,10 @@ private org.finos.legend.engine.protocol.pure.v1.model.packageableElement.domain
...
+            + "__" + HelperValueSpecificationGrammarComposer.getClassSignature(func.returnType) + "_" + HelperValueSpecificationGrammarComposer.getMultiplicitySignature(func.returnMultiplicity) + "_";
+        String terseSignatureSuffix =  func.parameters.size() > 0 ? "_" + functionSignature : functionSignature;
+        func.name = PureGrammarParserUtility.fromIdentifier(ctx.qualifiedName().identifier()) + terseSignatureSuffix;
```
MauricioUyaguari, line 403: lets move this logic out and call it like `buildFullFunctionSignature` to make it clear we are assigning the full function name to the .name property.

MauricioUyaguari: Dont we also have already defined methods we could use here like `getFunctionSignature` and getFunctionName ?

legend-engine-language-pure-compiler/src/main/java/org/finos/legend/engine/language/pure/compiler/toPureGraph/HelperModelBuilder.java
YannanGao-gs: tests are added

akphi:
@YannanGao-gs committing .DS_Store file (system files) is a no-no. `engine` `gitignore` should prevent you from doing this, but it didn't. But you should remove that from the commit

akphi (approved):
Looks good to me

akphi (changes requested):
Except for the problem with `DS_Store` file mentioned.

legend-engine-language-pure-grammar/src/main/java/org/finos/legend/engine/language/pure/grammar/to/HelperValueSpecificationGrammarComposer.java
```diff
@@ -250,23 +251,26 @@ public static String printFullPath(String fullPath, DEPRECATED_PureGrammarCompos
     public static String getFunctionName(org.finos.legend.engine.protocol.pure.v1.model.packageableElement.domain.Function fn)
     {
         int signatureIndex = fn.name.indexOf(getFunctionSignature(fn));
-        String name = signatureIndex > 0 ? fn.name.substring(0, signatureIndex) : fn.name;
-        return fn._package == null || fn._package.isEmpty() ? name : fn._package + "::" + name;
+        if (signatureIndex <= 0)
```
MauricioUyaguari, line 254: let's remove this throw for now.

pierredebelen (approved):

PR #776, opened 2022-06-30
add postgres Sql Generation Tests

PrateekGarg-gs:
add postgres Sql Generation Test Runner and github workflow


vikask-gs:
Looks good

PR #788, opened 2022-07-07
Implements Adaptive Batching in Graph Fetch

### LINKED ISSUE #494: Graph query execution with adaptive batching
## Feature Request

Legend engine provides batch execution capability for graph fetch queries. Batch execution capability lets the engine perform "graphFetch" operation (fetch the nested property structure as per the query) on a list of objects in one go.

This feature significantly improves performance of graph queries, specially in the case of relational stores where the cost of executing a SQL query is high. Let's consider a situation where a graph fetch operation is being performed on 2000 objects and the fetch tree requires 5 SQLs to be executed. Without batching, the number of SQL executions would be (2000 * 5 = 10000). The same graph fetch operation when done with a batch size of 1000 would require (2 * 5 = 10) SQL executions which is a lot less (2 is the number of batches). However, executing graph fetch with batching would increase the memory requirement from server perspective.

### Description of Problem:
Currently, batch size is configured in the query as an [extra parameter to graph fetch operation](https://github.com/finos/legend-pure/blob/master/legend-pure-code-compiled-core/src/main/resources/core/pure/graphFetch/graphFetch.pure#L26). It remains constant during the execution. Also, there is a memory limit which is put up to cap the memory being consumed for each batch in the execution (can result in failures if memory limit is breached). It requires some effort to tune the batch size and arrive at a good number, by doing a tradeoff between performance and memory utilization.

### Potential Solution:
One of the ways we can tackle this problem is to build an adaptive batching mechanism. Recently, some extra stats were added to graph fetch execution in this PR: https://github.com/finos/legend-engine/pull/492. The idea is to use the stats from previous batches to decide on the batch size for next batch thereby adapting to the dynamics of the execution (increasing batch size if the memory consumption is less compared to the memory limit or reducing batch size if memory consumption exceeds or is close to the memory limit). This feature can help improve performance in the case of executions which are less memory intensive and also avoid failures in executions which are memory intensive.

RIsHaV-1024:
The batch size in a graph fetch query was earlier fixed to 1000. Now the batch size can vary depending on the memory utilization of the previous batch. Resolves #494

### DIFF
--- a/legend-engine-executionPlan-execution/src/main/java/org/finos/legend/engine/plan/execution/PlanExecutor.java
+++ b/legend-engine-executionPlan-execution/src/main/java/org/finos/legend/engine/plan/execution/PlanExecutor.java
@@ -58,6 +58,8 @@ import java.util.stream.Collectors;
 public class PlanExecutor
 {
     public static final long DEFAULT_GRAPH_FETCH_BATCH_MEMORY_LIMIT = 52_428_800L; /* 50MB - 50 * 1024 * 1024 */
+    public static final long DEFAULT_GRAPH_FETCH_SOFT_MEMORY_LIMIT_PERCENTAGE = 50;
+    public static final boolean DEFAULT_USE_ADAPTIVE_BATCHING = true;
     public static final String USER_ID = "userId";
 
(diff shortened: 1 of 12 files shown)


hardikmaheshwari:
@cla-bot[bot] check

epsstan:
This PR has not been updated in a while and has merge conflicts. Closing this PR for now.

PR #800, opened 2022-07-13
Improve primitive property type check for mapping algorithm

YannanGao-gs:
loose primitive property type check for mapping algorithm


akphi (approved):

MauricioUyaguari (approved):

legend-engine-xt-analytics-mapping-pure/src/main/resources/core_analytics_mapping/modelCoverageAnalyticsTest.pure
```diff
@@ -144,7 +144,7 @@ Class meta::analytics::mapping::modelCoverage::test::Street extends meta::analyt
 
 Class meta::analytics::mapping::modelCoverage::test::_LegalEntity
 {
-  id: String[1];
+  id: Decimal[1];
```
MauricioUyaguari, line 147: Could we add a property of type Decimal instead of altering an existing property

pierredebelen (approved):

PR #805, opened 2022-07-14
Part I - return multiplicity should govern Service's json response format

YannanGao-gs:
The current behavior of alloy service for graphFetch is inconsistent. For multiplicity one, it will return one object.
For multiplicity 0 or many, it will wrap results with brackets.

To be backward compatible, we introduced a flag in AlloyConfig.
Now,  Users could set this config to true to always wrap results with brackets.


hardikmaheshwari:
Can we please add few tests for this

YannanGao-gs:
> Can we please add few tests for this

Test was added

pierredebelen:
The current behavior of checking the returned data set cardinality is obviously wrong as not 'predictable'. I think we should use the function return multiplicity (inferred by the compiler) to dictate if we wrap the result in a [] or not.

finos-admin:

finos-admin:

PR #821, opened 2022-07-22
Refactor union and merge to use specialized nodes

hardikmaheshwari:
Refactor union and merge to use specialized nodes


gs-ssh16 (approved):

epsstan (approved):
Marking as approved as @gs-ssh16 has already approved this.

PR #829, opened 2022-07-26
fix bug for java generation of functions

gs-gunjan:
fix bug for java generation of functions


gs-gunjan:
will get this merged with function testable mechanism

PR #834, opened 2022-07-29
provide better error messaging for NPE in service test runner

MauricioUyaguari:
- generic NPE is thrown when no test data is provided for connection when running service tests

### DIFF
--- a/legend-engine-test-runner-service/src/main/java/org/finos/legend/engine/testable/service/connection/TestConnectionBuilder.java
+++ b/legend-engine-test-runner-service/src/main/java/org/finos/legend/engine/testable/service/connection/TestConnectionBuilder.java
@@ -65,7 +65,11 @@ public class TestConnectionBuilder implements ConnectionVisitor<Pair<Connection,
                 .select(Optional::isPresent)
                 .collect(Optional::get)
                 .getFirstOptional()
-                .orElseThrow(() -> new UnsupportedOperationException("Unsupported embedded data type '" + embeddedData.getClass().getSimpleName() + "\" with connection type '" + connection.getClass().getSimpleName() + '"'));
+                .orElseThrow(() -> new UnsupportedOperationException(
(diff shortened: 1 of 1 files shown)


legend-engine-test-runner-service/src/main/java/org/finos/legend/engine/testable/service/extension/ServiceTestRunner.java
```diff
@@ -251,7 +251,10 @@ private Pair<Runtime, List<Closeable>> getTestRuntimeAndClosableResources(Runtim
                         embeddedData = connectionTestData.data;
                     }
                 }
-
+                else
```
hardikmaheshwari, line 254: I don't think this is right place for this check. This would break model chain connection tests, right?
Does it not break not break any existing tests?

legend-engine-test-runner-service/src/main/java/org/finos/legend/engine/testable/service/connection/TestConnectionBuilder.java
```diff
@@ -51,6 +51,10 @@ public class TestConnectionBuilder implements ConnectionVisitor<Pair<Connection,
 
     public TestConnectionBuilder(EmbeddedData embeddedData, PureModelContextData pureModelContextData)
     {
+        if (embeddedData == null)
```
hardikmaheshwari, line 54: I don't think we can add this here, for example modelChainConnection code below don't require test data

legend-engine-test-runner-service/src/main/java/org/finos/legend/engine/testable/service/extension/ServiceTestRunner.java
MauricioUyaguari: yeah seems too :) 
Was looking into a bug and thought we hit a NPE regardless after this. Let me recheck that.

PR #843, opened 2022-07-29
Fix FreeMarker in SQL generation as part of execution plan for property With Path.

vj2050gs:
Fixed the FreeMarker syntax in SQL generation as part of execution plan for property With Path to generate a valid SQL.

### DIFF
--- a/legend-engine-xt-relationalStore-pure/src/main/resources/core_relational/relational/sqlQueryToString/dbExtension.pure
+++ b/legend-engine-xt-relationalStore-pure/src/main/resources/core_relational/relational/sqlQueryToString/dbExtension.pure
@@ -421,7 +421,7 @@ function meta::relational::functions::sqlQueryToString::convertPlaceHolderToSQLS
 function meta::relational::functions::sqlQueryToString::processPlaceHolder(v:VarPlaceHolder[1], type:Type[1], prefix:String[1], suffix:String[1], dbTimeZone:String[0..1]): String[1]
 {
    let noPropertyPath              = $v.propertyPath->isEmpty();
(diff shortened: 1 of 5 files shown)


rafaelbey (approved):

legend-engine-xt-relationalStore-pure/src/main/resources/core_relational/relational/sqlQueryToString/dbExtension.pure
```diff
@@ -421,7 +421,7 @@ function meta::relational::functions::sqlQueryToString::convertPlaceHolderToSQLS
 function meta::relational::functions::sqlQueryToString::processPlaceHolder(v:VarPlaceHolder[1], type:Type[1], prefix:String[1], suffix:String[1], dbTimeZone:String[0..1]): String[1]
 {
    let noPropertyPath              = $v.propertyPath->isEmpty();
-   let placeHolderWithPath         = $v.name+if($noPropertyPath,|'',|'[\"')+$v.propertyPath->map(p|$p.name)->joinStrings('.')+if($noPropertyPath,|'',|'\"]');
```
vj2050gs, line 424: Currently if we have propertyPath, the SQL that gets generated has structure ${endDateCalendar[\"fiscalYear.value\"]} 
As per the FM documentation [FreeMarkerDocs](https://freemarker.apache.org/docs/dgui_template_exp.html#dgui_template_exp_direct_hash) 
eg. supported---->  book.author.name, book["author"].name, book.author["name"], book["author"]["name"]. 
what we have in SQL is like book["author.name"]
Hence we need to change this syntax for FreeMarker to correctly parse and perform substitution.

pierredebelen (approved):

PR #846, opened 2022-08-02
Standardize compile error handling to avoid infinite loop when serializing

rafaelbey:
If EngineException cause is a PureExcetpion, jackson fails to serialize the response given self reference on some of the get methods.

### DIFF
--- a/legend-engine-language-pure-compiler-api/src/main/java/org/finos/legend/engine/language/pure/compiler/api/Compile.java
+++ b/legend-engine-language-pure-compiler-api/src/main/java/org/finos/legend/engine/language/pure/compiler/api/Compile.java
@@ -124,16 +124,9 @@ public class Compile
 
     private Response handleException(UriInfo uriInfo, MutableList<CommonProfile> profiles, long start, Exception ex)
     {
-        if (ex instanceof EngineException)
-        {
-            MetricsHandler.incrementErrorCount(uriInfo != null ? uriInfo.getPath() : null, Response.Status.BAD_REQUEST.getStatusCode());
-            return Response.status(Response.Status.BAD_REQUEST).entity(ex).build();
-        }
-        else
-        {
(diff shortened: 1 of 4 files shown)


legend-engine-language-pure-compiler-api/src/main/java/org/finos/legend/engine/language/pure/compiler/api/Compile.java
```diff
@@ -124,16 +124,9 @@ public Response lambdaReturnType(LambdaReturnTypeInput lambdaReturnTypeInput, @A
...
-        else
-        {
-            Response errorResponse = ExceptionTool.exceptionManager(ex, LoggingEventType.COMPILE_ERROR, profiles);
-            MetricsHandler.incrementErrorCount(uriInfo != null ? uriInfo.getPath() : null, errorResponse.getStatus());
-            return errorResponse;
-        }
+        Response.Status status = ex instanceof EngineException ? Response.Status.BAD_REQUEST : Response.Status.INTERNAL_SERVER_ERROR;
+        Response errorResponse = ExceptionTool.exceptionManager(ex, LoggingEventType.COMPILE_ERROR, status, profiles);
+        MetricsHandler.incrementErrorCount(uriInfo != null ? uriInfo.getPath() : null, errorResponse.getStatus());
+        return errorResponse;
```
rafaelbey, line 130: @akphi @MauricioUyaguari - wondering if the rational to handle EngineException differently is driven by Studio handling of response.  If this is correct, can you point me to the code where this is handle to check if this standardization will work or break Studio?

rafaelbey: Confirmed with Mauricio that Studio leverage message, sourceInformation, and loosely errorType.  The ExceptionError that we put on the response contains these fields.

pierredebelen (approved):

PR #856, opened 2022-08-06
Middle Tier authz : Integrate authorizer into execution code path

epsstan:
This PR introduces the authorizer into the execution code path. 

Also a new server integration tests module has been added. This module is not a dependency of other engine modules.


davidharte-gs (approved):
Approved

PR #867, opened 2022-08-11
Create and enable protocol version v1_25_0

aziemchawdhary-gs:
Create protocol version v1_25_0


aziemchawdhary-gs:
Will break internal codebase until corresponding protocol changes made internally

PR #882, opened 2022-08-17
refactor bigquery execution

vikask-gs:


epsstan (approved):
Approving as the original  PR #865 that introduced this change was already approved by @davidharte-gs

PR #895, opened 2022-08-29
Fix error message for test suite execution to avoid NPE

hardikmaheshwari:
This PR fixes the NPE issue while reporting an error for the test execution flow

No issue raised, ad hoc fix

Yes, this will fix NPE issue few users faced when they incorrectly structured their tests

### DIFF
--- a/legend-engine-test-runner-service/src/main/java/org/finos/legend/engine/testable/service/connection/TestConnectionBuilder.java
+++ b/legend-engine-test-runner-service/src/main/java/org/finos/legend/engine/testable/service/connection/TestConnectionBuilder.java
@@ -65,7 +65,13 @@ public class TestConnectionBuilder implements ConnectionVisitor<Pair<Connection,
                 .select(Optional::isPresent)
                 .collect(Optional::get)
                 .getFirstOptional()
-                .orElseThrow(() -> new UnsupportedOperationException("Unsupported embedded data type '" + embeddedData.getClass().getSimpleName() + "\" with connection type '" + connection.getClass().getSimpleName() + '"'));
+                .orElseThrow(() ->
+                {
+                    String errorMessage = embeddedData == null
+                            ? "No test data provided for connection type '" + connection.getClass().getSimpleName() + "\". Either you need to provide test data for the connection type or connection type is not supported."
+                            : "Unsupported test data type '" + embeddedData.getClass().getSimpleName() + "\" with connection type '" + connection.getClass().getSimpleName() + '"';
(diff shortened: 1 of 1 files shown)


MauricioUyaguari (approved):

legend-engine-test-runner-service/src/main/java/org/finos/legend/engine/testable/service/connection/TestConnectionBuilder.java
```diff
@@ -65,7 +65,13 @@ public Pair<Connection, List<Closeable>> visit(Connection connection)
                 .select(Optional::isPresent)
                 .collect(Optional::get)
                 .getFirstOptional()
-                .orElseThrow(() -> new UnsupportedOperationException("Unsupported embedded data type '" + embeddedData.getClass().getSimpleName() + "\" with connection type '" + connection.getClass().getSimpleName() + '"'));
+                .orElseThrow(() ->
+                {
```
epsstan, line 69: Can we please add a test ? We already have a test class - TestTestConnectionBuilder

legend-engine-test-runner-service/src/main/java/org/finos/legend/engine/testable/service/connection/TestConnectionBuilder.java
```diff
@@ -65,7 +65,13 @@ public Pair<Connection, List<Closeable>> visit(Connection connection)
...
+                    String errorMessage = embeddedData == null
+                            ? "No test data provided for connection type '" + connection.getClass().getSimpleName() + "\". Either you need to provide test data for the connection type or connection type is not supported."
```
epsstan, line 71: The quoting of the type name is incorrect. You have a opening single quote (before connection.getClass()) but closing double quote ?

epsstan (changes requested):

legend-engine-test-runner-service/src/main/java/org/finos/legend/engine/testable/service/connection/TestConnectionBuilder.java
hardikmaheshwari: Done

legend-engine-test-runner-service/src/main/java/org/finos/legend/engine/testable/service/connection/TestConnectionBuilder.java
hardikmaheshwari: Thanks for pointing this. Done

epsstan (approved):

PR #672, opened 2022-05-05
Invalidate usage of new function call in mapping

saisha95:

PR #897, opened 2022-08-30
Revert "Invalidate usage of new function call in mapping (#672)"

aziemchawdhary-gs:
This reverts commit 2251c6ce25466a277fd85516c10e6394eb5d5070.


epsstan (approved):
PR #672 seems to have introduced a bug and more general performance degradation.

PR #904, opened 2022-09-06
Enable server to read configuration via environment variables

epsstan:
#### What type of PR is this?

Enhancement

This PR enables the engine serve to source its configuration from environment variables.

This is useful in cases where the runtime configuration is dynamic and injected as part of the runtime environment. E.g. instead of hard coding the Gitlab OAuth client id and secret in the server's config file, they can be injected into the runtime (e.g. Docker) as env variables. The server config file can then refer to these environment variables.

Not applicable.

No

### DIFF
--- a/legend-engine-server/pom.xml
+++ b/legend-engine-server/pom.xml
@@ -419,6 +419,10 @@
             <artifactId>jetty-server</artifactId>
         </dependency>
         <dependency>
+            <groupId>io.dropwizard</groupId>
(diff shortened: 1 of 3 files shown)


akphi (approved):

akphi:
Example of the usage

https://stackoverflow.com/questions/23464451/overriding-server-connector-config-with-env-variables-with-dropwizard

if in the `config` file, we have `${FOO}` at runtime, `DropWizard` will look up for `FOO` variable to replace this

PR #909, opened 2022-09-07
SQL and FTL injection fix

mateusz-klatt-gs:
- Use escapeSQL to create safe string in SQL query
- Do not resolve FTL templates recursively
- Do not replace @@ with ' in string literals

### DIFF
--- a/legend-engine-application-query/pom.xml
+++ b/legend-engine-application-query/pom.xml
@@ -19,7 +19,7 @@
     <parent>
         <groupId>org.finos.legend.engine</groupId>
         <artifactId>legend-engine</artifactId>
-        <version>3.16.0</version>
+        <version>3.16.1-SNAPSHOT</version>
     </parent>
     <modelVersion>4.0.0</modelVersion>
     <artifactId>legend-engine-application-query</artifactId>
--- a/legend-engine-configuration/pom.xml
+++ b/legend-engine-configuration/pom.xml
@@ -19,7 +19,7 @@
     <parent>
         <groupId>org.finos.legend.engine</groupId>
         <artifactId>legend-engine</artifactId>
-        <version>3.16.0</version>
+        <version>3.16.1-SNAPSHOT</version>
     </parent>
     <modelVersion>4.0.0</modelVersion>
     <artifactId>legend-engine-configuration</artifactId>
--- a/legend-engine-executionPlan-dependencies/pom.xml
+++ b/legend-engine-executionPlan-dependencies/pom.xml
@@ -19,7 +19,7 @@
     <parent>
         <groupId>org.finos.legend.engine</groupId>
         <artifactId>legend-engine</artifactId>
-        <version>3.16.0</version>
+        <version>3.16.1-SNAPSHOT</version>
     </parent>
     <modelVersion>4.0.0</modelVersion>
     <artifactId>legend-engine-executionPlan-dependencies</artifactId>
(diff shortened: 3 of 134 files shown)


legend-engine-executionPlan-execution/src/main/java/org/finos/legend/engine/plan/execution/nodes/helpers/freemarker/FreeMarkerExecutor.java
```diff
@@ -66,12 +67,19 @@ public static String process(String input, ExecutionState executionState, String
 
     public static String processRecursively(String input, Map<String, ?> variableMap, String templateFunctions)
     {
-        String result = process(input, variableMap, templateFunctions);
-        if (!result.equals(input.replace("\\\"", "\"")))
+        return processRecursivelyInternal(input, variableMap.entrySet().stream().collect(Collectors.toMap(Map.Entry::getKey, Map.Entry::getValue)), templateFunctions);
```
aziemchawdhary-gs, line 70: Could you extract out the second argument (the deep copy of variableMap) into a variable for readability?
Also might be worth adding a comment to say why you need to do a deep copy of variableMap

legend-engine-executionPlan-execution/src/main/java/org/finos/legend/engine/plan/execution/result/ResultNormalizer.java
```diff
@@ -71,7 +71,7 @@ public static Object normalizeToSql(Object o, String databaseTimeZone)
 
         if (o instanceof String)
         {
-            return ((String) o).replace("'", "\'").replace("\\", "\\\\");
+            return ((String) o).replace("\\", "\\\\");
```
aziemchawdhary-gs, line 74: Does the test TestResultNormalizer.testStringNormalization still pass with this change? Is the replace that is removed happening somewhere else now?

legend-engine-xt-relationalStore-executionPlan/src/main/java/org/finos/legend/engine/plan/execution/stores/relational/RelationalExecutor.java
```diff
@@ -358,24 +355,11 @@ private Connection getConnection(DatabaseConnection databaseConnection, String o
...
-            Template t = new Template("sqlTemplate", new StringReader(templateFunctions + "\n" + query), cfg);
-            StringWriter stringWriter = new StringWriter();
-            t.process(vars, stringWriter);
-            result = stringWriter.toString();
-        }
-        catch (Exception ignored)
-        {
-        }
-        return result;
+        return FreeMarkerExecutor.processRecursively("\n" + query, vars.entrySet().stream().collect(Collectors.toMap(e -> String.valueOf(e.getKey()), Map.Entry::getValue)), templateFunctions);
```
aziemchawdhary-gs, line 358: For readability could we extract the second argument out into a variable?

legend-engine-executionPlan-execution/src/main/java/org/finos/legend/engine/plan/execution/result/ResultNormalizer.java
mateusz-klatt-gs: replace("'", "\'") does nothing - as \' string literal is the same as ' in java - what was probably suppose to be here is replace("'", "\\'") - but this would break escaping as it would escape everything including the FreeMarker variables

PR #919, opened 2022-09-12
Enum as Service Parameter

mrudula-gs:
Users can have enum as service parameter for Alloy services.
This PR will cover the EQUAL, NOT EQUAL, IN, NOT IN operations


davidharte-gs (approved):
Approved

legend-engine-executionPlan-execution/src/test/java/org/finos/legend/engine/plan/execution/validation/test/TestParametersValidation.java
```diff
@@ -202,7 +203,7 @@ private void testRequiredToOneParameter(String cls, List<?> validValues, List<?>
         for (Object validValue : validValues)
         {
             ExecutionState state = newExecutionState(parameter.name, validValue);
-            FunctionParametersParametersValidation.validate(Lists.immutable.with(parameter), state);
+            FunctionParametersParametersValidation.validate(Lists.immutable.with(parameter), new ArrayList<>(), state);
```
rafaelbey, line 206: For empty collections, we should either use eclipse facotories, or jdk Collections utility methods, and avoid specific collection implementations constructors.  

```java
Collections.emptyList();
Lists.mutable.empty();
```

legend-engine-executionPlan-execution/src/main/java/org/finos/legend/engine/plan/execution/validation/FunctionParametersParametersValidation.java
```diff
@@ -67,12 +70,20 @@ private static void validateParameterValues(RichIterable<Variable> externalParam
...
     }
 
-    public static ValidationResult validate(Variable var, Object value)
+    public static ValidationResult validate(Variable var, List<ValidationContext> validationContext, Object value)
     {
         FunctionParameterTypeValidator validator = FunctionParameterTypeValidator.externalParameterTypeValidator(var._class);
         if (validator == null)
         {
-            return ValidationResult.errorValidationResult("Unknown external parameter type: " + var._class + ", valid external parameter types: " + FunctionParameterTypeValidator.getExternalParameterTypes().makeString("[", ", ", "]"));
+            ValidationContext enumContext = validationContext.stream().filter(v -> (v.getClass() == EnumValidationContext.class)).filter(e -> ((EnumValidationContext) e).varName.equals(var.name)).findFirst().orElse(null);
```
rafaelbey, line 78: We should instanceOf rather than class equality.  On some circumstances where different class loaders re used, this could be false while instanceof will work regardless of class loader.

I also wonder if the visitor method added to the context abstract class could have help with encapsulating some of this logic.

legend-engine-executionPlan-execution/src/main/java/org/finos/legend/engine/plan/execution/validation/FunctionParametersParametersValidation.java
```diff
@@ -67,12 +70,20 @@ private static void validateParameterValues(RichIterable<Variable> externalParam
...
+            if (enumContext != null)
+            {
+                return (((EnumValidationContext)enumContext).validEnumValues.contains(value.toString())) ? ValidationResult.successValidationResult() : ValidationResult.errorValidationResult("Value " + value + " is not a valid enum value for " + var._class);
```
rafaelbey, line 81: Should we add the valid enum values to the error message, so users knows what is accepted?

rafaelbey (approved):

Issue #920, opened 2022-09-12
Commit batch ingest library for use in persistence modules


davidsunglee:
### Description of Problem:
We are introducing a common library to handle batch ingestion logic with an initial focus on batch milestoning into relational databases. This library consolidates the implementation across multiple backend persistence implementations. It also provides an in-memory implementation that will be of use for fast testing.

finos-admin:
This issue is stale because it has been open for 30 days with no activity. Please remove stale label or add any comment to keep this open. Otherwise this will be closed in 5 days.

finos-admin:
This issue was closed because it has been inactive for 35 days. Please re-open if this issue is still relevant.

PR #921, opened 2022-09-12
Introduce persistence milestoning library logical plan.

davidsunglee:
#### What type of PR is this?
Feature.

We are introducing a common library to handle batch ingestion logic with an initial focus on batch milestoning into relational databases. This library consolidates the implementation across multiple backend persistence implementations. It also provides an in-memory implementation that will be of use for fast testing.

This PR introduces the first component of the solution above: logical plan.

None. But this is the first of a series of PRs addressing #920

No.

### DIFF
--- a/legend-engine-xt-persistence-component/legend-engine-xt-persistence-component-logical-plan/pom.xml
+++ b/legend-engine-xt-persistence-component/legend-engine-xt-persistence-component-logical-plan/pom.xml
@@ -1,0 +1,38 @@
+<?xml version="1.0" encoding="UTF-8"?>
+<!--
+ Copyright 2022 Goldman Sachs
+
+ Licensed under the Apache License, Version 2.0 (the "License");
+ you may not use this file except in compliance with the License.
+ You may obtain a copy of the License at
+
+      http://www.apache.org/licenses/LICENSE-2.0
+
+ Unless required by applicable law or agreed to in writing, software
+ distributed under the License is distributed on an "AS IS" BASIS,
+ WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ See the License for the specific language governing permissions and
+ limitations under the License.
+-->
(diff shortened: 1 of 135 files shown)


legend-engine-xt-persistence-component/legend-engine-xt-persistence-component-logical-plan/pom.xml
```diff
@@ -0,0 +1,38 @@
...
+         xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance"
+         xsi:schemaLocation="http://maven.apache.org/POM/4.0.0 http://maven.apache.org/xsd/maven-4.0.0.xsd">
+    <parent>
+        <groupId>org.finos.legend.engine</groupId>
+        <artifactId>legend-engine-xt-persistence-component</artifactId>
+        <version>3.15.4-SNAPSHOT</version>
+    </parent>
+    <modelVersion>4.0.0</modelVersion>
+
+    <artifactId>legend-engine-xt-persistence-component-logical-plan</artifactId>
```
rafaelbey, line 27: Wonder if all packages should be prefixed with 'logical plan' to ensure no conflicts with future components?

davidsunglee: Makes sense to have a prefix for namespacing.

'component' is too generic, so we'll replace that with something more descriptive (e.g. ingestion, milestoning, etc).

Will do it in one shot, once all related modules are pushed (targeting end of the week).

davidsunglee: And your instinct is right, we'll have a lineage component in the near future.

rafaelbey (approved):

davidsunglee:
@rafaelbey @epsstan Can you merge when you have a moment? I don't have authorization ... thanks!

PR #923, opened 2022-09-13
Updating Dockerfile

ivan-kyosev-gs:


epsstan (approved):
This PR disables the docker workflow from being globally run for all PRs.
Instead the action is run only if the PR changes the Dockerfile. 

We are doing this because many of the CVEs do not have simple fixes and all the PRs are reported as failing a PR check.

PR #924, opened 2022-09-13
Fix for javadoc that references classes produced by annotation processor (#921)

davidsunglee:
Maven javadoc plugin requires additional configuration to resolve references to classes produced by annotation processors. PR #921 introduced use of Immutables (limited to persistence modules only) which triggered this limitation.

Resolves a bug introduced by work on #920

None.

No.


rafaelbey:
Thanks @davidsunglee for confirming this generate the same javadoc as before.

rafaelbey (approved):

PR #941, opened 2022-09-16
Split deploy task into two commands to workaround maven-jar-plugin bug.

davidsunglee:
#### What type of PR is this?
Bug fix

This PR splits our deployment action into two separate maven commands.
- The first does an install on the project, solely to kick-off APT code generation.
- The second generates javadoc (which would fail without APT code generation) and performs deploy

These commands were previously smushed into a single command, but this triggered a bug in maven-jar-plugin which fails when executing JAR twice in a build for a module with no source files.

None

None

No

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -71,7 +71,9 @@ jobs:
           DOCKER_USERNAME: finos
           DOCKER_PASSWORD: ${{ secrets.DOCKER_PASSWORD }}
           MAVEN_OPTS: -Xmx4g
-        run: mvn install javadoc:javadoc deploy -P docker-snapshot
+        run: |
+          mvn install -DskipTests=true
+          mvn javadoc:javadoc deploy -P docker-snapshot
       - name: Upload Test Results
         if: always()
         uses: actions/upload-artifact@v3


epsstan (approved):
Excellent PR description. Thanks !

akphi:
I wonder if this solves the problem we had earlier with sonar plugins as well

davidsunglee:
@akphi I believe `install` and `deploy` both execute tests, so the two commands should result in one round of testing. Will verify in the logs.

akphi:
Gotcha! I guess this makes our build easier to reason about, the downside is we duplicate `install` steps which increases build time ~20mins I guess.

By the way, @epsstan I think we might be able to re-enable `sonar` (disabled by https://github.com/finos/legend-engine/pull/767)

davidsunglee:
@akphi @epsstan The deploy run post this change was notably longer than previous runs. If this continues, we can minimize run time by triggering `compile` on three individual persistence modules rather than `install` across the project.

That's the minimum required for javadoc to work, but @rafaelbey understandably concerned the build file becomes less maintainable. Also, those modules will eventually be moved out.

Will take a look at what caused the increased runtime.

When duplicate `install` steps run in the same command, many of the heavy-hitters are skipped:
- compile is skipped
- test execution is skipped

Will have to see if compile is still skipped when running sequentially.

akphi:
Another thing we could do is just to split the github actions:
1. Test CI: `mvn install` (run install and test): run our tests
2. Deploy CI: `mvn install -DSkipTests && mvn deploy javadoc:javadoc -DskipTests`

If we can do what @davidsunglee suggested in step 2, we technically can bring the deploy build time down significantly.

PR #942, opened 2022-09-16
Include default configuration files in Docker image

epsstan:
#### What type of PR is this?

Enhancement

This PR adds a set of default configuration files into the server Docker image. 

This simplifies the running of the Docker images. Users can simply run the container by passing configuration values as environment variables instead of having to manage the configuration files. 

No

### DIFF
--- a/legend-engine-server/Dockerfile
+++ b/legend-engine-server/Dockerfile
@@ -14,7 +14,7 @@
 
 FROM openjdk:11.0.16-bullseye
 COPY target/legend-engine-server-*.jar /app/bin/
-COPY target/classes/docker/config /config
+COPY src/main/resources/docker/config /config
 CMD java -cp /app/bin/*-shaded.jar \
 -XX:+ExitOnOutOfMemoryError \
 -XX:MaxRAMPercentage=60 \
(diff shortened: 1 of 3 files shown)


legend-engine-server/Dockerfile
```diff
@@ -14,6 +14,7 @@
 
 FROM openjdk:11.0.16-bullseye
 COPY target/legend-engine-server-*.jar /app/bin/
+COPY target/classes/docker/config /config
```
epsstan, line 17: @nsprimmer asks that we explicitly list each file to be copied

legend-engine-server/src/main/resources/docker/config/config.json
```diff
@@ -0,0 +1,80 @@
...
+    "pure": {
+      "host": "127.0.0.1",
+      "port": 8090
+    },
+    "alloy": {
+      "host": "127.0.0.1",
+      "port": 8090,
+      "prefix": "/depot/api"
+    }
+  },
```
kevin-m-knight-gs, line 73: Don't you want to parameterize at least the ports here?

epsstan: > Don't you want to parameterize at least the ports here?

:) I knew you would spot this. I was going to parameterize them as part of refactoring the installers for Legend Query.  But I can parameterize now too.

rafaelbey (approved):

gs-ssh16 (approved):

PR #958, opened 2022-09-21
Remove parser error listener subtypes.

davidsunglee:
#### What type of PR is this?
Cleanup.

It's currently possible to override the `ParserErrorListener.dereferenceTokens` method to provide more descriptive error messages to the user. This has led to a number of subtypes in persistence, mastery, and other modules.

This PR adds a constructor to `ParserErrorListener` that accepts a `Vocabulary` so subtyping is no longer necessary. It also deletes subtypes in legend-engine.

The `dereferenceTokens` method remains protected for now, so dependent projects have a chance to switch to the new constructor. It will be switched to private once the projects have been updated

None.

None.

No.


rafaelbey (approved):
Awesome!

PR #963, opened 2022-09-23
Add PersistenceTest model and PersistenceTestRunner flow

chloeminkyung:
PersistenceTest (AtomicTest)
- PersistenceTestBatches
- isTestDataFromServiceOutput (default: true)

PersistenceTestBatch (TestBatch)
- id
- TestData
- TestAssertions

PR #967, opened 2022-09-23
Tidy up persistence component and persistence test runner.

davidsunglee:
#### What type of PR is this?
Cleanup.

Make persistence component and persistence test runner more consistent with other Legend modules and related cleanup. Specifically:
- Fix multiplicity of TestBatch.batchId in pure model
- Use dependency management for version and scope control
- Fix newlines and line breaks
- Remove vestigial maven properties

None.

Cleanup for https://github.com/finos/legend-engine/pull/963.

No.

### DIFF
--- a/legend-engine-pure-code-compiled-core/src/main/resources/core/pure/test/testable.pure
+++ b/legend-engine-pure-code-compiled-core/src/main/resources/core/pure/test/testable.pure
@@ -31,7 +31,7 @@ Class <<typemodifiers.abstract>> meta::pure::test::AtomicTest extends Test
 Class <<typemodifiers.abstract>> meta::pure::test::TestBatch
 {
     id         : String[1];
-    batchId    : Integer[0..1];
+    batchId    : Integer[1];
     assertions : TestAssertion[*];
 }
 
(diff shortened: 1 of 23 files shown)


legend-engine-pure-code-compiled-core/src/main/resources/core/pure/test/testable.pure
```diff
@@ -31,7 +31,7 @@ Class <<typemodifiers.abstract>> meta::pure::test::AtomicTest extends Test
 Class <<typemodifiers.abstract>> meta::pure::test::TestBatch
 {
     id         : String[1];
-    batchId    : Integer[0..1];
+    batchId    : Integer[1];
```
rafaelbey, line 34: given this property will be required, should we change the protocol class to reflect this, and change from object Integer to primitive int?

https://github.com/finos/legend-engine/blob/master/legend-engine-protocol-pure/src/main/java/org/finos/legend/engine/protocol/pure/v1/model/test/TestBatch.java#L25

rafaelbey (approved):

legend-engine-pure-code-compiled-core/src/main/resources/core/pure/test/testable.pure
davidsunglee: Definitely makes sense. Will update.

PR #969, opened 2022-09-25
Enable mTLS for ServiceStore connections

epsstan:
#### What type of PR is this?

Enhancement

This PR allows engine to make ServiceStore connections with mTLS. 

This PR implements a tactical solution where if a JVM (e.g. engine) is configured with SSL key and trust stores, these key and trust stores are used to make HTTPS connections to services. 

This is a tactical solution which will be replaced by introducing mTLS as a ServiceStore SecurityScheme. 

This feature is introduced with a feature flag that can be enabled via the server's configuration. This feature is disabled by default.

N/A

No.


legend-engine-xt-serviceStore-executionPlan/pom.xml
```diff
@@ -175,6 +182,12 @@
             <artifactId>junit</artifactId>
             <scope>test</scope>
         </dependency>
+        <dependency>
```
hardikmaheshwari, line 185: this dependency has caused issues for us in past

hardikmaheshwari:
@epsstan I am afraid this seems a bit too hacky as essentially we are bypassing modelling store and achieving that via server config.

PR #972, opened 2022-09-26
Sql injection fix

mateusz-klatt-gs:
#### What type of PR is this?

SQL and FTL injection fix

Escape external variables with FreeMarker rendering, avoid recursion in FreeMarker rendering.

N/A

Some users may already escape variables before they reach engine, this change does escaping automatically, however as transition step strings that are un-escaped and escaped again produce the same output are considered already escaped.


finos-admin:

PR #978, opened 2022-09-27
support relational result sourcing

gs-gunjan:
We want to support sourcing relational result from one query into another as part of same execution block

NA

Sensitive changes to realizing relational objects

No

### DIFF
--- a/legend-engine-xt-relationalStore-executionPlan/src/main/java/org/finos/legend/engine/plan/execution/stores/relational/RelationalExecutor.java
+++ b/legend-engine-xt-relationalStore-executionPlan/src/main/java/org/finos/legend/engine/plan/execution/stores/relational/RelationalExecutor.java
@@ -69,6 +69,7 @@ import java.util.Map;
 import java.util.Optional;
 import java.util.ServiceLoader;
 import java.util.function.Consumer;
+import java.util.stream.Collectors;
 
 public class RelationalExecutor
 {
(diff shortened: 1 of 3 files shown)


legend-engine-xt-relationalStore-executionPlan/src/main/java/org/finos/legend/engine/plan/execution/stores/relational/RelationalExecutor.java
```diff
@@ -265,6 +266,15 @@ else if (var.getValue() instanceof PreparedTempTableResult && sqlQuery.contains(
             {
                 sqlQuery = sqlQuery.replace("(${" + var.getKey() + "})", ((PreparedTempTableResult) var.getValue()).getTempTableName());
             }
+            else if (var.getValue() instanceof RelationalResult && (sqlQuery.contains("inFilterClause_" + var.getKey() + "})") || sqlQuery.contains("${" + var.getKey() + "}")))
```
gs-ssh16, line 269: This check looks more like a hack

rafaelbey: As discussed with @gs-gunjan and @davidharte-gs - we will approve/merge this for the next release, and will continue looking into how we can enhance the code to be more robust, and be able to remove this hack.

rafaelbey (approved):

PR #981, opened 2022-09-28
fix multiplicity for pure platform expression node 

gs-gunjan:
This PR fixes a bug to set the multiplicity for pure platform expression nodes and in turn correctly build out execution plans

NA

NA
No


hardikmaheshwari:
Thanks for fixing.
Can we also fix it here at same time - https://github.com/finos/legend-engine/blob/master/legend-engine-pure-code-compiled-core/src/main/resources/core/store/platform/executionPlan/executionPlan_generation.pure

hardikmaheshwari (approved):

rafaelbey (approved):

PR #988, opened 2022-09-29
Add example of client/server using mTLS

epsstan:
#### What type of PR is this?

Enhancement

This PR adds test code and scripts that demonstrate a client and server communicating with mTLS.

Note : This is an addition to test code that will be used in future PRs for adding TLS and mTLS support in ServiceStore.

Fixes #

No


legend-engine-xt-serviceStore-executionPlan/src/test/java/org/finos/legend/engine/plan/execution/stores/service/utils/MTLSClientServerTest.java
```diff
@@ -0,0 +1,193 @@
...
+        CA_KEYSTORE_PATH = certs.caKeyStorePath;
+    }
+
+    private int port;
+    private WireMockServer wireMockServer;
+
+    @Before
+    public void setup()
+    {
+        this.port = DynamicPortGenerator.generatePort();
```
rafaelbey, line 71: Can we do port 0 and let wiremock pick one open port dynamically, and then use these APIs to find the port?

```java
int port = wireMock.port();
int httpsPort = wireMock.httpsPort();
```

legend-engine-xt-serviceStore-executionPlan/src/test/java/org/finos/legend/engine/plan/execution/stores/service/utils/MTLSClientServerTest.java
```diff
@@ -0,0 +1,193 @@
...
+        {
+            assumeTrue(false);
+        }
+        CertGenerator.Certs certs = new CertGenerator().generateCerts();
+        SERVER_KEYSTORE_PATH = certs.serverKeyStorePath;
+        CA_KEYSTORE_PATH = certs.caKeyStorePath;
+    }
+
+    private int port;
+    private WireMockServer wireMockServer;
```
rafaelbey, line 66: Is more efficient to start the server once per class rather than one per tests, and then reset the state and mock between test cases.  Anything that will not allow that approach for these test cases?

Also, we should consider using the junit rule that does some of the state management like guaranteeing the stop.

legend-engine-xt-serviceStore-executionPlan/pom.xml
```diff
@@ -115,6 +115,13 @@
         </dependency>
         <!-- ECLIPSE COLLECTIONS -->
 
+        <!-- LOG -->
```
hardikmaheshwari, line 118: Since changes are in test only, should this have test scope?

legend-engine-xt-serviceStore-executionPlan/src/test/java/org/finos/legend/engine/plan/execution/stores/service/utils/WireMockServerWrapper.java
```diff
@@ -0,0 +1,90 @@
...
+    WireMockServerWrapper withCAKeyStore(String caKeyStorePath, String caKeyStorePassword)
+    {
+        this.caKeyStorePath = caKeyStorePath;
+        this.caKeyStorePassword = caKeyStorePassword;
+        return this;
+    }
+
+    public WireMockServer build()
+    {
+        WireMockConfiguration wireMockConfiguration = new WireMockConfiguration();
```
hardikmaheshwari, line 68: Can we use WireMockConfiguration directly, if I understand correctly we are using builder pattern here and possibly Config already provides that, what do you think?

legend-engine-xt-serviceStore-executionPlan/src/test/java/org/finos/legend/engine/plan/execution/stores/service/utils/ServiceStoreTestUtils.java
```diff
@@ -103,12 +103,22 @@ public static SingleExecutionPlan buildPlanForQuery(String grammar, String mappi
...
+        return executePlan(plan, Maps.mutable.empty(), planExecutor);
+    }
+
     public static String executePlan(SingleExecutionPlan plan)
     {
         return executePlan(plan, Maps.mutable.empty());
     }
 
     public static String executePlan(SingleExecutionPlan plan, Map<String, ?> params)
+    {
```
hardikmaheshwari, line 117: do we need these changes?

hardikmaheshwari (changes requested):

legend-engine-xt-serviceStore-executionPlan/src/test/java/org/finos/legend/engine/plan/execution/stores/service/utils/MTLSClientServerTest.java
hardikmaheshwari: @rafaelbey this seems like a great suggestion, will try to use this across other tests too

rafaelbey: @hardikmaheshwari per my other comment - we should also look into use the junit rule, so that we start the server  once per test class, and configure the stubs per test case.  If a class has multiple test cases, this improve the execution of the test class given the IO cost of starting the server up is only does once.

finos-admin:

PR #989, opened 2022-09-29
minor fixes for query API error response

akphi:
#### What type of PR is this?

Bug fix

This fixes a small bug with `Query APIs` where the error object is not a valid JSON object

_NA_

No


rafaelbey (approved):
Thanks!

PR #994, opened 2022-09-29
Apply fix for false realizeInMemory calls during plan execution (edited)

sprisha:
#### What type of PR is this?
Bug Fix/ Feature Improvement

 Modifies the way plan is generated by separating authDependency on the plan and transformAllocation, which helps resolve logical fallacy occurring with setting TA inside execution nodes, that later is used to determine if plan is realized in Memory.

 PR responds to a keystone raised by user where plan fails to execute because it is unexpectedly realizing in Memory for their TDS Join operation.

### DIFF
--- a/legend-engine-xt-relationalStore-executionPlan/src/main/java/org/finos/legend/engine/plan/execution/stores/relational/result/RealizedRelationalResult.java
+++ b/legend-engine-xt-relationalStore-executionPlan/src/main/java/org/finos/legend/engine/plan/execution/stores/relational/result/RealizedRelationalResult.java
@@ -52,14 +52,15 @@ public class RealizedRelationalResult extends StreamingResult
         this.transformedRows = Lists.mutable.empty();
         this.resultSetRows = Lists.mutable.empty();
         ResultSet resultSet = relationalResult.resultSet;
+        int SUPPORTED_RESULT_ROWS = getRowLimit();
         int rowCount = 0;
         try
         {
             while (resultSet.next())
             {
-                if (rowCount > getRowLimit())
+                if (rowCount > SUPPORTED_RESULT_ROWS )
                 {
-                    throw new RuntimeException("Too many rows returned. Realization of relational results currently supports results with up to " + getRowLimit() + " rows.");
+                    throw new RuntimeException("Too many rows returned. Realization of relational results currently supports results with up to " + SUPPORTED_RESULT_ROWS + " rows.");
                 }
 
                 List<Object> transformedRow = Lists.mutable.empty();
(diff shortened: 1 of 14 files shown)


legend-engine-xt-relationalStore-executionPlan/src/main/java/org/finos/legend/engine/plan/execution/stores/relational/result/RealizedRelationalResult.java
```diff
@@ -82,6 +83,15 @@ public RealizedRelationalResult(RelationalResult relationalResult) throws SQLExc
         }
     }
 
+    public int getRowLimit()
+    {
+        if (System.getProperty(ROW_LIMIT_PROPERTY_NAME) == null)
+        {
+            return DEFAULT_ROW_LIMIT;
+        }
+        return Integer.valueOf(System.getProperty(ROW_LIMIT_PROPERTY_NAME));
```
rafaelbey, line 92: Lets simplify this with a call to [getInteger](https://docs.oracle.com/javase/8/docs/api/java/lang/Integer.html#getInteger-java.lang.String-java.lang.Integer-)

```java
Integer.getInteger(ROW_LIMIT_PROPERTY_NAME, DEFAULT_ROW_LIMIT)
```

This will return the default if property is not present, otherwise will get it and parse it.

legend-engine-xt-relationalStore-executionPlan/src/main/java/org/finos/legend/engine/plan/execution/stores/relational/result/RealizedRelationalResult.java
```diff
@@ -40,6 +40,7 @@ public class RealizedRelationalResult extends StreamingResult
     public List<List<Object>> transformedRows;
 
     private static final int DEFAULT_ROW_LIMIT = 1000;
+    private static final String ROW_LIMIT_PROPERTY_NAME = "DEFAULT_ROW_LIMIT";
```
rafaelbey, line 43: lets give this a name that reflect the usage and avoid colliding with other property names.  Something like org.finos.legend.engine.realizedRelationalResultRowLimit.

legend-engine-xt-relationalStore-executionPlan/src/main/java/org/finos/legend/engine/plan/execution/stores/relational/result/RealizedRelationalResult.java
```diff
@@ -56,9 +57,9 @@ public RealizedRelationalResult(RelationalResult relationalResult) throws SQLExc
         {
             while (resultSet.next())
             {
-                if (rowCount > 1000)
+                if (rowCount > getRowLimit())
```
rafaelbey, line 60: calling this on each iteration will be expensive given the look into the properties map + parsing from string.

Please cache this into a field/variable.

TheJuanAndOnly99:
@cla-bot[bot] check

epsstan (approved):

PR #1004, opened 2022-10-03
properly process connections for LegacyRuntime

Yasirmod17:
#### What type of PR is this?

Bug Fix

On processing of legacy runtimes, the second pass visitor needs to be executed to accommodate extra processing of new connection parameters

Fixes #

No


davidharte-gs (approved):
As discussed, let's add a test in Alloy to cover this.

PR #1008, opened 2022-10-04
Handle snapshot date in milestoning property mapping creation

gs-ssh16:
#### What type of PR is this?

Bug

This PR fixes the issue with snapshot milestoning not being handled when creating embedded property mapping for 'milestoning' property.

No


gs-ssh16:
Depends on https://github.com/finos/legend-pure/pull/614

PR #1020, opened 2022-10-06
allow specifying session cookie name

akphi:
Allow specifying session cookie name to help with local development (when we have multiple Legend applications)

see https://stackoverflow.com/questions/16789495/two-applications-on-the-same-server-use-the-same-jsessionid


epsstan (approved):

legend-engine-xt-relationalStore-sqlserver-execution-tests/src/test/resources/org/finos/legend/engine/server/test/userTestConfig_withSqlServerTestConnection.json
```diff
@@ -16,6 +16,7 @@
       }
     ]
   },
+  "sessionCookie": "LEGEND_ENGINE_JSESSIONID",
```
vikask-gs, line 19: Please remove all changes from userTestConfig_with*TestConnection.json files. They are not used with Server, but with RelationalTestServer, which is minimal server used for testing.

PR #1021, opened 2022-10-06
Milestoning: Bitemp VLF Exclude Source Validity From Time Column in Sink

kumuwu:
#### What type of PR is this?

Feature fix

This PR removes a redundant output column when doing bitemporal writes into a relational sink.

Currently, the milestoning library supports bitemporal writes into relational sinks when the user has a single business date in their source data:

- The business date is copied into FROM_Z in the sink
- Chaining is performed; the library populates THRU_Z based on subsequent updates
- As-of updates are supported
- IN_Z/OUT_Z chaining is managed as system time of writes into sink

Before this PR we also copied the business date field into a separate data column.

This PR removes that separate data column since it's already available as FROM_Z

Fixes #

No


davidsunglee:
LGTM. Will speak w/team to get this merged.

davidsunglee:
More background:

This PR removes a redundant output column when doing bitemporal writes into a relational sink.

Currently, the milestoning library supports bitemporal writes into relational sinks when the user has a single business date in their source data:

- The business date is copied into FROM_Z in the sink
- Chaining is performed; the library populates THRU_Z based on subsequent updates
- As-of updates are supported
- IN_Z/OUT_Z chaining is managed as system time of writes into sink

Before this PR we also copied the business date field into a separate data column.

This PR removes that separate data column since it's already available as FROM_Z

epsstan (approved):

PR #1035, opened 2022-10-10
enhance relational compiler extension - Allow connection to flow to postprocessor

gs-kotang:
#### What type of PR is this?

Feature Enhancement

- This change gives us access to connection information in the postprocessor during compile time
- This is useful when we want to do postprocessing based on the type of database

Fixes #

No

### DIFF
--- a/legend-engine-xt-relationalStore-grammar/src/main/java/org/finos/legend/engine/language/pure/compiler/toPureGraph/IRelationalCompilerExtension.java
+++ b/legend-engine-xt-relationalStore-grammar/src/main/java/org/finos/legend/engine/language/pure/compiler/toPureGraph/IRelationalCompilerExtension.java
@@ -25,6 +25,7 @@ import org.finos.legend.engine.language.pure.compiler.toPureGraph.extension.Comp
 import org.finos.legend.engine.language.pure.compiler.toPureGraph.extension.Processor;
 import org.finos.legend.engine.protocol.pure.v1.model.SourceInformation;
 import org.finos.legend.engine.protocol.pure.v1.model.context.EngineErrorType;
+import org.finos.legend.engine.protocol.pure.v1.model.packageableElement.connection.Connection;
 import org.finos.legend.engine.protocol.pure.v1.model.packageableElement.store.relational.connection.authentication.AuthenticationStrategy;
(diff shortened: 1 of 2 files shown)


legend-engine-xt-relationalStore-grammar/src/main/java/org/finos/legend/engine/language/pure/compiler/toPureGraph/IRelationalCompilerExtension.java
```diff
@@ -93,7 +99,7 @@ default Iterable<? extends Processor<?>> getExtraProcessors()
         return Lists.immutable.with();
     }
 
-    default List<Function2<PostProcessor, CompileContext, Pair<Root_meta_pure_alloy_connections_PostProcessor, PostProcessorWithParameter>>> getExtraConnectionPostProcessor()
+    default List<Function3<Connection, PostProcessor, CompileContext, Pair<Root_meta_pure_alloy_connections_PostProcessor, PostProcessorWithParameter>>> getExtraConnectionPostProcessor()
```
rafaelbey, line 102: Can we keep the Function2 method to make the change backward compatible? (and make this deprecated)

We will need to slightly name this different, maybe **_getExtraConnectionPostProcessors()_**  (plural with _**s**_ at the end)

We can then decorate the Function2 implementations so the fit into the Function3 implementation, enhancing the code to support function while still respect previous implementations:

```java

    default List<Function3<Connection, PostProcessor, CompileContext, Pair<Root_meta_pure_alloy_connections_PostProcessor, PostProcessorWithParameter>>> getExtraConnectionPostProcessors()
    {      
        return this.getExtraConnectionPostProcessor().stream()
                .map(x -> (Function3<Connection, PostProcessor, CompileContext, Pair<Root_meta_pure_alloy_connections_PostProcessor, PostProcessorWithParameter>>) (conn, postProcessor, compileContext) -> x.apply(postProcessor, compileContext))
                .collect(Collectors.toList());
    }

```

As legend engine grows, and we start to get multiple projects importing these classes, we need to be careful on making breaking changes, and if possible, leverage strategies that enable us to move forward while been backward compatible.  

Let me know what you think.

gs-ssh16 (approved):

PR #1037, opened 2022-10-10
Remove closing connections async since take CPU resources from other parallel executions

rafaelbey:
#### What type of PR is this?

Enhancement 

This add support for sharing a thread executor for side tasks when executing plans.  This will allow us to move out of the fork join common pool, and into a pool that we can custom.

The current use-case is for IO tasks related to relational temp tables clean up.

The shared pool created for this will also be monitored using DW metrics to have visibility on execution stats.

This could impact performance for connection that take longer to close.

### DIFF
--- a/legend-engine-xt-relationalStore-executionPlan/src/main/java/org/finos/legend/engine/plan/execution/stores/relational/blockConnection/BlockConnectionContext.java
+++ b/legend-engine-xt-relationalStore-executionPlan/src/main/java/org/finos/legend/engine/plan/execution/stores/relational/blockConnection/BlockConnectionContext.java
@@ -14,6 +14,7 @@
 
 package org.finos.legend.engine.plan.execution.stores.relational.blockConnection;
 
+import java.util.concurrent.Executor;
 import org.eclipse.collections.api.list.MutableList;
 import org.eclipse.collections.api.map.MutableMap;
 import org.eclipse.collections.impl.factory.Maps;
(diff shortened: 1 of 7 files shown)


legend-engine-xt-relationalStore-executionPlan/src/main/java/org/finos/legend/engine/plan/execution/stores/relational/blockConnection/BlockConnectionContext.java
```diff
@@ -68,9 +66,10 @@ public void closeAllBlockConnections()
         this.blockConnectionMap.values().forEach(BlockConnection::close);
     }
 
+    @Deprecated
     public void closeAllBlockConnectionsAsync()
     {
-        this.blockConnectionMap.values().forEach(blockConnection -> CompletableFuture.runAsync(blockConnection::close));
```
gs-ssh16, line 73: IIRC, this was required to ensure executions are not blocked by stuff like dropping temp tables etc. which can be costly sometimes (they execute as part of connection close). Can we try using a different thread pool?

rafaelbey: That cost is what concerns me, and why I will like to understand the rational of doing this a separate thread.

We are giving the impression to the user that their request finished, and that can create the impression that the physical stores is ready for more load.  That is not true and, and the connection has not been return to the hikari pool either.  This means a user might miscalculate how many parallel request they could make, and Alloy might need to open more connections than what a user expect (or that the store can take).

If we are ok with this hidden behavior/cost on each request, Im happy to send to a different thread pool, but want to make sure we explicitly think of the impact - gaining performance on some scenarios could lead to poor performance on others.

rafaelbey: Per discussion  - closing the connection outside the request thread ensures users get responses within agreed SLA.  

This change then create the notion of an external thread pool for these tasks., and as discussed, ensure we capture stats to better visualize if it becomes a bottle neck.

finos-admin:

PR #1038, opened 2022-10-11
Deserialize float value as BigDecimal in JSON assertion

gs-ssh16:
#### What type of PR is this?

Bug

Current JSON assertion code uses float type for decimal values which can lead to incorrect results due to rounding off and loss in precision.

### DIFF
--- a/legend-engine-testable/src/main/java/org/finos/legend/engine/testable/assertion/TestAssertionHelper.java
+++ b/legend-engine-testable/src/main/java/org/finos/legend/engine/testable/assertion/TestAssertionHelper.java
@@ -28,9 +28,14 @@ import java.io.IOException;
 
 public class TestAssertionHelper
 {
+    public static ObjectMapper buildObjectMapperForJSONComparison()
+    {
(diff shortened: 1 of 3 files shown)


legend-engine-testable/src/main/java/org/finos/legend/engine/testable/assertion/TestAssertionHelper.java
```diff
@@ -29,7 +30,7 @@
 {
     protected static AssertionStatus compareAssertionJSON(TestAssertion parentAssertion, String _expected, String _actual) throws IOException
     {
-        ObjectMapper objectMapper = new ObjectMapper();
+        ObjectMapper objectMapper = new ObjectMapper().enable(DeserializationFeature.USE_BIG_DECIMAL_FOR_FLOATS);
```
hardikmaheshwari, line 33: Should this be added to PureProtocolObjectMapperFactory and we use objectMapper from that factory? What do you think?

gs-ssh16: Don't think that's required in JSON comparison

hardikmaheshwari (approved):

PR #1055, opened 2022-10-17
Fix NPE bug in ServiceTestRunner

epsstan:
#### What type of PR is this?

Bug fix

Fixes a NPE in the ServiceTestRunner.

Fixes #

No

### DIFF
--- a/legend-engine-test-runner-service/src/main/java/org/finos/legend/engine/test/runner/service/ServiceTestRunner.java
+++ b/legend-engine-test-runner-service/src/main/java/org/finos/legend/engine/test/runner/service/ServiceTestRunner.java
@@ -173,9 +173,18 @@ public class ServiceTestRunner
             try (Scope scope = GlobalTracer.get().buildSpan("Generate Single Pure Tests And Run").startActive(true))
             {
                 SingleExecutionTest test = (SingleExecutionTest) service.test;
-                List<TestContainer> asserts = (test == null) ? Collections.emptyList() : test.asserts;
-                String noAssertMessage = "No test assert found !!";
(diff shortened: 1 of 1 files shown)


legend-engine-test-runner-service/src/main/java/org/finos/legend/engine/test/runner/service/ServiceTestRunner.java
```diff
@@ -172,9 +172,10 @@ else if (serviceExecution instanceof PureSingleExecution)
         {
             try (Scope scope = GlobalTracer.get().buildSpan("Generate Single Pure Tests And Run").startActive(true))
             {
-                List<TestContainer> asserts = ((SingleExecutionTest) service.test).asserts;
+                SingleExecutionTest test = (SingleExecutionTest) service.test;
+                List<TestContainer> asserts = (test == null) ? Collections.emptyList() : test.asserts;
                 String noAssertMessage = "No test assert found !!";
-                return Collections.singletonList(executeSingleExecutionTest((PureSingleExecution) service.execution, ((SingleExecutionTest) service.test).data, asserts, noAssertMessage, pureModelContextData, pureModel, scope));
+                return Collections.singletonList(executeSingleExecutionTest((PureSingleExecution) service.execution, test.data, asserts, noAssertMessage, pureModelContextData, pureModel, scope));
```
rafaelbey, line 178: test.data could lead to NPE based on the added check on line 176.  

Maybe will be easier to follow if we wrap line 175 on Optional.ofNullable() so that everywhere else we are force to make a choice on how to handle it.

rafaelbey (approved):

PR #1058, opened 2022-10-17
Add support for SQL query protocol

gs-manvig:
#### What type of PR is this?

Feature

Add a new query protocol - SQL

N/A


legend-engine-xt-sql-query/src/main/java/org/finos/legend/engine/query/sql/api/execute/MyCookie.java
```diff
@@ -0,0 +1,106 @@
+// Copyright 2022 Goldman Sachs
```
aormerod-gs, line 1: What the context for this class? (It feels like a test class at best?)

gs-manvig: The name is not great, but this cookie implementation is used to copy cookies from current call and then use those to make Legend SDLC API calls. Using this light weight implementation to avoid pulling in big libraries like apache client. We should probably move similar usage in GraphQL to a common library and re use it - [https://github.com/finos/legend-engine/blob/master/legend-engine-xt-graphQL-query/src/main/java/org/finos/legend/engine/query/graphQL/api/MyCookie.java](https://github.com/finos/legend-engine/blob/master/legend-engine-xt-graphQL-query/src/main/java/org/finos/legend/engine/query/graphQL/api/MyCookie.java)

aormerod-gs: Yes, moving to a shared implementation makes sense.  Should we include in this change set?

gs-manvig: We could move this to a shared implementation. However, the current strategy of pulling service definition from Legend SDLC is temporary until a shared implementation/cache for retrieving service definition is available

aormerod-gs (approved):

PR #1087, opened 2022-10-29
fix service test assertion failure related to float decimal

YannanGao-gs:
#### What type of PR is this?

Bug fix

https://user-images.githubusercontent.com/73408381/198812544-da5865d8-525c-4028-8353-fd78b9082035.mov

Fixes #


YannanGao-gs:
> lets add an engine bug issue for this and lets also add a test for this case. Thanks

one test is added

MauricioUyaguari:
> > lets add an engine bug issue for this and lets also add a test for this case. Thanks
> 
> one test is added

please add a test with execution. You can execute and check the assert failure and expected results. Thanks

legend-engine-testable/src/main/java/org/finos/legend/engine/testable/assertion/TestAssertionHelper.java
```diff
@@ -30,7 +32,10 @@
 {
     public static ObjectMapper buildObjectMapperForJSONComparison()
     {
-        return new ObjectMapper().enable(DeserializationFeature.USE_BIG_DECIMAL_FOR_FLOATS);
+        return new ObjectMapper()
+                .enable(DeserializationFeature.USE_BIG_DECIMAL_FOR_FLOATS)
+                .enable(JsonGenerator.Feature.WRITE_BIGDECIMAL_AS_PLAIN)
```
hardikmaheshwari, line 37: Do we really need this? Isn't setting node factory with the below statement enough?
If we need this can we have a test that fails without this statement, I believe the current test would pass even without this
If no, can we remove this and also the dependency that was added?

legend-engine-testable/src/test/java/org/finos/legend/engine/testable/assertion/TestTestAssertionEvaluator.java
```diff
@@ -152,4 +154,22 @@ public void testEqualToJsonAssertionWithDecimalPrecision()
         Assert.assertTrue(assertionStatus instanceof AssertPass);
         Assert.assertEquals("assert1", assertionStatus.id);
     }
+
+    @Test
+    public void testEqualToAssertionWithFloatPrecision() throws IOException
```
hardikmaheshwari, line 159: Can we please modify the test to not get object mapper and have replicated code already present in TestAssertionHelper.
Please take a look at this test `testEqualToJsonAssertionWithDecimalPrecision` above, we should use the visitor directly and have logic there itself and just assert on output.

legend-engine-testable/src/test/java/org/finos/legend/engine/testable/assertion/TestTestAssertionEvaluator.java
```diff
@@ -152,4 +154,22 @@ public void testEqualToJsonAssertionWithDecimalPrecision()
...
+    {
+        ConstantResult constantResult = new ConstantResult("{\"some\": 2.0}");
+        ExternalFormatData data = new ExternalFormatData();
+        data.contentType = "application/json";
+        EqualToJson equalToJson = new EqualToJson();
+        equalToJson.id = "assert1";
+        equalToJson.expected = data;
+        data.data = "{\"some\":2.0}";
```
hardikmaheshwari, line 167: Let's please add more scenarios with passes and failures, like (but not limited to)
- 2000000000000000000
- 2.000000000000000000
- 2.000000000000000001
- 2.0
- 20

hardikmaheshwari (changes requested):

legend-engine-testable/src/main/java/org/finos/legend/engine/testable/assertion/TestAssertionHelper.java
YannanGao-gs: These are needed. If we only keep `.setNodeFactory(JsonNodeFactory.withExactBigDecimals(true)`, `testEqualToJsonAssertionWithDecimalPrecision` will start failing. 
Please check my newly added tests.

hardikmaheshwari (approved):

hardikmaheshwari:
@YannanGao-gs Thanks for fixing this

MauricioUyaguari (approved):

Yasirmod17 (approved):

PR #1101, opened 2022-11-04
add support for property aggregation

akphi:
#### What type of PR is this?

New Feature

Added support for property aggregation: `AggregationKind.None`, `AggregationKind.Shared`, `AggregationKind.Composite` like what we have in `legend-pure`

### DIFF
--- a/legend-engine-language-pure-compiler/src/main/java/org/finos/legend/engine/language/pure/compiler/toPureGraph/HelperModelBuilder.java
+++ b/legend-engine-language-pure-compiler/src/main/java/org/finos/legend/engine/language/pure/compiler/toPureGraph/HelperModelBuilder.java
@@ -20,6 +20,7 @@ import org.eclipse.collections.impl.factory.Lists;
 import org.eclipse.collections.impl.utility.LazyIterate;
 import org.eclipse.collections.impl.utility.ListIterate;
 import org.finos.legend.engine.protocol.pure.v1.model.context.EngineErrorType;
+import org.finos.legend.engine.protocol.pure.v1.model.packageableElement.domain.AggregationKind;
 import org.finos.legend.engine.protocol.pure.v1.model.packageableElement.domain.Class;
 import org.finos.legend.engine.protocol.pure.v1.model.packageableElement.domain.Function;
 import org.finos.legend.engine.protocol.pure.v1.model.packageableElement.domain.Multiplicity;
@@ -52,6 +53,7 @@ import org.finos.legend.pure.m3.coreinstance.meta.pure.metamodel.type.generics.G
 import org.finos.legend.pure.m3.coreinstance.meta.pure.metamodel.valuespecification.ValueSpecification;
 import org.finos.legend.pure.m3.coreinstance.meta.pure.metamodel.valuespecification.ValueSpecificationContext;
 import org.finos.legend.pure.m3.coreinstance.meta.pure.metamodel.valuespecification.VariableExpression;
+import org.finos.legend.pure.m3.navigation.M3Paths;
 import org.finos.legend.pure.m4.coreinstance.CoreInstance;
 import org.finos.legend.pure.m4.coreinstance.SourceInformation;
 import org.finos.legend.pure.runtime.java.compiled.execution.CompiledExecutionSupport;
@@ -77,6 +79,14 @@ public class HelperModelBuilder
                     ._multiplicity(context.pureModel.getMultiplicity(property.multiplicity))
                     ._stereotypes(ListIterate.collect(property.stereotypes, s -> context.resolveStereotype(s.profile, s.value, s.profileSourceInformation, s.sourceInformation)))
                     ._taggedValues(ListIterate.collect(property.taggedValues, t -> new Root_meta_pure_metamodel_extension_TaggedValue_Impl("", null, context.pureModel.getClass("meta::pure::metamodel::extension::TaggedValue"))._tag(context.resolveTag(t.tag.profile, t.tag.value, t.tag.profileSourceInformation, t.sourceInformation))._value(t.value)))
+                    ._aggregation(AggregationKind.COMPOSITE.equals(property.aggregation)
+                            ? context.pureModel.getEnumValue(M3Paths.AggregationKind, "Composite")
+                            : AggregationKind.SHARED.equals(property.aggregation)
+                            ? context.pureModel.getEnumValue(M3Paths.AggregationKind, "Shared")
+                            : AggregationKind.NONE.equals(property.aggregation)
+                            ? context.pureModel.getEnumValue(M3Paths.AggregationKind, "None")
+                            : null
+                    )
                     ._owner(owner);
         };
     }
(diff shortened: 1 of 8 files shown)


legend-engine-language-pure-grammar/src/test/java/org/finos/legend/engine/language/pure/grammar/test/roundtrip/TestDomainGrammarRoundtrip.java
```diff
@@ -19,6 +19,28 @@
 
 public class TestDomainGrammarRoundtrip extends TestGrammarRoundtrip.TestGrammarRoundtripTestSuite
 {
+    @Test
```
epsstan, line 22: Not sure about this grammar.  Any reason why we this should not be modeled as a profile/tag value ? 

@pierredebelen might have a strong view on this one.

Generally, my preference is to keep the variations in syntax/grammar to a minimum.

akphi: I bring over what we support in Pure, don't think we should diverge here. Also this is a native UML capability

legend-engine-language-pure-compiler/src/main/java/org/finos/legend/engine/language/pure/compiler/toPureGraph/HelperModelBuilder.java
```diff
@@ -77,6 +79,14 @@ public static org.eclipse.collections.api.block.function.Function<Property, org.
...
                     ._stereotypes(ListIterate.collect(property.stereotypes, s -> context.resolveStereotype(s.profile, s.value, s.profileSourceInformation, s.sourceInformation)))
                     ._taggedValues(ListIterate.collect(property.taggedValues, t -> new Root_meta_pure_metamodel_extension_TaggedValue_Impl("", null, context.pureModel.getClass("meta::pure::metamodel::extension::TaggedValue"))._tag(context.resolveTag(t.tag.profile, t.tag.value, t.tag.profileSourceInformation, t.sourceInformation))._value(t.value)))
+                    ._aggregation(AggregationKind.COMPOSITE.equals(property.aggregation)
+                            ? context.pureModel.getEnumValue(M3Paths.AggregationKind, "Composite")
+                            : AggregationKind.SHARED.equals(property.aggregation)
+                            ? context.pureModel.getEnumValue(M3Paths.AggregationKind, "Shared")
+                            : AggregationKind.NONE.equals(property.aggregation)
+                            ? context.pureModel.getEnumValue(M3Paths.AggregationKind, "None")
+                            : null
```
kevin-m-knight-gs, line 88: If property.aggregation is a non-null and not one of the previous values, then we need to throw an exception.

kevin-m-knight-gs (changes requested):
I think we need to be more rigorous about throwing exceptions when we run into an unexpected value. Also, we need some tests for invalid cases.

legend-engine-language-pure-grammar/src/main/java/org/finos/legend/engine/language/pure/grammar/to/HelperDomainGrammarComposer.java
```diff
@@ -74,7 +75,18 @@ public static String renderMultiplicity(Multiplicity multiplicity)
...
+
+    private static String renderAggregation(AggregationKind aggregationKind)
+    {
+        return AggregationKind.NONE.equals(aggregationKind)
+                ? "(none) "
+                : AggregationKind.SHARED.equals(aggregationKind)
+                    ? "(shared) "
+                    : AggregationKind.COMPOSITE.equals(aggregationKind)
+                        ? "(composite) "
+                        : "";
```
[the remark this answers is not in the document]
akphi: I think we avoid throwing error in grammar composer

legend-engine-language-pure-grammar/src/main/java/org/finos/legend/engine/language/pure/grammar/from/domain/DomainParseTreeWalker.java
```diff
@@ -332,6 +333,15 @@ private Property visitSimpleProperty(DomainParserGrammar.PropertyContext ctx)
...
         property.sourceInformation = this.walkerSourceInformation.getSourceInformation(ctx);
         property.propertyTypeSourceInformation = this.walkerSourceInformation.getSourceInformation(ctx.propertyReturnType().type());
+        property.aggregation = ctx.aggregation() != null
+                ? ctx.aggregation().aggregationType().AGGREGATION_TYPE_COMPOSITE() != null
+                    ?  AggregationKind.COMPOSITE
+                    : ctx.aggregation().aggregationType().AGGREGATION_TYPE_SHARED() != null
+                        ? AggregationKind.SHARED
+                        : ctx.aggregation().aggregationType().AGGREGATION_TYPE_NONE() != null
+                            ? AggregationKind.NONE
+                            : null
```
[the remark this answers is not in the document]
akphi: I don't think we can do it here, it's part of ANTLR logic that has ensured that won't happen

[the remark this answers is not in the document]
akphi: Though I can definitely make this logic look clearer by not using ternary op here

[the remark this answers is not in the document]
akphi: Updated per suggestions, seem like I'm delusional to think we consistently don't throw error in composer 😢

legend-engine-language-pure-grammar/src/main/java/org/finos/legend/engine/language/pure/grammar/from/domain/DomainParseTreeWalker.java
```diff
@@ -332,6 +333,19 @@ private Property visitSimpleProperty(DomainParserGrammar.PropertyContext ctx)
...
+            if (ctx.aggregation().aggregationType().AGGREGATION_TYPE_COMPOSITE() != null)
+            {
+                property.aggregation = AggregationKind.COMPOSITE;
+            } else if (ctx.aggregation().aggregationType().AGGREGATION_TYPE_SHARED() != null)
+            {
+                property.aggregation = AggregationKind.SHARED;
+            } else if (ctx.aggregation().aggregationType().AGGREGATION_TYPE_NONE() != null)
+            {
+                property.aggregation = AggregationKind.NONE;
+            }
```
kevin-m-knight-gs: I still think you need to throw an exception here if you get a non-null aggregation kind you do not understand. For example, what if someone adds a new aggregation kind in the parser grammar without updating this code? Then any usages of the new aggregation kind will be parsed incorrectly. It's much better to throw an exception to make sure you're not letting any cases fall through the cracks.
```suggestion
            if (ctx.aggregation().aggregationType().AGGREGATION_TYPE_COMPOSITE() != null)
            {
                property.aggregation = AggregationKind.COMPOSITE;
            }
            else if (ctx.aggregation().aggregationType().AGGREGATION_TYPE_SHARED() != null)
            {
                property.aggregation = AggregationKind.SHARED;
            }
            else if (ctx.aggregation().aggregationType().AGGREGATION_TYPE_NONE() != null)
            {
                property.aggregation = AggregationKind.NONE;
            }
            else
            {
                throw new EngineException("Unknown aggregation kind", EngineErrorType.PARSER);
            }
```

kevin-m-knight-gs (changes requested):

kevin-m-knight-gs (approved):

legend-engine-language-pure-grammar/src/test/java/org/finos/legend/engine/language/pure/grammar/test/roundtrip/TestDomainGrammarRoundtrip.java
pierredebelen: Yep it's part of the UML metamodel

PR #1113, opened 2022-11-07
Using larger runners for the release action

ivan-kyosev-gs:
#### What type of PR is this?

This is an enhancement for the release action. The action will now use larger GitHub runners.

The current release action is quite slow. Using a larger runner will speed up the build.

No

### DIFF
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -20,7 +20,7 @@ env:
   CI_GPG_PASSPHRASE: ${{ secrets.CI_GPG_PASSPHRASE }}
   DOCKER_USERNAME: finos
   DOCKER_PASSWORD: ${{ secrets.DOCKER_PASSWORD }}
-  MAVEN_OPTS: -Xmx4g
+  MAVEN_OPTS: -Xms10g -Xmx10g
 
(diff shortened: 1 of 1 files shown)


.github/workflows/release.yml
```diff
@@ -31,7 +31,7 @@ on:
 
 jobs:
   release:
-    runs-on: ubuntu-latest
+    runs-on: ubuntu-latest-16-cores
```
akphi, line 34: is this something Github supports natively or is this our custom runner configured somewhere?

ivan-kyosev-gs: This is a custom runner that FINOS have set up. Only org admins have access to its configuration.

epsstan (approved):

.github/workflows/release.yml
akphi: @ivan-kyosev-gs this didn't seem to work though, is it because the runners are not yet available?

ivan-kyosev-gs: @akphi I've added you to the discussion with FINOS. Looks like some of the settings for the runner are misconfigured at the moment.

PR #1118, opened 2022-11-08
make GraphQL API endpoints support fetching models from depot/metadata server

akphi:
#### What type of PR is this?

new feature

Currently, GraphQL API endpoints only allow specifying the SDLC project coordinate to fetch model data, we would like to allow specifying pure model context data as input for more flexibility

This will work when https://github.com/finos/legend-depot/pull/139 is merged


pierredebelen:
I don't think a graphQL client would be able to send the metadata. Getting it from the 'depot' in Prod would be good to add though.

akphi:
> I don't think a graphQL client would be able to send the metadata. Getting it from the 'depot' in Prod would be good to add though.

Sorry, that's what I implemented, I just named the PR wrong. Fixed now.

elopezcastro (approved):

PR #1135, opened 2022-11-11
Enable spanner

roasher:
#### What type of PR is this?

Enabling spanner connector functionality

After this change spanner would be legal db for studio projects

this PR contains shading google spanner jdbc and relocation of all it's compiled classes to custom package, so there would be no dependency conflicts with other part of platform

No, just adding new db type among others


legend-engine-xt-relationalStore-spanner-jdbc-shaded/pom.xml
```diff
@@ -0,0 +1,338 @@
...
+         xsi:schemaLocation="http://maven.apache.org/POM/4.0.0 http://maven.apache.org/xsd/maven-4.0.0.xsd">
+    <parent>
+        <artifactId>legend-engine</artifactId>
+        <groupId>org.finos.legend.engine</groupId>
+        <version>3.22.4-SNAPSHOT</version>
+    </parent>
+    <modelVersion>4.0.0</modelVersion>
+
+    <artifactId>legend-engine-xt-relationalStore-spanner-jdbc-shaded</artifactId>
+    <name>Legend Engine - XT - Relational Store - Spanner - Execution</name>
```
vikask-gs, line 13: incorrect name?

legend-engine-xt-relationalStore-spanner-jdbc-shaded/pom.xml
```diff
@@ -0,0 +1,338 @@
...
+        <maven-shade-plugin.shadedPattern>org.finos.legend.engine.spanner.jdbc.shaded</maven-shade-plugin.shadedPattern>
+    </properties>
+    
+    <dependencies>
+        <dependency>
+            <groupId>com.google.cloud</groupId>
+            <artifactId>google-cloud-spanner-jdbc</artifactId>
+            <version>2.7.12</version>
+            <classifier>single-jar-with-dependencies</classifier>
+            <!--excluding everything cause everything would be in full jar and shaded-->
```
vikask-gs, line 25: Is there a way to exclude everything with a single directive, instead of explicit listing? When we increment the version, and miss excluding a newly added dependency, then it might be problem

roasher: It seems that templates are also working:

            <exclusions>
                <exclusion>
                    <groupId>*</groupId>
                    <artifactId>*</artifactId>
                </exclusion>
            </exclusions>

vikask-gs:
Review looks good at the initial look. However, we would need to understand the impact of shading within the legend-engine project (build time/project organization wise). @rafaelbey @davidharte-gs, pls add your thoughts

roasher:
image

vikask-gs (approved):

legend-engine-xt-relationalStore-spanner-jdbc-shaded/pom.xml
```diff
@@ -0,0 +1,137 @@
...
+                </exclusion>
+            </exclusions>
+        </dependency>
+    </dependencies>
+
+    <build>
+        <plugins>
+            <plugin>
+                <artifactId>maven-shade-plugin</artifactId>
+                <configuration>
```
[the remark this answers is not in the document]
roasher: After shading and relocation shade plugin creates dependency-reduced.pom with no dependencies and shaded.jar
To use it I triggered install plugin with -local postfix.
<classifier> won't work cause classifiers doesn't remove dependencies.

At the end of the day there is no exclusions now

roasher:
Connection tests passed
image

legend-engine-xt-relationalStore-spanner-jdbc-shaded/pom.xml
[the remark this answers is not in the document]
roasher: @rafaelbey 
It seems not possible due to fact that shaded library is a submodule, not separate project.
`mvn install legend-engine-xt-relationalStore-spanner-jdbc-shaded` is creating in local repo correct jar and correct pom **without** any deps.

Nevertheless, using this submodule in other submodules ends with an error. I think that this is because maven reads pom inside legend-engine-xt-relationalStore-spanner-jdbc-shaded, not using it from local repo.

But, if legend-engine-xt-relationalStore-spanner-jdbc-shaded would be moved to separate project, exclusion can be omitted.
Take a look at picture of maven with submodule being part of the project:
 
image

and now if it is separate:
image

Note than in first case I can expand deps, and in second not. 
I did build with external project without exclusions and it went well.

Also, in Alloy now jdbc submodule is used without exclusion that proves that having shaded lib in separate project is crutual.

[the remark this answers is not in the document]
vikask-gs: It seems like maven enforcer plugin uses pom.xml instead of dependency-reduced-pom.xml, when the dependency is a module within the project. This requires us to put * exclusion to satisfy enforcer.

However, @roasher can you try adding https://maven.apache.org/plugins/maven-shade-plugin/shade-mojo.html#shadedArtifactAttached to the shade-plugin, and then add <classifier>shaded</classifier> to the dependency declarations. Maybe, enforcer plugin will handle shaded classifier as a special case, and not complain.

If it still complains, then using * exclusion with a comment explaining why, is better.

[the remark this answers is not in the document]
roasher: @vikask-gs 
Checked that - didn't work.
if shadedClassifierName option is used, one need to set shadedArtifactAttached to true as well: 
```
<shadedArtifactAttached>true</shadedArtifactAttached>
<shadedClassifierName>shaded</shadedClassifierName>
```
So, it that case we are telling maven to create additional version with "shaded" classifier, but the problem is that we are not replacing output jar and pom with shaded one: there 2 jars in outcome and single pom. And that pom as long as it's not replaced with dependency-reduced.pom would still have google-cloud-spanner-jdbc dependency that would force to exclude everything not only in legend-engine, but in all projects that are using is as a dependency.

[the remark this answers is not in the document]
vikask-gs: We are not replacing the pom. But I was hoping that if you add the dependency with `<classifier>shaded</classifier>`, then enforcer plugin might automatically exclude transitive deps from pom.

Can you confirm again that it is not the case? 

If yes, then using * exclusion with a comment is fine.

legend-engine-xt-relationalStore-spanner-execution/pom.xml
```diff
@@ -68,9 +68,15 @@
 
         <!-- DRIVER -->
         <dependency>
-            <groupId>com.google.cloud</groupId>
-            <artifactId>google-cloud-spanner-jdbc</artifactId>
+            <groupId>org.finos.legend.engine</groupId>
+            <artifactId>legend-engine-xt-relationalStore-spanner-jdbc-shaded</artifactId>
             <scope>runtime</scope>
+            <exclusions>
```
vikask-gs, line 74: do we need exclusion here, since it is already done in dependecy management of parent pom?

legend-engine-server/pom.xml
```diff
@@ -351,6 +351,33 @@
...
+            <groupId>org.finos.legend.engine</groupId>
+            <artifactId>legend-engine-xt-relationalStore-spanner-execution</artifactId>
+            <scope>runtime</scope>
+        </dependency>
+        <dependency>
+            <groupId>org.finos.legend.engine</groupId>
+            <artifactId>legend-engine-xt-relationalStore-spanner-jdbc-shaded</artifactId>
+            <scope>runtime</scope>
+            <exclusions>
+                <exclusion>
```
vikask-gs, line 375: pls add a comment here describing that we need exclusion to satisfy the enforcer plugin, when shaded dependency is a module within the project

legend-engine-xt-relationalStore-spanner-jdbc-shaded/pom.xml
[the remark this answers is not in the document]
roasher: Yes, I can.
Added comment for exclusion.

legend-engine-xt-relationalStore-spanner-execution/pom.xml
roasher: It is not done in parent pom, but I will check if I could do that

roasher: You were right, only exclusion in dependency management is needed. Refactored this

legend-engine-server/pom.xml
roasher: Added comment in dependency management, all other exclusions removed

pom.xml
```diff
@@ -771,6 +771,19 @@
                 <artifactId>legend-engine-xt-relationalStore-spanner-grammar</artifactId>
                 <version>${project.version}</version>
             </dependency>
+            <dependency>
+                <groupId>org.finos.legend.engine</groupId>
+                <artifactId>legend-engine-xt-relationalStore-spanner-jdbc-shaded</artifactId>
+                <version>${project.version}</version>
+                <!-- need to exclude everything as long is this submodule is not extracted to separate project
```
vikask-gs, line 778: Use this as comment: we need exclusion to satisfy the enforcer plugin, when shaded dependency is a module within the project, as enforcer (incorrectly) relies on module's source pom.xml instead of it's published pom.xml. As expected, the problem goes away when the same dependency comes from a different project.

vikask-gs:
Good to merge! But since alloy changes are involved, needs to be merged, just before release

gs-ssh16 (approved):
Approving and merging as per Vikas' comment

PR #1137, opened 2022-11-11
Ability to consume java Json representation of results

rafaelbey:
#### What type of PR is this?

Enhancement 

Enhance the service Result classes, providing the result output on a Java json object (jackson JsonNode), allowing for more simpler and efficient consumption of result when results are meant to be consume by a Java process.

Before, all results were required to be written to output stream, not a format that is easy to consume efficiently on a Java process.

Fixes #


rafaelbey:
Some changes on test cases to make them OS independent.

PR #1040, opened 2022-10-12
Support for services with only query lambda

gs-gunjan:
#### What does this PR do / why is it needed ?
This PR would allow users to create services with just a query written as part of the execution of services.

#### Which issue(s) this PR fixes:

NA

#### Other notes for reviewers:
This is for specifying single execution services with query lambda

#### Does this PR introduce a user-facing change?

No

PR #1152, opened 2022-11-15
Support for services with only query lambda

gs-gunjan:

Support for services with only query lambda

Fixes #


gs-ssh16:
Same as https://github.com/finos/legend-engine/pull/1040

gs-ssh16 (approved):

PR #1153, opened 2022-11-15
Draft: Persistence Testing Framework: Assert on Open Records Only

kumuwu:
#### What type of PR is this?

Feature fix

Originally, users assert on ALL rows of the sink - including rows which are already milestoned/invalidated. This PR introduces a change to the assertion mechanism such that users only assert on open records (rows which are still open and not invalidated) - as this makes more sense to the user as to their view, milestoning is more like a black box and they care more about only the open records.

Fixes #

Yes


finos-admin:

finos-admin:

PR #1180, opened 2022-11-21
Persistence: Bitemporal Milestoning Prefixing Table/Column Aliases

kumuwu:
#### What type of PR is this?

Bug fix

Bitemporal Delta Milestoning uses some constant aliases for columns/tables in the SQLs generated. The aliases used have been too generic, such as "start_date" which may conflict with actual user columns. This PR prefixes such internally used aliases with "legend_presistence_" (for example, the column alias "start_date" will now become "legend_persistence_start_date") so as to prevent such collisions.

Fixes #

No


prasar-ashutosh (approved):
Changes looks good!

rafaelbey (approved):
This is a good step to make this more unique hence approved.  Still, wonder if we should consider a runtime "unique" key to be use at some point to guarantee no conflict in the future.

PR #1193, opened 2022-11-28
[Draft] Persistence Component: Implement AssertAllRows and AssertActiveRows

kumuwu:

Fixes #


legend-engine-pure-code-compiled-core/src/main/resources/core/pure/test/testable.pure
```diff
@@ -95,3 +107,13 @@ Class meta::pure::test::assertion::EqualToJson extends TestAssertion
 {
     expected    : EmbeddedData[1];
 }
+
+Class meta::pure::test::assertion::AssertAllRows extends TestAssertion
```
hardikmaheshwari, line 111: This seems specific to Persistence in this state. Let's not add these in core and have in persisitence rather

PR #1213, opened 2022-12-02
Provide a configurable way to set up Infinite batch_id value and ANSI tests for bitemporal delta statistics

rengam32:
#### What type of PR is this?

Feature

This PR introduces a configurable way to set up Infinite batch_id value from the user. 
Also includes the tests for bitemporal delta statistics (with all the scenarios) in the ANSI module. 

None

No


prasar-ashutosh (approved):
Looks good to me

PR #1245, opened 2022-12-12
Support for non-relational database (document store) within Legend engine

ghub-real:
What type of PR is this?
Enhancement

What does this PR do / why is it needed ?
Support for non-relational stores (starting with document store as the first one)

In this PR, the goal is to propose a store model for document store (eg., MongoDB) and a basic query model - that can be leveraged to generated a pipeline string for execution against the store.

Other notes for reviewers:
Basic model for review before adding additional query capabilities.

Does this PR introduce a user-facing change?
No user-facing change

### DIFF
--- a/.github/workflows/database-athena-integration-test.yml
+++ b/.github/workflows/database-athena-integration-test.yml
@@ -64,6 +64,9 @@ jobs:
         env:
           MAVEN_OPTS: -Xmx4g
         run: mvn clean install -DskipTests=true
+      - name: Download Athena Driver Jar
+        run: |
+          wget --output-document=athenaDriver.jar http://s3.amazonaws.com/athena-downloads/drivers/JDBC/SimbaAthenaJDBC-2.0.34.1000/AthenaJDBC42-2.0.34.1000.jar
       - name: Run Connection Protocol Integration Tests
         env:
           MAVEN_OPTS: -Xmx4g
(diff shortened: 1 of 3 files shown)


rafaelbey:
I do think some generalization will indeed help, but don't think we should expose to users a common grammar to specify document stores.  

I've been working on ElasticSearh, and Mongo and ES have some differences on how they treat their documents and the typing system.  Hence, a common model will either limit features that are unique to some stores,  or create usability issues by trying to support the superset of features.  

So, we should try to keep the commonalities as an implementation detail, but expose to users grammar that reflect the store they want to model, ensuring we can deviate if required to properly represent a given store feature.

ghub-real:
Would be good to synch up on ES store model.  Looking at the non-relational space in general our view was that document store(mongo)  & key-value stores(dynamo even though it does support document data types) & graph databases could end up with different store models.  So the idea was to have a store model that works for mongo/documentdb/firestore etc.  The datatypes can be abstracted away.    Happy to chat about fragments which might be work in the context of re-usable document models.

rafaelbey:
Will try to contact this week to discuss further.

For ES reference, here is a branch with my changes (this is very early): https://github.com/rafaelbey/legend-engine/tree/elastic_search/legend-engine-xt-elasticsearch-pure/src/main/resources/core_elasticsearch

You will see some similarities on the metamodel, but there are some places where it diverges.

PR #1253, opened 2022-12-15
Add new apis for transformGrammarToJson and transformJsonToGrammar

gayathrir11:

Related to finos/legend-studio#1657

Fixes #

### DIFF
--- a/legend-engine-language-pure-grammar/src/main/java/org/finos/legend/engine/language/pure/grammar/from/PureGrammarParser.java
+++ b/legend-engine-language-pure-grammar/src/main/java/org/finos/legend/engine/language/pure/grammar/from/PureGrammarParser.java
@@ -26,6 +26,7 @@ import org.finos.legend.engine.language.pure.grammar.from.extension.PureGrammarP
 import org.finos.legend.engine.language.pure.grammar.from.extension.SectionParser;
 import org.finos.legend.engine.language.pure.grammar.from.mapping.MappingParser;
 import org.finos.legend.engine.language.pure.grammar.from.runtime.RuntimeParser;
+import org.finos.legend.engine.protocol.pure.v1.model.ElementBasedSourceInformation;
 import org.finos.legend.engine.protocol.pure.v1.model.SourceInformation;
 import org.finos.legend.engine.protocol.pure.v1.model.context.EngineErrorType;
 import org.finos.legend.engine.protocol.pure.v1.model.context.PureModelContextData;
@@ -42,6 +43,7 @@ import org.finos.legend.engine.shared.core.operational.logs.LogInfo;
 import org.finos.legend.engine.shared.core.operational.logs.LoggingEventType;
 import org.slf4j.Logger;
 
+import java.util.Map;
 import java.util.function.Consumer;
 
 public class PureGrammarParser
@@ -75,6 +77,11 @@ public class PureGrammarParser
     }
 
     public PureModelContextData parseModel(String code, String sourceId, int lineOffset, int columnOffset, boolean returnSourceInfo)
+    {
+        return this.parse(code, this.parsers, sourceId, lineOffset, columnOffset, returnSourceInfo);
+    }
+
+    public PureModelContextData parseElements(Map<String, String> code, String sourceId, int lineOffset, int columnOffset, boolean returnSourceInfo)
     {
         return this.parse(code, this.parsers, sourceId, lineOffset, columnOffset, returnSourceInfo);
     }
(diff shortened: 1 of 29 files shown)


legend-engine-language-pure-grammar/src/main/java/org/finos/legend/engine/language/pure/grammar/to/PureGrammarComposer.java
```diff
@@ -105,6 +107,55 @@ public String renderPureModelContextData(PureModelContextData pureModelContextDa
...
+    public Map<String, String> renderPureModelContextDataForEachElement(PureModelContextData pureModelContextData)
+    {
+        List<PackageableElement> elements = pureModelContextData.getElements();
+        Set<PackageableElement> elementsToCompose = new HashSet<>(elements);
+        Map<String, String> composedSections = Maps.mutable.empty();
+        if (ListIterate.anySatisfy(elements, e -> e instanceof SectionIndex))
+        {
+            Map<String, PackageableElement> elementByPath = new HashMap<>();
+            // NOTE: here we handle duplication, first element with the duplicated path wins
+            elements.forEach(element -> elementByPath.putIfAbsent(element.getPath(), element));
```
MauricioUyaguari, line 119: should we throw is there is duplicate elements?

legend-engine-language-pure-grammar/src/main/java/org/finos/legend/engine/language/pure/grammar/to/PureGrammarComposer.java
```diff
@@ -105,6 +107,55 @@ public String renderPureModelContextData(PureModelContextData pureModelContextDa
...
+                if (result != null)
+                {
+                    composedSections.put(element.getPath(), result.value);
+                    // mark that the elements already been rendered by one of the extensions
+                    result.composedElements.forEach(elementsToCompose::remove);
+                }
+            }
+        }
+
+        // FIXME: the following logic should be removed completely when we move to use extensions
```
MauricioUyaguari, line 139: bit confused here, how do now visit other grammars we define?

legend-engine-language-pure-grammar/src/main/java/org/finos/legend/engine/language/pure/grammar/from/mapping/PureInstanceClassMappingParseTreeWalker.java
```diff
@@ -107,7 +107,7 @@ private Lambda visitLambda(PureInstanceClassMappingParserGrammar.CombinedExpress
         int lineOffset = walkerSourceInformation.getLineOffset() + startLine - 1;
         // only add current walker source information column offset if this is the first line
         int columnOffset = (startLine == 1 ? walkerSourceInformation.getColumnOffset() : 0) + ctx.getStart().getCharPositionInLine();
-        ParseTreeWalkerSourceInformation combineExpressionSourceInformation = new ParseTreeWalkerSourceInformation.Builder(walkerSourceInformation.getSourceId(), lineOffset, columnOffset).withReturnSourceInfo(this.walkerSourceInformation.getReturnSourceInfo()).build();
+        ParseTreeWalkerSourceInformation combineExpressionSourceInformation = new ParseTreeWalkerSourceInformation.Builder(walkerSourceInformation.getSourceId(), lineOffset, columnOffset, walkerSourceInformation.getElementPath()).withReturnSourceInfo(this.walkerSourceInformation.getReturnSourceInfo()).build();
```
MauricioUyaguari, line 110: are we setting this so when there is a compilaton error we are able to locate the element path where it occured?

gayathrir11: yes

legend-engine-xt-graphQL-query/src/test/java/org/finos/legend/engine/query/graphQL/api/grammar/test/TestGrammarApi.java
```diff
@@ -137,5 +138,17 @@ public Function2<Map<String, ExecutableDocument>, RenderStyle, Response> jsonToG
     {
         return (a, b) -> graphQLGrammar.jsonToGrammarBatch(a, b, null);
     }
+
+    @Override
+    public Function5<Map<String, String>, String, Integer, Integer, Boolean, Response> grammarToJsonE()
```
akphi, line 143: can we spell these out to `Element` and rename those above to `Batch`?

legend-engine-shared-core/src/main/java/org/finos/legend/engine/shared/core/api/grammar/GrammarAPI.java
```diff
@@ -103,6 +121,19 @@ protected <T> Response jsonToGrammar(T graphFetchTree, RenderStyle renderStyle,
         }
     }
 
+    protected <T> Response jsonToGrammarWithElements(T graphFetchTree, RenderStyle renderStyle, Function2<T, RenderStyle, Map<String, String>> func, ProfileManager<CommonProfile> pm, String spanText)
```
akphi, line 124: let's rename `T graphFetchTree` to `input`, let's do it for the function above as well

legend-engine-shared-core/src/main/java/org/finos/legend/engine/shared/core/api/grammar/GrammarAPI.java
```diff
@@ -57,6 +58,23 @@ protected <T> Response grammarToJson(String text, Function<String, T> func, Prof
...
+                return ManageConstantResult.manageResult(profiles, data, objectMapper);
+            }
+            catch (Exception e)
+            {
+                return ExceptionTool.exceptionManager(e, LoggingEventType.TRANSFORM_GRAMMAR_TO_JSON_ERROR, Response.Status.BAD_REQUEST, profiles);
+            }
+        }
+    }
+
     protected <T> Response grammarToJsonBatch(Map<String, ParserInput> input, Function5<String, String, Integer, Integer, Boolean, T> func, Map<String, T> result, ProfileManager<CommonProfile> pm, String spanText)
```
akphi, line 78: why can't we reuse this method for our purpose? after all the input and output object shapes look right to me

legend-engine-shared-core/src/main/java/org/finos/legend/engine/shared/core/api/grammar/GrammarAPI.java
akphi: also, same comment as above about the need for this

legend-engine-protocol/src/main/java/org/finos/legend/engine/protocol/pure/v1/model/SourceInformation.java
```diff
@@ -41,6 +42,16 @@ public SourceInformation(String sourceId, int startLine, int startColumn, int en
         this.endColumn = endColumn;
     }
 
+    public SourceInformation(String sourceId, int startLine, int startColumn, int endLine, int endColumn, String elementPath)
```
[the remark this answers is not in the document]
akphi: Personally, I think `SourceInformation` is a very fundamental object that should be kept simple and in-sync with what we have in `Pure`, I don't want to meddle with this class, I think we can get away with that by creating a subtype of `SourceInformation`, for example:

```java
public class ElementBasedSourceInformation extends SourceInformation {...}

// and then we can produce this object in the `getSourceInformation()` method you added in `ParseTreeWalkerSourceInformation`
```

That way, we contain the `benefit/damage` to just our our use case for now.

Another thing to consider the unforseen cost we might add when we add a new field here, certain serialization setup might add a field to the resultant JSON, e.g. `"sourceInformation": {..., "elementPath": null }` which is not great.

legend-engine-language-pure-grammar/src/main/java/org/finos/legend/engine/language/pure/grammar/to/PureGrammarComposer.java
```diff
@@ -105,6 +107,55 @@ public String renderPureModelContextData(PureModelContextData pureModelContextDa
...
     }
 
+    public Map<String, String> renderPureModelContextDataForEachElement(PureModelContextData pureModelContextData)
+    {
+        List<PackageableElement> elements = pureModelContextData.getElements();
+        Set<PackageableElement> elementsToCompose = new HashSet<>(elements);
+        Map<String, String> composedSections = Maps.mutable.empty();
+        if (ListIterate.anySatisfy(elements, e -> e instanceof SectionIndex))
+        {
+            Map<String, PackageableElement> elementByPath = new HashMap<>();
```
akphi, line 117: wait, what's all this complexity above, you take the PMCD and you produce a `Map<String, String>`, section index is completely irrelevant here in terms of ordering, only good in terms of getting the `import` statements right? I expect the logic to be utterly simple here, as if `pureModelContextData.elements.forEach`, am I missing something? 🤷

akphi (changes requested):

legend-engine-language-pure-grammar/src/main/java/org/finos/legend/engine/language/pure/grammar/to/PureGrammarComposer.java
gayathrir11: yeah that makes more sense

legend-engine-shared-core/src/main/java/org/finos/legend/engine/shared/core/api/grammar/GrammarAPI.java
gayathrir11: We can't really use this function. As the output of function would return json for each grammar element. It would be of format Map<key, PMCD>. But that's not something we would want for our use case. If we receive result of this format we won't be able to send this as an input to compilation in studio. We need to join everthing before sending it as an input.
Another disadvantage is we get sectionIndex for every grammar element which is something we won't be able to compile.

legend-engine-language-pure-grammar/src/main/java/org/finos/legend/engine/language/pure/grammar/from/PureGrammarParser.java
```diff
@@ -129,14 +136,68 @@ private PureModelContextData parse(String code, DEPRECATED_PureGrammarParserLibr
         return builder.withElement(sectionIndex).build();
     }
 
+    private PureModelContextData parse(Map<String, String> code, DEPRECATED_PureGrammarParserLibrary parserLibrary, String sourceId, int lineOffset, int columnOffset, boolean returnSourceInfo)
```
akphi, line 139: let's just rename this method to `parseElements`, and remove the other method, also, in this particular case, does it even make sense to allow specifying `sourceId` and `lineOffset`, `columnOffset`?

akphi, line 139: let's just rename this method to `parseElements`, and remove the other method, also, in this particular case, does it even make sense to allow specifying `sourceId` and `lineOffset`, `columnOffset`?

legend-engine-language-pure-grammar/src/main/java/org/finos/legend/engine/language/pure/grammar/to/PureGrammarComposer.java
```diff
@@ -105,6 +107,55 @@ public String renderPureModelContextData(PureModelContextData pureModelContextDa
         return composedSections.select(section -> !section.isEmpty()).makeString("\n\n");
     }
 
+    public Map<String, String> renderPureModelContextDataForEachElement(PureModelContextData pureModelContextData)
```
akphi, line 110: These method still seem rather clunky to me. Will it be simpler if we do something like:
1. Read the section index if any, transform section index into 1 section per element (with the import of the whole section)
2. Create an index of element path by section (we would need to handle duplicate somehow)
3. Go through that index and compose each element

I felt like the logic is much simpler that way

Also, since we're doing it, there's a case that we haven't thought about where users switch from single text-mode to multiple-file text-mode. When they go back, their sections would be broken down completely and import paths would be duplicated every where. What should we do here? @gayathrir11 and @MauricioUyaguari ?

akphi (changes requested):

PR #1256, opened 2022-12-16
Spanner enable

roasher:
#### What type of PR is this?
Enabling spanner connector functionality

After this change spanner would be legal db for studio projects

this PR contains shading google spanner jdbc and relocation of all it's compiled classes to custom package, so there would be no dependency conflicts with other part of platform

copy of https://github.com/finos/legend-engine/pull/1135 with java doc jar sonatype problem fix

No, just adding new db type among others


vikask-gs (approved):

roasher:
@gs-ssh16 
1) Problem with https://github.com/finos/legend-engine/actions/runs/3710941855/jobs/6290890928 is due to fact that we run 2 maven tasks instead of one:
```
mvn -B -e install -DskipTests=true
mvn -B -e javadoc:javadoc deploy -DargLine="-Xmx2200m" -P docker-snapshot
```
And there is no "clean" on second one. First build runs successfully, second fails.
This is due to shade plugin bug: https://github.com/elastic/apm-agent-java/issues/961. This was fixed in September, so I did plugin update and checked in my local env and it ok now.

2) Release build problem https://github.com/finos/legend-engine/actions/runs/3711380363/jobs/6291747703 is fixed via adding blank JavaPlaceholder.java file. Although it looks like a hack, but I couldn't find much better solution and it seems that this approach is already used by Pierre: https://github.com/finos/legend-pure/blob/4327c2a3b8dacabba73b3c80693760036ff55675/legend-pure-m3-platform/src/main/java/org/finos/legend/pure/m3/platform/JavaPlaceholder.java#L17

davidharte-gs (approved):

PR #1269, opened 2022-12-23
Refresh connection if eviction duration reached during access

vikask-gs:
#### What type of PR is this?

Connection Refresh Strategy

Refresh connection if eviction duration reached during access. Currently we evict stale connections only via a separate thread which runs once every "eviction duration" interval (this can potentially leave us with connections which are stale, but not evicted/refreshed).

Yes. It should help us with certain cases, where users are facing issues with stale connection when executing services.

### DIFF
--- a/legend-engine-xt-relationalStore-executionPlan-connection/src/main/java/org/finos/legend/engine/plan/execution/stores/relational/connection/ds/state/ConnectionStateManager.java
+++ b/legend-engine-xt-relationalStore-executionPlan-connection/src/main/java/org/finos/legend/engine/plan/execution/stores/relational/connection/ds/state/ConnectionStateManager.java
@@ -119,6 +119,7 @@ public class ConnectionStateManager implements Closeable
     private static final String SEPARATOR = "_";
     private static final String DBPOOL = "DBPool_";
     private static ConnectionStateManager INSTANCE;
+    private static long evictionDurationInSeconds;
 
     static
     {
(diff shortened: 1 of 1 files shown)


legend-engine-xt-relationalStore-executionPlan-connection/src/main/java/org/finos/legend/engine/plan/execution/stores/relational/connection/ds/state/ConnectionStateManager.java
```diff
@@ -356,13 +357,14 @@ public DataSourceWithStatistics getDataSourceForIdentityIfAbsentBuild(IdentitySt
...
+        DataSourceWithStatistics dataSourceWithStatistics = this.connectionPools.get(poolName);
+        if (!dataSourceWithStatistics.getIdentityState().isValid() || (dataSourceWithStatistics.getStatistics().getLastConnectionRequestAge() > (evictionDurationInSeconds * 1000) && !dataSourceWithStatistics.hasActiveConnections()))
         {
             LOGGER.info("Pool [{}] for datasource [{}] does not have a valid identity state", principal, connectionKey.shortId());
             synchronized (poolLockManager.getLock(poolName))
             {
-                DataSourceWithStatistics dataSourceWithStatistics = this.connectionPools.get(poolName);
-                if (!dataSourceWithStatistics.getIdentityState().isValid())
+                dataSourceWithStatistics = this.connectionPools.get(poolName);
+                if (!dataSourceWithStatistics.getIdentityState().isValid() || (dataSourceWithStatistics.getStatistics().getLastConnectionRequestAge() > (evictionDurationInSeconds * 1000) && !dataSourceWithStatistics.hasActiveConnections()))
```
elopezcastro, line 367: as discussed this is not the right approach, you need to specify the expiry logic for the datasource so we can get rid of pools with expired or invalid credentials

PR #1273, opened 2022-12-28
Throw warning if both legacy and strategic tests are used

hardikmaheshwari:
#### What type of PR is this?
Error Reporting

Throw warning if both legacy and strategic tests are used

Fixes #

yes, better warning for user


hardikmaheshwari:
> Why is this a warning as opposed to an error?

To be backward compatible. Plan to start with warning(deprecate) and then demise. Does it make sense?

Yasirmod17:
> 
Is there any user that has both legacy and strategic today ? (I don’t see how they would have gotten into that state) If there isn’t then we should consider upgrading to an error from the start so no one thinks of using both.

hardikmaheshwari:
> > 
> 
> Is there any user that has both legacy and strategic today ? (I don’t see how they would have gotten into that state) If there isn’t then we should consider upgrading to an error from the start so no one thinks of using both.

There are actually

Yasirmod17 (approved):

MauricioUyaguari (approved):

PR #1276, opened 2023-01-02
WIP

epsstan:

Fixes #


rafaelbey:
This is awesome!  

One thing I notice is that you assume Subject, using SubjectTool.  Could that be too narrow.  

Wonder if we should delegate to something like IdentityFactory, or something configurable/provided, so that if the main authentication of a given deployment is a SSO token, they could also handle how to extract the user name out of it.

Thoughts?

finos-admin:

PR #1292, opened 2023-01-09
Persistence: Test Restructuring and Bug Fixes for Bitemporal Delta 

prasar-ashutosh:
#### What type of PR is this?
Test Restructuring and Bug Fix

1) Test Restructuring for Bitemporal Delta
2) Fix the bug about automatic case conversion for aliases 

1) Add validation while loading test data in H2
2) Define test scenarios and test cases for Bitemporal Delta
3) Add Bitemporal Delta tests in Memsql and Snowflake Modules
4) Fix the alias converting to upper case issue for Bitemporal Delta
5) Support upper and lower case optimisation for alias

NO


rafaelbey (changes requested):
This UpperCaseOptimizer and LowerCaseOptimizer seems to be duplicated code with the different of the operation on the string.

Can we create an abstract class StringCaseOptimizer, where it receive in the constructor the strategy to apply, and then Upper and Lower just set the right strategy.

```java
public StringCaseOptimizer(Funtion<String, String> strategy){
  this.strategy = strategy;
}
```
Then, operations will look like this:

`field.setAlias(this.strategy.apply(fielld.getAlias()));`


The upper and lower class will only need a constructor:

```java
public UpperCaseOtmizer()
{
   super(String::toUpperCase);
}
```

rafaelbey:
Besides that duplication I highlighted on the prev comment, the rest looks good.

prasar-ashutosh:
> This UpperCaseOptimizer and LowerCaseOptimizer seems to be duplicated code with the different of the operation on the string.
> 
> Can we create an abstract class StringCaseOptimizer, where it receive in the constructor the strategy to apply, and then Upper and Lower just set the right strategy.
> 
> ```java
> public StringCaseOptimizer(Funtion<String, String> strategy){
>   this.strategy = strategy;
> }
> ```
> 
> Then, operations will look like this:
> 
> `field.setAlias(this.strategy.apply(fielld.getAlias()));`
> 
> The upper and lower class will only need a constructor:
> 
> ```java
> public UpperCaseOtmizer()
> {
>    super(String::toUpperCase);
> }
> ```



> This UpperCaseOptimizer and LowerCaseOptimizer seems to be duplicated code with the different of the operation on the string.
> 
> Can we create an abstract class StringCaseOptimizer, where it receive in the constructor the strategy to apply, and then Upper and Lower just set the right strategy.
> 
> ```java
> public StringCaseOptimizer(Funtion<String, String> strategy){
>   this.strategy = strategy;
> }
> ```
> 
> Then, operations will look like this:
> 
> `field.setAlias(this.strategy.apply(fielld.getAlias()));`
> 
> The upper and lower class will only need a constructor:
> 
> ```java
> public UpperCaseOtmizer()
> {
>    super(String::toUpperCase);
> }
> ```

Thanks's Rafael.
That's a great suggestion.
Made the code changes using the suggested approach.
Also added applyCase method to reduce some more duplication.

rafaelbey (approved):
Thanks for the refactoring!

PR #1328, opened 2023-01-21
Fix first join JSON to grammar bug

xannem:
#### What type of PR is this?
- Bug Fix

When trying to translate a first join as an INNER join, the returned string causes an unexpected token error due to structure. 

Fixes https://github.com/finos/legend-studio/issues/1826

### DIFF
--- a/legend-engine-xt-relationalStore-grammar/src/main/java/org/finos/legend/engine/language/pure/grammar/to/HelperRelationalGrammarComposer.java
+++ b/legend-engine-xt-relationalStore-grammar/src/main/java/org/finos/legend/engine/language/pure/grammar/to/HelperRelationalGrammarComposer.java
@@ -237,7 +237,7 @@ public class HelperRelationalGrammarComposer
         StringBuilder builder = new StringBuilder();
         if (!elementWithJoins.joins.isEmpty())
         {
-            JoinPointer firstJoinPointer = (JoinPointer) elementWithJoins.joins.get(0);
+            JoinPointer firstJoinPointer = elementWithJoins.joins.get(0);
             builder.append(renderFirstJoinPointer(firstJoinPointer));
             if (elementWithJoins.joins.size() > 1)
             {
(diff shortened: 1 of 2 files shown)


legend-engine-xt-relationalStore-grammar/src/main/java/org/finos/legend/engine/language/pure/grammar/to/HelperRelationalGrammarComposer.java
```diff
@@ -237,7 +237,13 @@ private static String renderElementWithJoins(ElementWithJoins elementWithJoins,
         StringBuilder builder = new StringBuilder();
         if (!elementWithJoins.joins.isEmpty())
         {
-            builder.append(LazyIterate.collect(elementWithJoins.joins, HelperRelationalGrammarComposer::renderJoinPointer).makeString(" > "));
+            String joins = LazyIterate.collect(elementWithJoins.joins.subList(0,1), HelperRelationalGrammarComposer::renderFirstJoinPointer).makeString(" > ");
```
MauricioUyaguari, line 240: Could we make this more readable. Isn't this really just on join so the `makeString(" > ")` doesn't really come into effect?

Something like 
`
JoinPointer firstJoinPointer = ..
builder.append(renderFirstJoin(firstJoinPointer). 

... then continue if you if statement below. You dont need to concatenate. Leverage the string builder to append as needed.

legend-engine-xt-relationalStore-grammar/src/main/java/org/finos/legend/engine/language/pure/grammar/to/HelperRelationalGrammarComposer.java
```diff
@@ -486,6 +492,13 @@ private static String renderViewColumnMapping(ColumnMapping columnMapping, List<
         return builder.toString();
     }
 
+    private static String renderFirstJoinPointer(JoinPointer joinPointer)
```
MauricioUyaguari, line 495: lets add a comment here explaining why the first join pointer is different. 
We do this because the first join pointer requires the database pointer to be first  inside the main join operator. 
See: https://github.com/finos/legend-engine/blob/master/legend-engine-xt-relationalStore-grammar/src/main/antlr4/org/finos/legend/engine/language/pure/grammar/from/antlr4/RelationalParserGrammar.g4#L197

MauricioUyaguari (changes requested):

MauricioUyaguari (approved):

PR #1337, opened 2023-01-23
activate Reproducible Builds

hboutemy:
Signed-off-by: Hervé Boutemy <hboutemy@apache.org>

#### What type of PR is this?

enhancement: Reproducible Builds

see https://maven.apache.org/guides/mini/guide-reproducible-builds.html

Fixes #


hboutemy:
the CLA verification workflow seems broken: https://github.com/finos/open-developer-platform/pull/350 says everything is ok, but this does not reflect here

please help

pom.xml
```diff
@@ -457,6 +458,11 @@
                     <artifactId>maven-checkstyle-plugin</artifactId>
                     <version>${maven.checkstyle.plugin.version}</version>
                 </plugin>
+                <plugin>
+                    <groupId>org.apache.maven.plugins</groupId>
+                    <artifactId>maven-release-plugin</artifactId>
+                    <version>3.0.0-M7</version>
```
maoo, line 464: This should not be needed, since we inherit it from the FINOS Parent POM, see https://github.com/finos/legend-engine/blob/master/pom.xml#L23

Also note that 3.0.0-M7 breaks integration with maven-gpg-plugin, see https://github.com/apache/maven-release/pull/125#issuecomment-1302613138 ; better to stick with M5 version.

maoo:
> the CLA verification workflow seems broken: [finos/open-developer-platform#350](https://github.com/finos/open-developer-platform/pull/350) says everything is ok, but this does not reflect here
> 
> please help

Thanks for your contribution @hboutemy ! Legend haven't switched yet to EasyCLA, so we still need to onboard contributors in our previous CLA Bot system; we're working on it, should be done by the end of today.

TheJuanAndOnly99:
@cla-bot[bot] check

TheJuanAndOnly99:
@hboutemy you are all set regarding the CLA check

pom.xml
hboutemy: yes, now that parent POM 5 is released (thank you), it's better to upgrade parent

hboutemy:
@maoo parent updated, now the update is better scoped

maoo:
@beekemarie @epsstan @ivan-kyosev-gs - could you please have a look at this PR ? Hervé has kindly worked on making the Legend Engine build "reproducible", you can read more on http://reproducible-builds.org/ ; this is a great opportunity enforce Supply Chain security on Legend, and the idea is to extend this approach across all other Legend repositories.

Eager to hear your thoughts. Thanks!

hboutemy:
@beekemarie @epsstan @ivan-kyosev-gs please review

kevin-m-knight-gs (approved):
I resolved one conflict. Otherwise, this looks good.

kevin-m-knight-gs:
Thanks for the contribution!

akphi:
@hboutemy Thank you for your contribution, we notice that after the upgrade, our release time almost doubled:
Version 4.4.4: https://github.com/finos/legend-engine/actions/runs/4507569197 (build time 1h20m)
Version 4.4.5: https://github.com/finos/legend-engine/actions/runs/4517314220 (build time 2h37m)

I have downloaded the log and do a diff to compute the time, I realized that for some reason version 3 of `maven-release-plugin` no longer allow parallelism for the `release:perform` goal - https://maven.apache.org/maven-release/maven-release-plugin/perform-mojo.html

We need to roll back for now as this gain in time is quite detrimental for us as it translate to actual cost due to usage of large runner. Having said that, we really appreciate your contribution and the value of reproducible builds 🙏 , hopefully someone could find a way to solve the parallelism problem with `maven-release-plugin@3` and we could move forward.

hboutemy:
@akphi thanks for the info: sorry if that caused that issue, I'm as surprised as you
I'll keep that as a good learning and I'll try to work on this

PR #1340, opened 2023-01-23
add Spans for Test Suite

janeenyamak1:
This add spans to Test Suites so users will be able to see what tests/asserts pass/fail when they register their service

Fixes #
This gives users more awareness around what test(s) are passing/failing

### DIFF
--- a/legend-engine-test-runner-service/src/main/java/org/finos/legend/engine/testable/service/extension/ServiceTestRunner.java
+++ b/legend-engine-test-runner-service/src/main/java/org/finos/legend/engine/testable/service/extension/ServiceTestRunner.java
@@ -15,6 +15,9 @@
 
 package org.finos.legend.engine.testable.service.extension;
 
+import io.opentracing.Scope;
+import io.opentracing.Span;
+import io.opentracing.util.GlobalTracer;
 import org.eclipse.collections.api.RichIterable;
 import org.eclipse.collections.api.factory.Lists;
 import org.eclipse.collections.api.factory.Maps;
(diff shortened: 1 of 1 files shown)


legend-engine-test-runner-service/src/main/java/org/finos/legend/engine/testable/service/extension/ServiceTestRunner.java
```diff
@@ -297,8 +301,9 @@ private org.finos.legend.engine.protocol.pure.v1.model.test.result.TestResult ex
 
         SerializationFormat testSerializationFormat = getSerializationFormatForTest(serviceTest);
 
-        try
+        try (Scope scope = GlobalTracer.get().buildSpan("Test Status for: " + serviceTest.id).startActive(true))
```
Yasirmod17, line 304: can we ensure we are consistent with the case of characters

legend-engine-test-runner-service/src/main/java/org/finos/legend/engine/testable/service/extension/ServiceTestRunner.java
```diff
@@ -347,6 +353,22 @@ private org.finos.legend.engine.protocol.pure.v1.model.test.result.TestResult ex
 
                 testResult = testFailed;
                 testResult.atomicTestId = atomicTestId;
+
+                testFailed.assertStatuses.forEach(assertion ->
+                {
+                    if (assertion instanceof EqualToJsonAssertFail)
+                    {
+                       span.log("ASSERTION: " + assertion.id + " FAILED : {\"expected\": " + ((EqualToJsonAssertFail)assertion).expected + ", \"actual\": " + ((EqualToJsonAssertFail)assertion).actual + "}");
```
Yasirmod17, line 361: we should log these separately. 
Span.log(assert failed..)
Span.log(expected:..)
Span.log(actual:..) 
instead of building a json

PR #1364, opened 2023-02-01
Add store execution tracker

AFine-gs:
#### What type of PR is this?
Improvement

Introduce a new feature that enables the management of running executions. 

No


davidharte-gs (approved):

legend-engine-executionPlan-execution/pom.xml
```diff
@@ -110,6 +110,11 @@
             <groupId>org.finos.legend.engine</groupId>
             <artifactId>legend-engine-executionPlan-dependencies</artifactId>
         </dependency>
+        <dependency>
+            <groupId>javax.servlet</groupId>
+            <artifactId>javax.servlet-api</artifactId>
+            <scope>compile</scope>
+        </dependency>
```
rafaelbey, line 117: I dont think we should expose HTTP concepts into the execution plan.  Is there way to design this such that you can expose the data you want while keeping this module clean of this depdency?

legend-engine-executionPlan-execution/src/main/java/org/finos/legend/engine/plan/execution/stores/StoreExecutableManager.java
```diff
@@ -0,0 +1,104 @@
...
+import javax.servlet.http.HttpSessionEvent;
+import javax.servlet.http.HttpSessionListener;
+
+
+/**
+ * This is a singleton that maintains the relationship between  http sessions and store executions.
+ * It can be used to operate on executions (for example to cancel running executions)
+ * The manager must be initialized as well as added as a HttpSessionListener
+ */
+public enum StoreExecutableManager implements HttpSessionListener
```
rafaelbey, line 33: Can we avoid the singleton here, and just let the sever create the one instance we need, and propagate it accordantly?  Or even further, do we need to this inside the Result class?  Cannot we do this management on the Server resource that will end serializing the result?

AFine-gs: The singleton is intended to ensure there is only one instance of the manager created. That is critical to ensure the state of running executions is tracked correctly.

legend-engine-executionPlan-execution/src/main/java/org/finos/legend/engine/plan/execution/stores/StoreExecutableManager.java
```diff
@@ -0,0 +1,97 @@
...
+    }
+
+    public void registerManager()
+    {
+        this.isRegistered = true;
+    }
+
+     public void cancelExecutablesOnSession(String sessionID)
+    {
+        sessionExecutableMap.get(sessionID).toList().stream().forEach(
```
rafaelbey, line 81: We can make things little more efficient by just removing all the executables in one call?

`sessionExecutableMap.removeAll(sessionID).stream().forEach()`

legend-engine-executionPlan-execution/src/main/java/org/finos/legend/engine/plan/execution/stores/StoreExecutableManager.java
```diff
@@ -0,0 +1,97 @@
...
+
+/**
+ * This is a singleton that maintains the relationship between  http sessions and store executions.
+ * It can be used to operate on executions (for example to cancel running executions)
+ * The manager must be initialized as well as added as a HttpSessionListener
+ */
+public enum StoreExecutableManager
+{
+    INSTANCE;
+    private MutableMultimap<String, StoreExecutable> sessionExecutableMap = new FastListMultimap<>();
```
rafaelbey, line 33: This collection, or the methods accessing it should guarantee thread safety.

rafaelbey (approved):

PR #1366, opened 2023-02-02
Support parallel execution of nodes and enable parallel execution for union nodes

hardikmaheshwari:
#### What type of PR is this?
Improvement 

Support parallel execution of nodes and enable parallel execution for union nodes

Fixes #


legend-engine-executionPlan-execution/src/main/java/org/finos/legend/engine/plan/execution/concurrent/ConcurrentExecutionNodeExecutorPool.java
```diff
@@ -0,0 +1,158 @@
...
+import java.util.concurrent.ThreadPoolExecutor;
+import java.util.concurrent.atomic.AtomicReference;
+
+public final class ConcurrentExecutionNodeExecutorPool
+{
+    private static final AtomicReference<ConcurrentExecutionNodeExecutorPool> INSTANCE = new AtomicReference<>();
+
+    private final int poolSize;
+    private final ThreadPoolExecutor executor;
+    private final Semaphore availableThreads;
```
rafaelbey, line 48: I think we could avoid the semaphore, and set the setRejectedExecutionHandler to `CallerRunsPolicy`, such that if the pool is busy, the caller thread will handle it.  This simplifies too the code around gathering the semaphor and executing locally when cannot acquire.  

Thoughts?

legend-engine-executionPlan-execution/src/main/java/org/finos/legend/engine/plan/execution/concurrent/ConcurrentExecutionNodeExecutorPool.java
```diff
@@ -0,0 +1,158 @@
...
+    }
+
+    public static ConcurrentExecutionNodeExecutorPool getExecutorPool()
+    {
+        return INSTANCE.get();
+    }
+
+    public List<? extends Result> execute(final List<ExecutionNode> nodes, final MutableList<CommonProfile> profiles, final ExecutionState executionState)
+    {
+        if (INSTANCE.get() != null && availableThreads.tryAcquire(nodes.size()))
```
rafaelbey, line 87: why do an INSTANCE.get() here, when you already inside the pool instance?

legend-engine-shared-core/src/main/java/org/finos/legend/engine/shared/core/operational/opentracing/TracedSupplier.java
```diff
@@ -0,0 +1,73 @@
...
+import io.opentracing.Span;
+import io.opentracing.Tracer;
+import io.opentracing.util.GlobalTracer;
+
+import java.util.function.Supplier;
+
+/**
+ * Consider tracing the executor service rather than the supplier via opentracing-contrib/java-concurrent
+ */
+public class TracedSupplier<T> implements Supplier<T>
```
rafaelbey, line 27: Did you look at the classes inside package `io.opentracing.contrib.concurrent`.  They seem to already handle concurrency on tracing.  Like, you can wrap the threadpoolexecutor on `TracedExecutorService` to handle this  seamlessly for y ou.

rafaelbey (commented):
This is very impressive change!  Made few comments to see simplify code even further by taking advantage of existing tooling.  

One last comment - we should also think how we could collect metric on this pool.  Something like DropWizard metric library provide wrappers to expose useful execution data.

legend-engine-shared-core/src/main/java/org/finos/legend/engine/shared/core/operational/opentracing/TracedSupplier.java
hardikmaheshwari: Thanks, this makes sense updated the PR

legend-engine-executionPlan-execution/src/main/java/org/finos/legend/engine/plan/execution/concurrent/ConcurrentExecutionNodeExecutorPool.java
hardikmaheshwari: Moved away from Singleton pattern so no longer need this

legend-engine-executionPlan-execution/src/main/java/org/finos/legend/engine/plan/execution/concurrent/ConcurrentExecutionNodeExecutorPool.java
hardikmaheshwari: This completely makes sense. Though for 1st pass we want design to be on defensive side a bit and not combine execution with parallel pool thread and current thread. Though totally agree this would be a good optimization

hardikmaheshwari:
> This is very impressive change! Made few comments to see simplify code even further by taking advantage of existing tooling.
> 
> One last comment - we should also think how we could collect metric on this pool. Something like DropWizard metric library provide wrappers to expose useful execution data.

Thanks a lot for reviewing. Added a endpoint to expose details around state of pool

PR #1368, opened 2023-02-03
scalar Date, DateTime, BigDecimal in GraphQL schema generation

abhishoya-gs:

Fixes #

### DIFF
--- a/legend-engine-xt-graphQL-pure/src/main/resources/core_external_query_graphql/deprecated/generation.pure
+++ b/legend-engine-xt-graphQL-pure/src/main/resources/core_external_query_graphql/deprecated/generation.pure
@@ -15,6 +15,8 @@
 import meta::external::query::graphQL::generation::*;
 import meta::pure::generation::metamodel::*;
 import meta::pure::generation::*;
+import meta::pure::metamodel::type::*;
+import meta::pure::functions::collection::*;
 
 Class <<doc.deprecated>> meta::external::query::graphQL::generation::GraphQLConfig extends GenerationConfiguration
 {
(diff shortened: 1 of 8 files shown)


legend-engine-xt-graphQL-pure/src/main/resources/core_external_query_graphql/deprecated/generation.pure
```diff
@@ -38,7 +38,13 @@ function <<doc.deprecated>> meta::external::query::graphQL::generation::generate
 {
    $config.scopeElements->filter(e| !$e->instanceOf(Package))->map(e|assert($e->instanceOf(Type),' GraphQL generation of ' + $e->type().name->toOne()+ ' is not currently supported. Tried to generate GraphQL for '+ $e->elementToPath()) ;);
    let scopeElements = $config.allPackageScopeElements()->filter(p|$p->instanceOf(Class))->cast(@Class<Any>);
-   $scopeElements->map(c| $config->meta::external::query::graphQL::generation::generateGraphQLFromPure($c));
+   $scopeElements->map(c| $config->meta::external::query::graphQL::generation::generateGraphQLFromPure($c))->concatenate(
```
epsstan, line 41: Couple of comments ...

1/ Let's not generate a custom scalar type definition if the corresponding Pure type is not in the generation scope 

2/ Separately, we should not be inlining logic to compute a GraphQLOutput in generateGraphQL. Let's move into a separate function

epsstan (changes requested):

legend-engine-xt-graphQL-pure/src/main/resources/core_external_query_graphql/deprecated/generation.pure
abhishoya-gs: done

epsstan (approved):

PR #1386, opened 2023-02-13
External Format Contract Refactor

hardikmaheshwari:
#### What type of PR is this?
Improvement

External Format Contract Refactor, this PR brings in 2 changes
- Deleting `SchemaDetail`: Currently we are enforcing each external format to define a schema metamodel which should extend SchemaDetail. With this change, we are removing this so that external format developers don't have to know the internals of the platform to model their schema metamodel. In essence, we are relaxing this in order to make integration more smooth.
- Changing ExternalFormatContract: We are replacing property `sourceRecordTree` with properties `internalizeReturnsChecked` & `sourceRecordSerializeTree`. Again checked is an internal platform concern and in the happy path, external format developers should not have to worry about it. Moving forward if a developer wants to provide record-level lineage and defect reporting they can integrate with the checked framework by providing `sourceRecordSerializeTree` and setting `internalizeReturnsChecked` as `true`

Fixes #

No


MauricioUyaguari:
Can we add a description on the key takeaways from the refactor and why we are doing for example SchemaDetail -> Any . 
Thanks

hardikmaheshwari:
> Can we add a description on the key takeaways from the refactor and why we are doing for example SchemaDetail -> Any . Thanks

Thanks a lot for pointing this out, have updated this PR will try to update other PR's in future

epsstan (approved):

PR #1400, opened 2023-02-17
Update Snowflake/Postgres workflows to upload test results as Github artifacts

epsstan:
#### What type of PR is this?
Improvement

Publishes test results from the Snowflake/Postgres  workflows to Github actions.
This is so that other workflow actions can download these results (artifacts) and assemble reports etc. 

We are already doing this in a couple of places -
https://github.com/finos/legend-engine/blob/master/.github/workflows/test-result.yml 
https://github.com/finos/legend-engine/blob/master/.github/workflows/summarize-sql-generation-integration-tests.yml 

This PR publishes the full/raw Junit output instead of a summary. 

Once we get the report generation working for Snowflake/Postgres, we can roll out to the other database workflows and also cleanup other summary artifacts that the workflows are currently publishing.

Fixes #

No


epsstan:
@rafaelbey @ivan-kyosev-gs Can you please take a look ?

PR #1401, opened 2023-02-17
Update GraphQL API endpoints with user/group workspace parameters.

epsstan:
#### What type of PR is this?
Bug Fix

This PR allows users to use the dev GraphQL endpoints with both user and group SDLC workspaces. 
Previously, the code only supported user workspaces (isGroupWorkspace was hardcoded to false). 

Now there are 2 new API endpoints 
```
/workspace/{wsId}
/groupWorkspace/{wsId}
```
that are routed to the appropriate SDLC APIs to load models. 

In addition, the previous API is being deprecated and will be removed in the next release of engine.

Fixes #

Yes. New APIs are being introduced.  Existing APIs have been deprecated but not deleted.


epsstan:
@abhishoya-gs  - Can you please take a look ? We should get this into the next release.

gs-ssh16 (approved):

PR #1448, opened 2023-02-28
Support parallel execution of platform union nodes

hardikmaheshwari:
#### What type of PR is this?
Improvement

Support parallel execution of platform union nodes

Fixes #


legend-engine-executionPlan-execution/src/main/java/org/finos/legend/engine/plan/execution/concurrent/ConcurrentExecutionNodeExecutorPool.java
```diff
@@ -0,0 +1,157 @@
...
+import java.util.concurrent.CompletableFuture;
+import java.util.concurrent.ExecutorService;
+import java.util.concurrent.Executors;
+import java.util.concurrent.Semaphore;
+
+/**
+ * Pool management class. This class should be instantiated only during server spin up to help manage thread pool and guard system against thread explosions.
+ */
+@JsonSerialize(using = ConcurrentExecutionNodeExecutorPoolSerializer.class)
+public final class ConcurrentExecutionNodeExecutorPool implements AutoCloseable
```
rafaelbey, line 48: Not a blocker, but a trick worth knowing - you could implement [JsonSerializable](https://fasterxml.github.io/jackson-databind/javadoc/2.6/com/fasterxml/jackson/databind/JsonSerializable.html) interface, sp Jackson will then know how to serialize the object without the need of the custom serializer.

rafaelbey (approved):

legend-engine-executionPlan-execution/src/main/java/org/finos/legend/engine/plan/execution/concurrent/ConcurrentExecutionNodeExecutorPool.java
hardikmaheshwari: hmm thanks for letting me know, this seems interesting and efficient. Just checking through the codebase it seems we have been using annotations and not implementing interface. Plan to keep this as it is for now but interface seems better aproach

PR #1449, opened 2023-02-28
Add logic for returning default incoming credential in CredentialProvider

jinanisha:
#### What type of PR is this?
Improvement

If the incoming credential type is same as output credential type, we can directly return the incoming credential. There is no need to look for intermediation rules. It will reduce the effort to add default intermediation rules for every authentication spec.

Fixes #

No


rafaelbey (commented):
Is there a test we can add to cover this change?

epsstan (approved):

PR #1454, opened 2023-03-01
Support chaining M2M transformations with binding queries

hardikmaheshwari:
#### What type of PR is this?
Improvement

Support chaining M2M transformations with binding queries

Fixes #


gs-ssh16 (approved):

legend-engine-pure-code-compiled-core/src/main/resources/core/pure/router/store/routing.pure
```diff
@@ -230,7 +230,7 @@ function  meta::pure::router::store::routing::specializedFunctionExpressionRoute
                                                | getRoutingStrategyFromMappingAndRuntime($fromMapping, $fromRuntime));
                 let newState               = ^$state(routingStrategy = $newRoutingStrategy);
 
-                let processedFunction      = routeFunctionExpressionFunctionDefinition($f, $fe, $newState, $fromExecutionContext, $vars, $inScopeVars, $extensions, $debug);
+                let processedFunction      = routeFunctionExpressionFunctionDefinition($fe.parametersValues->at(0)->cast(@FunctionExpression).func, $fe.parametersValues->at(0)->cast(@FunctionExpression), $newState, $fromExecutionContext, $vars, $inScopeVars, $extensions, $debug);
```
rafaelbey, line 233: Some of my local development stop working after this commit, and I think the problem is on this line.  Can you explain the rational for this?

hardikmaheshwari: The logic here is to enable special routing for `from` functions. The strategy was to intercept from functions in our queries and extract runtime/mapping from the query and route the left part of the expression in that scope. This is what this change ensures earlier we were routing mapping/runtime too which should not be done. Hope that clarifies.

rafaelbey: Here are some test cases that were passing before this change:

https://github.com/rafaelbey/legend-engine/blob/es-store/legend-engine-xt-elasticsearch/legend-engine-xt-elasticsearch-executionPlan-test/src/main/resources/core_elasticsearch_execution_test/elasticsearch_plan_test.pure#L191

Full review: https://github.com/finos/legend-engine/pull/1485

The query function looks as this:

`|indexToTds($store, 'indexName')->from(EmptyMapping, $runtime)->project(...)`

_indexToTds_ is similar as tableToTds but for Elasticsearch.

And after this change, rather that routing/clustering using the ES store extension, things ends on platform execution nodes.  Maybe Im missing something on the preval setup for this function?

rafaelbey: I the issue is that this change ignore that the function might be handled by specialiedFunctionExpressionRoutedProcessor.  Let me see if I can figure out the fix.

rafaelbey: See my proposed fix here: https://github.com/finos/legend-engine/pull/1485/files#diff-c770886a723f60a339a0710398b7c2a6c3917b7f371d4070cc5d76adab033def

I still send the from function, but remove the parameters that have already been processed/routed.

PR #1472, opened 2023-03-03
Refactor to use snowflake bouncycastle in EncryptedPrivateKeyFromVaultProvider

jinanisha:
#### What type of PR is this?

Improvement

Refactor to use snowflake bouncycastle in EncryptedPrivateKeyFromVaultProvider

Fixes #

No


rafaelbey:
Will this work within a JVM that has security features enable (ie like when validating signed jars are valid)

epsstan:
As discussed, we should be moving away from the snowflake driver for this piece of code that is not specific to Snowflake.

PR #1475, opened 2023-03-03
User friendly error for slice, drop and limit/take parameters

abhishoya-gs:
#### What type of PR is this?
Improvement in errors

Currently upon using a dyna function inside slice, drop and limit/take we get a cast exception. This MR introduces user friendly errors.

Fixes #

yes, message displayed on a specific error

### DIFF
--- a/legend-engine-xt-relationalStore-pure/src/main/resources/core_relational/relational/pureToSQLQuery/pureToSQLQuery.pure
+++ b/legend-engine-xt-relationalStore-pure/src/main/resources/core_relational/relational/pureToSQLQuery/pureToSQLQuery.pure
@@ -2383,9 +2383,12 @@ function meta::relational::functions::pureToSqlQuery::processSlice(f:FunctionExp
    let mainQuery = processValueSpecification($f.parametersValues->at(0), $currentPropertyMapping, $operation, $vars, $state, $joinType, $nodeId, $aggFromMap, $context, $extensions)->toOne()->cast(@SelectWithCursor);
    let mainSelect = $mainQuery.select;
 
-   let fromRow = processValueSpecification($f.parametersValues->at(1), $currentPropertyMapping, $operation, $vars, $state, JoinType.LEFT_OUTER, $nodeId, $aggFromMap, $context, $extensions)->toOne()->cast(@SelectWithCursor).select.columns->at(0)->cast(@Literal);
-   let toRow = processValueSpecification($f.parametersValues->at(2), $currentPropertyMapping, $operation, $vars, $state, JoinType.LEFT_OUTER, $nodeId, $aggFromMap, $context, $extensions)->toOne()->cast(@SelectWithCursor).select.columns->at(0)->cast(@Literal);
-
+   let param1 = processValueSpecification($f.parametersValues->at(1), $currentPropertyMapping, $operation, $vars, $state, JoinType.LEFT_OUTER, $nodeId, $aggFromMap, $context, $extensions)->toOne()->cast(@SelectWithCursor).select.columns->at(0);
+   assert($param1->instanceOf(Literal),'Only literals are supported inside the slice function, you are passing a ' + $param1->type()->toString());
+   let fromRow = $param1->cast(@Literal);
+   let param2 = processValueSpecification($f.parametersValues->at(2), $currentPropertyMapping, $operation, $vars, $state, JoinType.LEFT_OUTER, $nodeId, $aggFromMap, $context, $extensions)->toOne()->cast(@SelectWithCursor).select.columns->at(0);
+   assert($param2->instanceOf(Literal),'Only literals are supported inside the slice function, you are passing a ' + $param2->type()->toString());
+   let toRow = $param2->cast(@Literal); 
    let processedSelect = ^$mainSelect (
                   fromRow = if ($fromRow.value->instanceOf(Integer) && $fromRow.value->cast(@Integer) == 0, |[], |$fromRow),
                   toRow = $toRow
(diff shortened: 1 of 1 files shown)


legend-engine-xt-relationalStore-pure/src/main/resources/core_relational/relational/pureToSQLQuery/pureToSQLQuery.pure
```diff
@@ -2383,9 +2383,12 @@ function meta::relational::functions::pureToSqlQuery::processSlice(f:FunctionExp
...
-   let toRow = processValueSpecification($f.parametersValues->at(2), $currentPropertyMapping, $operation, $vars, $state, JoinType.LEFT_OUTER, $nodeId, $aggFromMap, $context, $extensions)->toOne()->cast(@SelectWithCursor).select.columns->at(0)->cast(@Literal);
-
+   let param1 = processValueSpecification($f.parametersValues->at(1), $currentPropertyMapping, $operation, $vars, $state, JoinType.LEFT_OUTER, $nodeId, $aggFromMap, $context, $extensions)->toOne()->cast(@SelectWithCursor).select.columns->at(0);
+   assert($param1->instanceOf(Literal),'Only literals are supported inside the slice function, you are passing a ' + $param1->type()->toString());
+   let fromRow = $param1->cast(@Literal);
+   let param2 = processValueSpecification($f.parametersValues->at(2), $currentPropertyMapping, $operation, $vars, $state, JoinType.LEFT_OUTER, $nodeId, $aggFromMap, $context, $extensions)->toOne()->cast(@SelectWithCursor).select.columns->at(0);
+   assert($param2->instanceOf(Literal),'Only literals are supported inside the slice function, you are passing a ' + $param2->type()->toString());
```
epsstan, line 2390: Can we change the message to be more succinct.

Something like 'Invalid type for parameter 0. Expected=Lietral, Found=X'

legend-engine-xt-relationalStore-pure/src/main/resources/core_relational/relational/pureToSQLQuery/pureToSQLQuery.pure
```diff
@@ -2721,8 +2725,9 @@ function meta::relational::functions::pureToSqlQuery::processDrop(f:FunctionExpr
    let mainQuery = processValueSpecification($f.parametersValues->at(0), $currentPropertyMapping, $operation, $vars, $state, $joinType, $nodeId, $aggFromMap, $context, $extensions)->toOne()->cast(@SelectWithCursor);
    let mainSelect = $mainQuery.select;
 
-   let fromValue = processValueSpecification($f.parametersValues->at(1), $currentPropertyMapping, $operation, $vars, $state, JoinType.LEFT_OUTER, $nodeId, $aggFromMap, $context, $extensions)->toOne()->cast(@SelectWithCursor).select.columns->at(0)->cast(@Literal);//.value->cast(@Integer);
-
+   let param1 = processValueSpecification($f.parametersValues->at(1), $currentPropertyMapping, $operation, $vars, $state, JoinType.LEFT_OUTER, $nodeId, $aggFromMap, $context, $extensions)->toOne()->cast(@SelectWithCursor).select.columns->at(0);//.value->cast(@Integer);
+   assert($param1->instanceOf(Literal),'Only literals are supported inside the drop function, you are passing a ' + $param1->type()->toString());
+   let fromValue = $param1->cast(@Literal);
```
epsstan, line 2730: No tests ?

epsstan (changes requested):

legend-engine-xt-relationalStore-pure/src/main/resources/core_relational/relational/pureToSQLQuery/pureToSQLQuery.pure
abhishoya-gs: changed

legend-engine-xt-relationalStore-pure/src/main/resources/core_relational/relational/pureToSQLQuery/pureToSQLQuery.pure
abhishoya-gs: I am unaware of any way we could write failure tests in pure i.e. asserting on exception

rafaelbey:
I don think the user understand what a 'Literal' is.  We should express the messages from the point of view on how the users are defining their queries allowing them to hopefully fix it.  Like in this case I suspect we expect a 'constant' value, and they are passing a variable.

abhishoya-gs:
> I don think the user understand what a 'Literal' is since this is a pure2sql interpretation of the VS.
> 
> We should express the messages from the point of view on how the users are defining their queries allowing them to hopefully fix it. Like in this case I suspect we expect a 'constant' value, and they are passing a variable.

@rafaelbey I see, this makes sense. We are expecting a value (constant or variable) and they are making use of an operation/function such as `->limit(1+2)`/`->limit(someFunction())`

How about if we keep it `Invalid type for parameter inside the take/limit function. Expected a value, found an operation/function`?

rafaelbey (approved):

legend-engine-xt-relationalStore-pure/src/main/resources/core_relational/relational/pureToSQLQuery/pureToSQLQuery.pure
abhishoya-gs: Since this is a small non-functional change I feel existing tests should be enough to confirm it as a non-breaking change, though if we still need tests I will write those in engine. Thoughts?

Creating assertFail sounds reasonable to me.

epsstan (approved):

PR #1476, opened 2023-03-03
make legacy test optional, add warning for missing mapping in runtime

janeenyamak1:
…runtime

#### What type of PR is this?
This PR will make legacy tests in a service optional as well as adding a warning if there is no mapping in the runtime.

These changes will allow users to migrate their legacy test to testSuite

Fixes #


legend-engine-pure-code-compiled-core/src/main/resources/core/legend/service/metamodel.pure
```diff
@@ -25,9 +25,9 @@ Class meta::legend::service::metamodel::Service extends PackageableElement, meta
    multiExecutionKeyMustBePartOfPattern: if($this.execution->instanceOf(meta::legend::service::metamodel::PureMultiExecution) && $this.execution->cast(@meta::legend::service::metamodel::PureMultiExecution).executionParameters->isNotEmpty() ,|$this.pattern->contains('{'+ $this.execution->cast(@PureMultiExecution).executionKey+'}'),|true),
 
    executionAndTestTypesMatch: if($this.execution->instanceOf(meta::legend::service::metamodel::PureMultiExecution),
-                                  | assert($this.test->instanceOf(MultiExecutionTest),'Service with multiple executions requires a MultiExecution test'),
+                                  |if($this.test->isNotEmpty(),| assert($this.test->toOne()->instanceOf(MultiExecutionTest),'Service with multiple executions requires a MultiExecution test'),| true),
```
rafaelbey, line 28: could we do an && rather than an if on the assert?

`assert($this.test->isNotEmpty() && ...)`

or you could just do map to execute whenever it exists:

`this.test->map(a | assert(...))`

janeenyamak1: ended up going with assert($this.test->isNotEmpty() && ...)

Yasirmod17 (dismissed):

PR #1482, opened 2023-03-06
GraphQL caching fix

epsstan:
#### What type of PR is this?
Improvement

Fixes caching of GraphQL execution plans, by including nested selection sets. 
Previously nested selection sets where not being included which resulted in the two queries with the same top level selection set, but different nested selection sets, using the same plan.

Fixes #

No.


epsstan:
@abhishoya-gs @AFine-gs  - Please take a look.

legend-engine-xt-graphQL-query/src/main/java/org/finos/legend/engine/query/graphQL/api/execute/model/GraphQLCachableVisitorHelper.java
```diff
@@ -71,6 +71,7 @@ private static Selection selectionVisitor(Selection selection)
             public Selection visit(Field val)
             {
                 Field field = new Field();
+                field.name = val.name;
```
epsstan, line 74: I'm including the name to make the JSON serialized key a tad more readable

legend-engine-xt-graphQL-query/src/test/java/org/finos/legend/engine/query/graphQL/api/test/TestGraphQLAPI.java
```diff
@@ -405,6 +406,59 @@ public void testCacheUsed() throws Exception
 
     }
 
+    @Test
+    public void testCachingUsingNestedSelectionSets() throws Exception
```
abhishoya-gs, line 410: This tests the cache is missed, so name should be testCacheMissOnNestedSelectionSets or something like this.

Also we should have a test for cache hit using nested selection sets.

rafaelbey (approved):

PR #1483, opened 2023-03-06
Rename sample JUnit test reports

epsstan:
#### What type of PR is this?
Bug Fix

Rename sample JUnit test reports that were previously named TEST-<full class name>.xml.

When legend-engine is being run under TeamCity,  TeamCity's surefire report watcher was picking up these files ( **/TEST-*.xml) and was reporting the results from the sample reports as real test failures. 

Fixes #

No.


rafaelbey (approved):
I think you meant 'sample' rather than 'sanple'

PR #1149, opened 2022-11-14
Fix a bug with diagram property view referencing association

gayathrir11:
#### What type of PR is this?
 Bug fix

#### What does this PR do / why is it needed ?

Related to https://github.com/finos/legend-studio/issues/1593
#### Which issue(s) this PR fixes:

Fixes #

PR #1498, opened 2023-03-09
fix property mapping not properly resolving from associations property

akphi:
#### What type of PR is this?

Bug Fix

Fixes # https://github.com/finos/legend-studio/issues/1991
This is very similar to https://github.com/finos/legend-engine/pull/1149
Sometimes, clients could send protocol models with `PropertyPointer.class` pointing at an association instead of a class, for this reason, we need to properly resolve the owner of the property when building the graph


rafaelbey:
Is there a way we can add a test case for this?

akphi:
> Is there a way we can add a test case for this?

of course! will do

MauricioUyaguari (approved):

PR #1500, opened 2023-03-09
Preval - simplify filter with constant boolean return

gs-jp1:
#### What type of PR is this?
Improvement

Simplifies $x->filter(f | false) to []
Simplifies $x->filter(f | true) to $x

No

### DIFF
--- a/legend-engine-pure-code-compiled-core/src/main/resources/core/pure/router/preeval/preeval.pure
+++ b/legend-engine-pure-code-compiled-core/src/main/resources/core/pure/router/preeval/preeval.pure
@@ -675,7 +675,40 @@ function <<access.private>> meta::pure::router::preeval::prevalGenericSimpleFunc
                                   openVars = [],
                                   modified = true
                                 );
-                            })
+                            }),
+                            pair(
+                              {theSfe:SimpleFunctionExpression[1] | 
+                                //TODO getAll check is only needed as router does not route FunctionDefinition wth constant result.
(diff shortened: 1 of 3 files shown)


legend-engine-pure-code-compiled-core/src/main/resources/core/pure/router/preeval/preeval.pure
```diff
@@ -714,6 +747,29 @@ function <<access.private>> meta::pure::router::preeval::prevalGenericSimpleFunc
...
 }
 
+function meta::pure::router::preeval::isGetAll(v:ValueSpecification[0..1]):Boolean[1]
+{
+  $v->match([
+    s:SimpleFunctionExpression[1] | $s.func->in([
+      getAll_Class_1__T_MANY_,getAllVersions_Class_1__T_MANY_,getAll_Class_1__Date_1__T_MANY_,
+      getAll_Class_1__Date_1__Date_1__T_MANY_,getAllVersionsInRange_Class_1__Date_1__Date_1__T_MANY_,
+      getAllForEachDate_Class_1__Date_MANY__T_MANY_
+    ]) || $s.parametersValues->first()->isGetAll(),
```
rafaelbey, line 757: should we put this list on a common place?  At a glance seems incomplete compared with what the router have:

https://github.com/finos/legend-engine/blob/master/legend-engine-pure-code-compiled-core/src/main/resources/core/pure/router/routing/router_routing.pure#L317

Also, there are few places we been changing from ->in() to do || given it performs better on paths that are on the critical path.

aormerod-gs (approved):

PR #1523, opened 2023-03-15
Persistence Component: Schema Evolution Enhancements and Bug Fixes

prasar-ashutosh:
#### What type of PR is this?

Choose one of the following labels :
- Improvement
- Bug Fix

- Schema evolution support for snowflake by adding its capability set and implicit/explicit data type mapping.
- Add the feature for users to provide a custom capability set for schema evolution and returning a list of performed evolution SQLs for user's reference.
- Add the feature to derive Main dataset from Staging Dataset
- Add the feature to derive Main dataset from Database using jdbc connection 
- Added unit tests and end to end tests for different scenario's.
- Case conversion Support for Datasets and Ingest Mode
- Fixed the bugs in executor mode

Fixes #

NO


rafaelbey (approved):
Is a big change - the only things obvious I could notice was the fact that optimizer ends casting the result always.  Wondering if the classes can be generalized?

PR #1528, opened 2023-03-15
add supportsBinding to xt description

MauricioUyaguari:

Fixes #


legend-engine-external-shared-format-model/src/main/java/org/finos/legend/engine/external/shared/format/model/api/ExternalFormatDescription.java
```diff
@@ -48,6 +50,7 @@ public static ExternalFormatDescription newDescription(ExternalFormatExtension<?
         result.modelGenerationProperties = result.supportsModelGeneration ? toGenerationProperties(((ExternalFormatModelGenerationExtension<?, ?>) extension).getModelGenerationProperties(pureModel)) : Collections.emptyList();
         result.supportsSchemaGeneration = extension instanceof ExternalFormatSchemaGenerationExtension;
         result.schemaGenerationProperties = result.supportsSchemaGeneration ? toGenerationProperties(((ExternalFormatSchemaGenerationExtension<?, ?>) extension).getSchemaGenerationProperties(pureModel)) : Collections.emptyList();
+        result.supportsBinding = extension.getExternalFormatContract()._externalFormatBindingValidator() != null;
```
hardikmaheshwari, line 53: Can we rather have default method in `ExternalFormatExtension` interface for this?
I would prefer us not doing analytics on contract outside of interface and rather use this default function. Does it make sense?

hardikmaheshwari (changes requested):

hardikmaheshwari (approved):
Thanks for helping add this

akphi (approved):

PR #1533, opened 2023-03-16
Identity propagation in graphql

abhishoya-gs:
#### What type of PR is this?
Improvement

Currently we don't propagate identity to plan executor in graphql, this PR improves that flow

Fixes #

yes, users will be able to use identity based auth flows with graphql


legend-engine-xt-graphQL-query/src/main/java/org/finos/legend/engine/query/graphQL/api/execute/GraphQLExecute.java
```diff
@@ -296,11 +298,11 @@ private Response executeGraphQLQuery(String queryClassPath, String mappingPath,
                             JsonStreamingResult result = null;
                             try
                             {
-                                Map<String, Object> parameterMap = new HashMap<>();
-                                extractFieldByName(graphQLQuery, p.propertyName).arguments.stream().forEach(a -> parameterMap.put(a.name, argumentValueToObject(a.value)));
+                                Map<String, Result> parameterMap = new HashMap<>();
```
epsstan, line 301: Is this change related to the identity propagation ?

epsstan (changes requested):

epsstan (approved):

PR #1535, opened 2023-03-16
Persist createAt, lastUpateAt timestamp info for a query

gayathrir11:
 This PR adds a new field `lastUpdateTimeStamp` info for a query 

Fixes #

### DIFF
--- a/legend-engine-application-query/src/main/java/org/finos/legend/engine/application/query/api/QueryStoreManager.java
+++ b/legend-engine-application-query/src/main/java/org/finos/legend/engine/application/query/api/QueryStoreManager.java
@@ -28,6 +28,7 @@ import org.eclipse.collections.impl.utility.LazyIterate;
 import org.eclipse.collections.impl.utility.ListIterate;
 import org.finos.legend.engine.application.query.model.Query;
 import org.finos.legend.engine.application.query.model.QueryEvent;
+import org.finos.legend.engine.application.query.model.QueryProjectCoordinates;
 import org.finos.legend.engine.application.query.model.QuerySearchSpecification;
 import org.finos.legend.engine.application.query.model.QueryStoreStats;
 import org.finos.legend.engine.protocol.pure.v1.model.packageableElement.domain.StereotypePtr;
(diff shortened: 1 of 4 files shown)


legend-engine-application-query/src/main/java/org/finos/legend/engine/application/query/model/Query.java
```diff
@@ -32,6 +32,7 @@
     public String content;
     // We make it clear that we only allow a single owner
     public String owner;
+    public Long lastUpdateTimeStamp;
```
akphi, line 35: Let's do `lastUpdateAt`, also let's add `createdAt` timestamp

legend-engine-application-query/src/main/java/org/finos/legend/engine/application/query/api/QueryStoreManager.java
```diff
@@ -233,7 +234,7 @@ public List<Query> getQueries(
         return LazyIterate.collect(this.getQueryCollection()
                 .find(filters.isEmpty() ? EMPTY_FILTER : Filters.and(filters))
                 // NOTE: return a light version of the query to save bandwidth
-                .projection(Projections.include("id", "name", "versionId", "groupId", "artifactId", "owner"))
+                .projection(Projections.include("id", "name", "versionId", "groupId", "artifactId", "owner", "lastUpdateTimeStamp"))
```
akphi, line 237: let's also add `createdAt` to this

legend-engine-application-query/src/main/java/org/finos/legend/engine/application/query/model/Query.java
akphi: Similar to SDLC, let's expose these fields to the search API
image

Another thing you need to add is what can you modify the search API so that you can fetch `lastest 10 queries`, `latest 20 queries`, base on `lastUpdatedAt` - maybe introduce a boolean in the payload?

akphi (changes requested):

akphi (approved):

akphi (approved):

PR #1553, opened 2023-03-20
Add API to get schema for SQL

gs-manvig:
#### What type of PR is this?
Improvement

New feature to get metadata for SQL string

Fixes #

Screenshot 2023-03-22 at 3 41 37 PM


legend-engine-xt-sql-query/src/main/java/org/finos/legend/engine/query/sql/model/SchemaColumn.java
```diff
@@ -0,0 +1,66 @@
...
+package org.finos.legend.engine.query.sql.model;
+
+import com.fasterxml.jackson.annotation.JsonProperty;
+import org.apache.commons.lang3.builder.EqualsBuilder;
+import org.apache.commons.lang3.builder.HashCodeBuilder;
+
+public class SchemaColumn
+{
+    private final String name;
+    private final String type;
```
aormerod-gs, line 25: type as a string...Is there an enum domain we can use?

gs-manvig: The type is derived from TDS Column, which has the pure `DataType` which can be primitive (only primitive for now), or other pure classes (in the future). The pure `DataType` is converted to string using the pure function `meta::pure::functions::meta::elementToPath`

There is a datatype enum defined in persistence [model](https://github.com/finos/legend-engine/blob/master/legend-engine-xt-persistence-component/legend-engine-xt-persistence-component-logical-plan/src/main/java/org/finos/legend/engine/persistence/components/logicalplan/datasets/DataType.java), and an abstract class in [relational store](https://github.com/finos/legend-engine/blob/master/legend-engine-xt-relationalStore-protocol/src/main/java/org/finos/legend/engine/protocol/pure/v1/model/packageableElement/store/relational/model/datatype/DataType.java). But there is no straightforward way to convert a pure `DataType` to either of those

legend-engine-xt-sql-query/src/main/java/org/finos/legend/engine/query/sql/model/SchemaColumn.java
```diff
@@ -0,0 +1,66 @@
...
+        return name;
+    }
+
+    public String getType()
+    {
+        return type;
+    }
+
+    @Override
+    public boolean equals(Object o)
```
rafaelbey, line 44: I will suggest to avoid using commons equals/has, and let IntelliJ generate these methods with only JDK dependencies.  

The issue with these tools is that if you end calling equals and hashcode a lot, these will perform worst that if using just JDK.  

For example, each call needs to allocate a new object to do their function, which adds up when executing these methods often.

rafaelbey (changes requested):

legend-engine-xt-sql-query/src/main/java/org/finos/legend/engine/query/sql/model/SchemaColumn.java
gs-manvig: That makes sense. Updated in the latest revision.

rafaelbey (approved):

gs-manvig:
This PR adds metadata support only for primitive types.
Support for Enums will be added in a future PR

aormerod-gs (approved):

PR #1563, opened 2023-03-23
Add trino dependencies to server module

jinanisha:
#### What type of PR is this?

Improvement

Add trino dependencies to server module

Fixes #

No


vikask-gs:
we will also need to add the deps in execution-collection and generation-collection modules

vikask-gs (approved):

PR #1567, opened 2023-03-23
Enum schema

gs-manvig:
#### What type of PR is this?
Improvement

Add support for pure Enums in the schema APIs introduced in #1553 

Fixes #


aormerod-gs (dismissed):

aormerod-gs (commented):
Not specific to these changes, but should we be aligning these schema types to meta::relational::metamodel::datatype::* ?

aormerod-gs (approved):

aormerod-gs:
> Not specific to these changes, but should we be aligning these schema types to meta::relational::metamodel::datatype::* ?

Or perhaps we should be reflecting the types naming/concepts from Postgres https://www.postgresql.org/docs/current/datatype.html

PR #1577, opened 2023-03-27
Support complex property mapping without class mapping

hardikmaheshwari:
#### What type of PR is this?
Improvement

This PR adds support for mapping complex property to a transform returning complex property directly without a need for a explicit mapping. 

Ex:
```
Mapping test::Mapping
(
    Person : Pure
    {
        ~src SPerson
        firstName: $src.firstName
        firm: $src.firm      // Just like a String can be matched to a string, Firm can be matched to a Firm
    }
)
```

Fixes #


rafaelbey:
@cla-bot check

rafaelbey (approved):
Looks great!  One thought - wonder if we can find a path to move the java test like `TestComplexPropertyMappingWithoutExplicitClassMapping` to pure using legend compile + execute.

hardikmaheshwari:
> Looks great! One thought - wonder if we can find a path to move the java test like `TestComplexPropertyMappingWithoutExplicitClassMapping` to pure using legend compile + execute.

We can though have replicated it in java on purpose to test engine compilation flow which is not there when we use execute directly. tried legendCompile but it seems there is a breakage with that

vikask-gs (approved):

PR #1584, opened 2023-03-28
Add Legend Service OpenAPI Specification generation extension

### LINKED ISSUE #1583: Support OpenAPI Specification Generation for Legend Services
### Description of Problem:
Legend services do not expose an openapi specification. OpenAPI specification is a standard, programming language-agnostic interface description for HTTP APIs, which allows both humans and computers to discover and understand the capabilities of a service without requiring access to source code, additional documentation, or inspection of network traffic.

It is a standard contract which most clients would expect to see and having this feature available on default for legend service will make it easy to discover, consume and understand the legend service.

### Potential Solutions:
Solution 1:
Write a pure function that accepts a legend service and returns an openapi specification string. This pure function can be packaged into an API/plugin and invoked in build/service layers to generate openapi specification.

Solution 2:
Get a pure function definition in either pure/json format and use service translations to convert into a Java object. Derive URL patterns, return types and all classes involved. Build the OpenAPI object using swagger-parser library.

kranthikiran01:
#### What type of PR is this?
Improvement

Provides ability to generate openapi specification for a given legend service. OpenAPIArtifactGenerationExtension can be used with legend sdlc maven file plugin to generate openapi spec as a build artifact

Fixes #1583 

Support includes GET and POST methods. Metamodel supports other operations, but implementation to derive methods from service is required. More iterations required to make OpenAPI specification fully implemented i.e security extensions etc.
No


legend-engine-pure-code-compiled-core/src/main/resources/core/external/format/openapi/tests.pure
```diff
@@ -0,0 +1,67 @@
...
...dexTDS":{"type":"object","properties":{"index":{"$ref":"#/components/schemas/Elasticsearch7StoreIndexMapping"}}},"TokenCountProperty":{"type":"object","properties":{"index":{"type":"boolean"},"enable_position_increments":{"type":"boolean"},"null_value":{"type":"float"},"analyzer":{"type":"string"},"boost":{"type":"float"},"type":{"type":"string"},"_pure_protocol_type":{"type":"string"}}},"AggregateMetricDoubleProperty":{"type":"object","properties":{"default_metric":{"type":"string"},"metrics":{"type":"string"},"type":{"type":"string"},"_pure_protocol_type":{"type":"string"}}},"BinaryProperty":{"type":"object","properties":{"type":{"type":"string"},"_pure_protocol_type":{"type":"string"}}},"Package":{"type":"object","properties":{"children":{"type":"array","items":[{"$ref":"#/components/schemas/PackageableElement"}]}}},"RankFeaturesProperty":{"type":"object","properties":{"type":{"type":"string"},"_pure_protocol_type":{"type":"string"}}},"LatLonGeoLocation":{"type":"object","properties":{"lon":{"type":"float"},"lat":{"type":"float"}}},"DateRangeProperty":{"type":"object","properties":{"format":{"type":"string"},"type":{"type":"string"},"_pure_protocol_type":{"type":"string"}}}}}}';
```
rafaelbey, line 50: Can we put this on a file on the pure repo, and then use readFile function to get it?

Then, we can use assertJsonEquals function.

This would help with readability.

legend-engine-xt-openapi-generation/src/main/java/org/finos/legend/engine/generation/OpenApiArtifactGenerationExtension.java
```diff
@@ -0,0 +1,56 @@
...
+    {
+        return element instanceof Root_meta_legend_service_metamodel_Service;
+    }
+
+    @Override
+    public List<Artifact> generate(PackageableElement element, PureModel pureModel, PureModelContextData data, String clientVersion)
+    {
+        assertTrue(this.canGenerate(element), () -> "OpenApi extension only supports service elements");
+        String result = core_external_format_openapi_generation.Root_meta_external_format_openapi_generation_serviceToOpenApi_Service_1__Server_1__String_1_((Root_meta_legend_service_metamodel_Service) element, new Root_meta_external_format_openapi_metamodel_Server_Impl("")._url("localhost"), pureModel.getExecutionSupport());
+        Artifact output = new Artifact(result, element.getName() + "_openapi.json", "json");
```
rafaelbey, line 53: One thing worth noting is that the generator is already qualifying the generated file by the path and the "key"

For example, `abc::abc::ABCService` will get this file under:  `abc/abc/ABCService/openapi_spec/{your_filename_here}`

Hence, you can make this name more vanilla to make thing easier for the system that will like to use these.  For example, you can just set the name to `spec.json`.

legend-engine-xt-openapi-generation/src/main/java/org/finos/legend/engine/generation/OpenApiArtifactGenerationExtension.java
```diff
@@ -0,0 +1,56 @@
...
+    @Override
+    public boolean canGenerate(PackageableElement element)
+    {
+        return element instanceof Root_meta_legend_service_metamodel_Service;
+    }
+
+    @Override
+    public List<Artifact> generate(PackageableElement element, PureModel pureModel, PureModelContextData data, String clientVersion)
+    {
+        assertTrue(this.canGenerate(element), () -> "OpenApi extension only supports service elements");
```
rafaelbey, line 51: The platform is already calling `canGenerate` before calling generate, hence this check is not necessary,

rafaelbey (changes requested):
One things that is not necessary relevant to the actual change - do we know what is the impact to the Studio GitLab projects?  This means every service on every project will now generate these files that before was not the case.

legend-engine-pure-code-compiled-core/src/main/resources/core/external/format/openapi/tests.pure
kranthikiran01: @rafaelbey Done. Can you please review and approve this?

legend-engine-xt-openapi-generation/src/main/java/org/finos/legend/engine/generation/OpenApiArtifactGenerationExtension.java
kranthikiran01: Makes sense, will update the file name suffix to something simpler like spec.json

legend-engine-xt-openapi-generation/src/main/java/org/finos/legend/engine/generation/OpenApiArtifactGenerationExtension.java
kranthikiran01: Thanks for letting me know, I was not aware platform calls it already.

legend-engine-xt-openapi-generation/src/main/java/org/finos/legend/engine/generation/OpenApiArtifactGenerationExtension.java
kranthikiran01: Also, can we get the url/host of legend service somehow? Hardcoding localhost here doesn't seem right. Can we read any configuration values for host?

rafaelbey: Not really.  These files are generated during the Studio project pipelines on GitLab.  There is no server.  Can we use some sort of template, like ${HOST} or something that reflects that is a value that needs to change.  How to change it and who will need to change it is a different discussion.

rafaelbey:
One last concern - once we merge this change, it will start generating for every existing service.  Either we need to make sure the open api generation is comprehensive and covers all use case, or we need a try/catch that wont fail the Studio project pipelines.

legend-engine-xt-openapi-generation/src/main/java/org/finos/legend/engine/generation/OpenApiArtifactGenerationExtension.java
kranthikiran01: Thanks for the suggestion. Applied the changes along with logs for failure

hardikmaheshwari:
Should we leverage external formats for this?
We support open API standard schemagen for classes which are triggered via file gen.
Can we not trigger similar schema gen for services (like classes) which can glued in with other file generation mechanics

legend-engine-pure-code-compiled-core/src/main/resources/core/external/format/openapi/testOpenApiSpec.txt
AFine-gs, line 1: It  looks like this has scoped in all of the TabularDataset metamodel, which is likely not what we want here

kranthikiran01: @AFine-gs  My latest commit generates models only in scope. Please review

MauricioUyaguari:
> One things that is not necessary relevant to the actual change - do we know what is the impact to the Studio GitLab projects? This means every service on every project will now generate these files that before was not the case.

This was addressed by adding a stereotype check on the service that would trigger the generation. This will only be generated if the user adds the stereotype and not on every service

MauricioUyaguari (approved):

PR #1585, opened 2023-03-28
Fix prepared statement params

shtirlets:
#### What type of PR is this?
Bug Fix

Resolved a bug that prevented execution of prepared statements with parameters. 

Fixes #


gs-manvig:
LGTM

rafaelbey (changes requested):
While the comment on test container is one to think, if we end with H2, we need to set the port to zero to avoid port collisions.

finos-admin:

finos-admin:

PR #1586, opened 2023-03-28
Switch to static filename for SearchDocumentArtifactGenerationExtension

tomwilson-gs:
#### What type of PR is this?

Bug Fix

Switch the filename for artifacts generated by SearchDocumentArtifactGenerationExtension to be static rather than using the element's path to avoid scenarios where the filename is too large for the filesystem.

N/A

### DIFF
--- a/legend-engine-xt-analytics-search-generation/src/main/java/org/finos/legend/engine/generation/SearchDocumentArtifactGenerationExtension.java
+++ b/legend-engine-xt-analytics-search-generation/src/main/java/org/finos/legend/engine/generation/SearchDocumentArtifactGenerationExtension.java
@@ -16,8 +16,6 @@
 
 package org.finos.legend.engine.generation;
 
-import static org.finos.legend.engine.language.pure.compiler.toPureGraph.HelperModelBuilder.getElementFullPath;
-
 import java.util.Collections;
 import java.util.List;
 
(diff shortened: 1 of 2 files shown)


legend-engine-xt-analytics-search-generation/src/main/java/org/finos/legend/engine/generation/SearchDocumentArtifactGenerationExtension.java
```diff
@@ -56,8 +56,7 @@ public boolean canGenerate(PackageableElement element)
     @Override
     public List<Artifact> generate(PackageableElement element, PureModel pureModel, PureModelContextData data, String clientVersion)
     {
-        String path = getElementFullPath(element, pureModel.getExecutionSupport());
-        String fileName = String.format("%s.json", path);
+        String fileName = "SearchDocumentResult.json";
```
aormerod-gs, line 59: Should we define this as a constant? (That can then be referenced in the tests)

aormerod-gs (dismissed):

legend-engine-xt-analytics-search-generation/src/main/java/org/finos/legend/engine/generation/SearchDocumentArtifactGenerationExtension.java
tomwilson-gs: Updated

aormerod-gs (approved):

PR #1590, opened 2023-03-28
support extensible entitlements analytics for dataspace

YannanGao-gs:

support extensible entitlements analytics for data-space

Fixes #

### DIFF
--- a/legend-engine-xt-analytics-store-entitlement/src/main/java/org/finos/legend/engine/entitlement/model/specification/DatasetSpecification.java
+++ b/legend-engine-xt-analytics-store-entitlement/src/main/java/org/finos/legend/engine/entitlement/model/specification/DatasetSpecification.java
@@ -1,0 +1,46 @@
+//  Copyright 2023 Goldman Sachs
+//
+//  Licensed under the Apache License, Version 2.0 (the "License");
+//  you may not use this file except in compliance with the License.
+//  You may obtain a copy of the License at
+//
+//       http://www.apache.org/licenses/LICENSE-2.0
+//
+//  Unless required by applicable law or agreed to in writing, software
+//  distributed under the License is distributed on an "AS IS" BASIS,
+//  WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+//  See the License for the specific language governing permissions and
+//  limitations under the License.
+
+package org.finos.legend.engine.entitlement.model.specification;
+
+import com.fasterxml.jackson.annotation.JsonSubTypes;
+import com.fasterxml.jackson.annotation.JsonTypeInfo;
+import org.finos.legend.engine.entitlement.model.specification.relationalDatabase.RelationalDatabaseTableSpecification;
+
+@JsonTypeInfo(use = JsonTypeInfo.Id.NAME, property = "_type")
+@JsonSubTypes({
+        @JsonSubTypes.Type(value = RelationalDatabaseTableSpecification.class, name = "relationalDatabaseTable")
+})
+
+public class DatasetSpecification
+{
+    private String name;
+    private String type;
+
+    public DatasetSpecification(String name, String type)
+    {
+        this.name = name;
+        this.type = type;
+    }
+
+    public String getType()
+    {
+        return type;
+    }
+
+    public String getName()
+    {
+        return name;
+    }
+}
--- a/pom.xml
+++ b/pom.xml
@@ -258,6 +258,8 @@
         <module>legend-engine-xt-nonrelationalStore-mongodb-grammar</module>
         <module>legend-engine-xt-nonrelationalStore-mongodb-protocol</module>
         <module>legend-engine-xt-nonrelationalStore-mongodb-grammar-integration</module>
+        <module>legend-engine-xt-analytics-store-entitlement</module>
+        <module>legend-engine-xt-analytics-store-entitlement-api</module>
 
         <!-- NOTE: Do not change the ordering of this block, we want to build these modules last -->
         <module>legend-engine-test-reports</module>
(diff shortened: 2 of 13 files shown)


MauricioUyaguari (approved):

legend-engine-entitlement/src/main/java/org/finos/legend/engine/entitlement/services/CheckEntitlements.java
```diff
@@ -0,0 +1,93 @@
...
+    @Consumes({MediaType.APPLICATION_JSON})
+    public Response generateDatasetSpecifications(CheckEntitlementInput input, @ApiParam(hidden = true) @Pac4JProfileManager ProfileManager<CommonProfile> pm)
+    {
+        MutableList<CommonProfile> profiles = ProfileManagerHelper.extractProfiles(pm);
+        List<EntitlementServiceExtension> extensions = EntitlementServiceExtensionLoader.extensions();
+        try (Scope scope = GlobalTracer.get().buildSpan("generate datasetSpecifications").startActive(true))
+        {
+            try
+            {
+                return ManageConstantResult.manageResult(profiles, LazyIterate.flatCollect(extensions, extension -> extension.generateDatasetSpecifications(input, modelManager, profiles)).toList());
```
akphi, line 64: I think this logic is good, but over-simplified, could we end up with duplicates? like 2 extension both know how to handle a runtime connection?

pom.xml
```diff
@@ -258,6 +258,7 @@
         <module>legend-engine-xt-nonrelationalStore-mongodb-grammar</module>
         <module>legend-engine-xt-nonrelationalStore-mongodb-protocol</module>
         <module>legend-engine-xt-nonrelationalStore-mongodb-grammar-integration</module>
+        <module>legend-engine-entitlement</module>
```
akphi, line 261: as discussed, we would like to split this module into 2 part: core logic and API logic, since the core logic can be depended on by other stores, we don't want to mingle in the API depeendencies (like dropwizard for example)

akphi: Also, we don't we expose the API in legend-engine-server?

legend-engine-entitlement/src/main/java/org/finos/legend/engine/entitlement/model/specification/relationalDatabase/RelationalDatabaseTableSpecification.java
```diff
@@ -0,0 +1,47 @@
...
+//  distributed under the License is distributed on an "AS IS" BASIS,
+//  WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+//  See the License for the specific language governing permissions and
+//  limitations under the License.
+
+package org.finos.legend.engine.entitlement.model.specification.relationalDatabase;
+
+import org.finos.legend.engine.entitlement.model.specification.DatasetSpecification;
+
+public class RelationalDatabaseTableSpecification extends DatasetSpecification
```
akphi, line 19: for now, it's being here is fine, we can totally offset this to the store itself though

akphi (changes requested):
Minor comments

legend-engine-entitlement/src/main/java/org/finos/legend/engine/entitlement/services/CheckEntitlements.java
YannanGao-gs: EntitlementServiceExtension is on store level. 
For relationalDatabase, it will have to create a class that implements EntitlementServiceExtension. 
For ServiceStore, it will have to create another. 
For each EntitlementServiceExtension, it has two methods.

legend-engine-xt-analytics-store-entitlement/src/main/java/org/finos/legend/engine/entitlement/model/specification/DatasetSpecification.java
```diff
@@ -0,0 +1,46 @@
...
+
+package org.finos.legend.engine.entitlement.model.specification;
+
+import com.fasterxml.jackson.annotation.JsonSubTypes;
+import com.fasterxml.jackson.annotation.JsonTypeInfo;
+import org.finos.legend.engine.entitlement.model.specification.relationalDatabase.RelationalDatabaseTableSpecification;
+
+@JsonTypeInfo(use = JsonTypeInfo.Id.NAME, property = "_type")
+@JsonSubTypes({
+        @JsonSubTypes.Type(value = RelationalDatabaseTableSpecification.class, name = "RelationalDatabaseTableSpecification")
```
akphi, line 23: TODO: change this to `relationalDatabaseTable`

legend-engine-xt-analytics-store-entitlement-api/src/main/java/org/finos/legend/engine/api/analytics/EntitlementServiceExtension.java
```diff
@@ -0,0 +1,38 @@
...
+import org.finos.legend.engine.api.analytics.model.StoreEntitlementAnalyticsInput;
+import org.finos.legend.engine.entitlement.model.entitlementReport.DatasetEntitlementReport;
+import org.finos.legend.engine.entitlement.model.specification.DatasetSpecification;
+import org.finos.legend.engine.language.pure.modelManager.ModelManager;
+import org.pac4j.core.profile.CommonProfile;
+
+import java.util.Collections;
+import java.util.List;
+
+public interface EntitlementServiceExtension
```
akphi, line 27: this looks wrong, this and the service loader should be at the lower layer, i.e. legend-engine-xt-analytics-store-entitlement

legend-engine-entitlement/src/main/java/org/finos/legend/engine/entitlement/model/specification/relationalDatabase/RelationalDatabaseTableSpecification.java
akphi: This is a TODO

legend-engine-entitlement/src/main/java/org/finos/legend/engine/entitlement/services/CheckEntitlements.java
akphi: Think you misunderstood me a bit, I meant we could end up in a situation where more than one extension implementations handle the same connection from a runtime, and their analysis might both yield the same dataset, I might be mistaken here, but let's talk about it tomorrow.

akphi (dismissed):

akphi:
> TODO: change this to `relationalDatabaseTable` 

to be consistent, we should do lowercase -> we can adjust later, no worries

akphi (approved):

PR #1605, opened 2023-03-30
Fix bug in sourcetree calculation of unions with filters

aziemchawdhary-gs:
#### What type of PR is this?
Bugfix

Fixes sourcetree calculation in certain cases involving union and mapping filters. 
When performing a union on two classes with properties of the same name, if one mapping in the union has a filter then we incorrectly duplicate properties in parts of the source tree

No

### DIFF
--- a/legend-engine-pure-code-compiled-core/src/main/resources/core/store/m2m/tests/legend/simpleObject.pure
+++ b/legend-engine-pure-code-compiled-core/src/main/resources/core/store/m2m/tests/legend/simpleObject.pure
@@ -801,6 +801,30 @@ meta::pure::mapping::modelToModel::test::alloy::simple::testMappingWithTypeName(
    assert(jsonEquivalent('{"firstName":"Pierre","lastName":"Doe","type":"_S_Person","description":"meta::pure::mapping::modelToModel::test::shared::src::_S_Person"}'->parseJSON(), $result.values->toOne()->parseJSON()));
 }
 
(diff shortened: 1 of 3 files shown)


legend-engine-pure-code-compiled-core/src/main/resources/core/store/m2m/tests/legend/simpleObject.pure
```diff
@@ -801,6 +801,27 @@ meta::pure::mapping::modelToModel::test::alloy::simple::testMappingWithTypeName(
    assert(jsonEquivalent('{"firstName":"Pierre","lastName":"Doe","type":"_S_Person","description":"meta::pure::mapping::modelToModel::test::shared::src::_S_Person"}'->parseJSON(), $result.values->toOne()->parseJSON()));
 }
 
+function <<meta::pure::profiles::test.Test, meta::pure::profiles::test.AlloyOnly>>
```
hardikmaheshwari, line 804: @aziemchawdhary-gs can we have a test with test data so that filter is executed or maybe just source tree calculation test which demonstrates that property is not added for both trees and only for one having filter?
Or do you think execution of such mapping is good enough?

aziemchawdhary-gs: good suggestion - I have added a test for source tree calculation

rafaelbey (approved):

PR #1606, opened 2023-03-30
Support formatting DateTime fields via AlloySerlizationConfig

mrudula-gs:
#### What type of PR is this?
Improvement

Users can configure formatting the DateTime properties on SerializationConfig. This is particularly important when the returned JSON is being sent to 3rd party systems where certain formatting is required.

No


legend-engine-executionPlan-execution/src/main/java/org/finos/legend/engine/plan/execution/nodes/ExecutionNodeExecutor.java
```diff
@@ -259,7 +260,17 @@ public Result visit(PureExpressionPlatformExecutionNode pureExpressionPlatformEx
 
                 AppliedFunction f = (AppliedFunction) pureExpressionPlatformExecutionNode.pure;
                 SerializationConfig config = f.parameters.size() == 3 ? (SerializationConfig) ((ClassInstance) f.parameters.get(2)).value : null;
-
+                if (config != null && config.dateTimeFormat != null)
+                {
+                    try
+                    {
+                        DateTimeFormatter.ofPattern(config.dateTimeFormat.replaceAll("\"","'"));
```
hardikmaheshwari, line 267: This seems like a check to be done at registration or if possible compilation time, no?

legend-engine-executionPlan-execution/src/main/java/org/finos/legend/engine/plan/execution/nodes/helpers/ExecutionNodeSerializerHelper.java
```diff
@@ -157,6 +158,7 @@ private static class Writer implements ISerializationWriter
         {
             this.generator = generator;
             this.includeEnumType = config != null && config.includeEnumType;
+            this.dateTimeFormat = config != null ? config.dateTimeFormat : null;
```
hardikmaheshwari, line 161: we don't ahve similar requirement for strictDate?

legend-engine-language-pure-dsl-service-execution/src/test/java/org/finos/legend/engine/language/pure/dsl/service/execution/TestServiceRunner.java
```diff
@@ -376,6 +376,38 @@ public void SimpleM2MServiceRunnerForRecursiveExecution()
...
+
+        @Override
+        public List<ServiceVariable> getServiceVariables()
+        {
+            return Collections.emptyList();
+        }
+    }
+
+    @Test
+    public void SimpleM2MServiceRunnerForDateTimeSerialization()
```
hardikmaheshwari, line 394: can we please update the test to work with few diff formats and compare it without format. Maybe including timezone

legend-engine-language-pure-dsl-service-execution/src/test/java/org/finos/legend/engine/language/pure/dsl/service/execution/TestServiceRunner.java
```diff
@@ -376,6 +376,38 @@ public void SimpleM2MServiceRunnerForRecursiveExecution()
...
+                .withSerializationFormat(SerializationFormat.PURE);
+        String result = validDateTimeSerialization.run(serviceRunnerInput1);
+        Assert.assertEquals("{\"age\":25,\"birthDate\":\"2014-02-27T15:01:35.231+0000\"}", result);
+
+        SimpleM2MServiceRunnerForDateTimeSerialization invalidDateTimeSerialization = new SimpleM2MServiceRunnerForDateTimeSerialization("test::serializeInvalidDateTimeFormat__String_1_");
+        ServiceRunnerInput serviceRunnerInput2 = ServiceRunnerInput
+                .newInstance();
+        IllegalArgumentException e1 = Assert.assertThrows(IllegalArgumentException.class, () -> invalidDateTimeSerialization.run(serviceRunnerInput2));
+        MatcherAssert.assertThat(e1, ThrowableMessageMatcher.hasMessage(CoreMatchers.startsWith("java.lang.IllegalArgumentException: Too many pattern letters: m\n" +
+                "yyyy-MM-dd\"T\"HH:mmm:ss.SSSZ is not a valid dateTime format in SerializationConfig")));
```
hardikmaheshwari, line 408: is it possible to inform users of available formats?

hardikmaheshwari (changes requested):
Thanks Mrudula for adding this

legend-engine-executionPlan-execution/src/main/java/org/finos/legend/engine/plan/execution/nodes/helpers/ExecutionNodeSerializerHelper.java
mrudula-gs: As of now it is just the DateTime fields as Pure serialization handles this by default, we don't want to default the same on Alloy by adding the offset +0000 because that may potentially break production usecases.

legend-engine-executionPlan-execution/src/main/java/org/finos/legend/engine/plan/execution/nodes/ExecutionNodeExecutor.java
mrudula-gs: So, we did try to have this at compilation time but as of now users have to create new instances of AlloySerializationConfig on text mode (and this is not supported on form mode yet) so the pure model will have the instance as a Simple Function Expression and to validate this particular config setting, we had to assume quite few things which doesn't seem reliable, so adding here

legend-engine-language-pure-dsl-service-execution/src/test/java/org/finos/legend/engine/language/pure/dsl/service/execution/TestServiceRunner.java
mrudula-gs: Each symbol used in DateTimeFormatter patterns have different set of possible patterns and so they can be combined as per requirement. The possible list would be too long and the exception that users will encounter will point out what symbol in the pattern is problematic.

legend-engine-language-pure-dsl-service-execution/src/test/java/org/finos/legend/engine/language/pure/dsl/service/execution/TestServiceRunner.java
mrudula-gs: Added tests with few other dateTime formats and with no format

legend-engine-language-pure-dsl-service-execution/src/test/java/org/finos/legend/engine/language/pure/dsl/service/execution/TestServiceRunner.java
rafaelbey: What are these rules/possibilities?  Normal JDK?  If that is the case, wonder if we should include a link to the JDK docs (?)

legend-engine-executionPlan-execution/src/main/java/org/finos/legend/engine/plan/execution/nodes/ExecutionNodeExecutor.java
```diff
@@ -259,7 +260,17 @@ public Result visit(PureExpressionPlatformExecutionNode pureExpressionPlatformEx
...
+                    }
+                    catch (IllegalArgumentException e)
+                    {
+                        throw new IllegalArgumentException(e + "\n" + config.dateTimeFormat + " is not a valid dateTime format in SerializationConfig");
+                    }
```
rafaelbey, line 272: lets avoid concatenating exception messages (ie `e + ""...`.  The best practice is to include the original exception as the cause of the new one:  `new Exception("your msg", e)`

This will ensure at least us on support  will get proper context of the original error.

rafaelbey (changes requested):

legend-engine-executionPlan-execution/src/main/java/org/finos/legend/engine/plan/execution/nodes/ExecutionNodeExecutor.java
mrudula-gs: We will move this validation to Studio end

legend-engine-language-pure-dsl-service-execution/src/test/java/org/finos/legend/engine/language/pure/dsl/service/execution/TestServiceRunner.java
mrudula-gs: We will list out the possibilities on studio

legend-engine-executionPlan-execution/src/main/java/org/finos/legend/engine/plan/execution/nodes/ExecutionNodeExecutor.java
mrudula-gs: Makes sense, sure. Also, touchbased with Mauricio on this, we will add the validation on studio end

PR #1609, opened 2023-03-31
Minor improvements to query search api

gayathrir11:
#### What type of PR is this?

Improvement

 Modified search api so that you can
- [x] search by query id
- [x] get results sorted in a way that queries owned by current user appears first
- [x] modified search by tagged value to filter on and condition instead of or condition

Fixes #

### DIFF
--- a/legend-engine-application-query/src/main/java/org/finos/legend/engine/application/query/api/QueryStoreManager.java
+++ b/legend-engine-application-query/src/main/java/org/finos/legend/engine/application/query/api/QueryStoreManager.java
@@ -28,7 +28,6 @@ import org.eclipse.collections.impl.utility.LazyIterate;
 import org.eclipse.collections.impl.utility.ListIterate;
 import org.finos.legend.engine.application.query.model.Query;
 import org.finos.legend.engine.application.query.model.QueryEvent;
-import org.finos.legend.engine.application.query.model.QueryProjectCoordinates;
 import org.finos.legend.engine.application.query.model.QuerySearchSpecification;
 import org.finos.legend.engine.application.query.model.QueryStoreStats;
 import org.finos.legend.engine.protocol.pure.v1.model.packageableElement.domain.StereotypePtr;
@@ -39,9 +38,7 @@ import org.finos.legend.engine.shared.core.vault.Vault;
 import javax.lang.model.SourceVersion;
 import javax.ws.rs.core.Response;
 import java.time.Instant;
-import java.time.format.DateTimeFormatter;
 import java.util.ArrayList;
-import java.util.Date;
 import java.util.List;
 import java.util.regex.Pattern;
 
@@ -200,17 +197,22 @@ public class QueryStoreManager
         // TODO: we can potentially create a pattern check for version
     }
 
+    public List<Query> getQuerySearchResults(QuerySearchSpecification searchSpecification, List<Bson> filters)
+    {
+        return LazyIterate.collect(this.getQueryCollection()
+                .find(filters.isEmpty() ? EMPTY_FILTER : Filters.and(filters))
+                .sort(searchSpecification.showLatestQueriesFirst != null && searchSpecification.showLatestQueriesFirst ? descending("lastUpdatedAt") : EMPTY_FILTER)
+                // NOTE: return a light version of the query to save bandwidth
+                .projection(Projections.include("id", "name", "versionId", "groupId", "artifactId", "owner", "createdAt", "lastUpdatedAt"))
+                .limit(Math.min(MAX_NUMBER_OF_QUERIES, searchSpecification.limit == null ? Integer.MAX_VALUE : searchSpecification.limit)), QueryStoreManager::documentToQuery).toList();
+    }
+
     public List<Query> getQueries(QuerySearchSpecification searchSpecification, String currentUser)
     {
         List<Bson> filters = new ArrayList<>();
-        if (searchSpecification.showCurrentUserQueriesOnly != null && searchSpecification.showCurrentUserQueriesOnly)
-        {
-            // NOTE: every user is considered owner of the queries created by unknown user
-            filters.add(Filters.in("owner", currentUser, null));
-        }
         if (searchSpecification.searchTerm != null)
         {
-            filters.add(Filters.regex("name", Pattern.quote(searchSpecification.searchTerm), "i"));
+            filters.add(Filters.or(Filters.regex("name", Pattern.quote(searchSpecification.searchTerm), "i"), Filters.regex("id", Pattern.quote(searchSpecification.searchTerm), "i")));
         }
         if (searchSpecification.projectCoordinates != null && !searchSpecification.projectCoordinates.isEmpty())
         {
@@ -229,7 +231,7 @@ public class QueryStoreManager
         }
         if (searchSpecification.taggedValues != null && !searchSpecification.taggedValues.isEmpty())
         {
-            filters.add(Filters.or(
+            filters.add(Filters.and(
                     ListIterate.collect(searchSpecification.taggedValues, taggedValue ->
                             Filters.and(Filters.eq("taggedValues.tag.profile", taggedValue.tag.profile), Filters.eq("taggedValues.tag.value", taggedValue.tag.value), Filters.eq("taggedValues.value", taggedValue.value)))));
         }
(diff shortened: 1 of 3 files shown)


legend-engine-application-query/src/main/java/org/finos/legend/engine/application/query/api/QueryStoreManager.java
```diff
@@ -229,7 +231,7 @@ public List<Query> getQueries(QuerySearchSpecification searchSpecification, Stri
         }
         if (searchSpecification.taggedValues != null && !searchSpecification.taggedValues.isEmpty())
         {
-            filters.add(Filters.or(
+            filters.add(Filters.and(
```
akphi, line 234: I don't think this is correct, the default behavior of search is `or`ing the conditions, you want a flag to customize this behavior, basically, `and` the tagged values condition together, maybe something like `SearchSpecification.combineTaggedValuesCondition: Boolean[0..1]`? Please make sure you document this well in code and Swagger

legend-engine-application-query/src/main/java/org/finos/legend/engine/application/query/api/QueryStoreManager.java
```diff
@@ -200,17 +197,22 @@ public static void validateQuery(Query query)
         // TODO: we can potentially create a pattern check for version
     }
 
+    public List<Query> getQuerySearchResults(QuerySearchSpecification searchSpecification, List<Bson> filters)
+    {
+        return LazyIterate.collect(this.getQueryCollection()
+                .find(filters.isEmpty() ? EMPTY_FILTER : Filters.and(filters))
+                .sort(searchSpecification.showLatestQueriesFirst != null && searchSpecification.showLatestQueriesFirst ? descending("lastUpdatedAt") : EMPTY_FILTER)
```
akphi, line 204: Like said, if we are already opinionated about how we sort the list of queries (i.e. returning user queries first), we can remove `showLatestQueriesFirst` flag, I don't see why we need that anymore

legend-engine-application-query/src/main/java/org/finos/legend/engine/application/query/api/QueryStoreManager.java
```diff
@@ -239,11 +241,20 @@ public List<Query> getQueries(QuerySearchSpecification searchSpecification, Stri
...
-                .limit(Math.min(MAX_NUMBER_OF_QUERIES, searchSpecification.limit == null ? Integer.MAX_VALUE : searchSpecification.limit)), QueryStoreManager::documentToQuery).toList();
+        filters.add(Filters.in("owner", currentUser, null));
+        List<Query> currentUserQueryResults = this.getQuerySearchResults(searchSpecification, filters);
+        if (searchSpecification.showCurrentUserQueriesOnly != null && searchSpecification.showCurrentUserQueriesOnly)
+        {
+            return currentUserQueryResults;
+        }
+        filters.remove(Filters.in("owner", currentUser, null));
+        filters.add(Filters.nin("owner", currentUser, null));
+        searchSpecification.limit = searchSpecification.limit != null ? searchSpecification.limit - currentUserQueryResults.size() : Integer.MAX_VALUE - currentUserQueryResults.size();
```
akphi, line 252: this looks hacky, say if we remove `showLatestQueriesFirst`, the only thing left from search specification that `getQuerySearchResults()` needs is the limit, we can just send in the limit and the filters

legend-engine-application-query/src/main/java/org/finos/legend/engine/application/query/api/QueryStoreManager.java
```diff
@@ -200,17 +197,22 @@ public static void validateQuery(Query query)
         // TODO: we can potentially create a pattern check for version
     }
 
+    public List<Query> getQuerySearchResults(QuerySearchSpecification searchSpecification, List<Bson> filters)
+    {
+        return LazyIterate.collect(this.getQueryCollection()
+                .find(filters.isEmpty() ? EMPTY_FILTER : Filters.and(filters))
+                .sort(searchSpecification.showLatestQueriesFirst != null && searchSpecification.showLatestQueriesFirst ? descending("lastUpdatedAt") : EMPTY_FILTER)
+                // NOTE: return a light version of the query to save bandwidth
+                .projection(Projections.include("id", "name", "versionId", "groupId", "artifactId", "owner", "createdAt", "lastUpdatedAt"))
```
akphi, line 206: Let's make this projection static final (constant), `LIGHT_QUERY_PROJECTION`

legend-engine-application-query/src/main/java/org/finos/legend/engine/application/query/api/QueryStoreManager.java
```diff
@@ -239,11 +241,20 @@ public List<Query> getQueries(QuerySearchSpecification searchSpecification, Stri
                     ListIterate.collect(searchSpecification.stereotypes, stereotype ->
                             Filters.and(Filters.eq("stereotypes.profile", stereotype.profile), Filters.eq("stereotypes.value", stereotype.value)))));
         }
-        return LazyIterate.collect(this.getQueryCollection()
-                .find(filters.isEmpty() ? EMPTY_FILTER : Filters.and(filters)).sort(searchSpecification.showLatestQueriesFirst != null && searchSpecification.showLatestQueriesFirst ? descending("lastUpdatedAt") : EMPTY_FILTER)
-                // NOTE: return a light version of the query to save bandwidth
-                .projection(Projections.include("id", "name", "versionId", "groupId", "artifactId", "owner", "createdAt", "lastUpdatedAt"))
-                .limit(Math.min(MAX_NUMBER_OF_QUERIES, searchSpecification.limit == null ? Integer.MAX_VALUE : searchSpecification.limit)), QueryStoreManager::documentToQuery).toList();
+        filters.add(Filters.in("owner", currentUser, null));
```
akphi, line 244: Let's add a new line here and add some docs explaining what we're trying to do here

legend-engine-application-query/src/test/java/org/finos/legend/engine/application/query/api/TestQueryStoreManager.java
```diff
@@ -734,4 +734,28 @@ public void testGetLightQueriesSortWithLastUpdatedAt() throws Exception
         Query lightQuery = queries.get(0);
```
akphi, line 734: `testGetLightQueriesSortWithLastUpdatedAt` --> `testSearchQueriesSortedByLastUpdatedAt`

legend-engine-application-query/src/main/java/org/finos/legend/engine/application/query/api/QueryStoreManager.java
```diff
@@ -200,17 +197,22 @@ public static void validateQuery(Query query)
...
         List<Bson> filters = new ArrayList<>();
-        if (searchSpecification.showCurrentUserQueriesOnly != null && searchSpecification.showCurrentUserQueriesOnly)
-        {
-            // NOTE: every user is considered owner of the queries created by unknown user
-            filters.add(Filters.in("owner", currentUser, null));
-        }
         if (searchSpecification.searchTerm != null)
         {
-            filters.add(Filters.regex("name", Pattern.quote(searchSpecification.searchTerm), "i"));
+            filters.add(Filters.or(Filters.regex("name", Pattern.quote(searchSpecification.searchTerm), "i"), Filters.regex("id", Pattern.quote(searchSpecification.searchTerm), "i")));
```
akphi, line 215: let's add doc here

legend-engine-application-query/src/test/java/org/finos/legend/engine/application/query/api/TestQueryStoreManager.java
```diff
@@ -734,4 +734,28 @@ public void testGetLightQueriesSortWithLastUpdatedAt() throws Exception
...
     }
+
+    @Test
+    public void testGetQueriesWithSearchByQueryId() throws Exception
+    {
+        String currentUser = "testUser";
+        queryStoreManager.createQuery(TestQueryBuilder.create("11", "query_a", currentUser).build(), currentUser);
+        queryStoreManager.createQuery(TestQueryBuilder.create("12", "query_b", currentUser).build(), currentUser);
+        queryStoreManager.createQuery(TestQueryBuilder.create("13", "query_c", currentUser).build(), currentUser);
+        Assert.assertEquals(3, queryStoreManager.getQueries(new TestQuerySearchSpecificationBuilder().build(), currentUser).size());
```
akphi, line 745: I would like us to be a little more refined here, allowing search id like this would harbor mumbo jumbo in our search results, I think the best way is to check if the search term is of a particular format (i.e. (uuid)[https://en.wikipedia.org/wiki/Universally_unique_identifier] or something slightly looser), then we would proceed to do the ID matching

legend-engine-application-query/src/main/java/org/finos/legend/engine/application/query/api/QueryStoreManager.java
akphi: adding my comment in test -- I would like us to be a little more refined here, allowing search id like this would harbor mumbo jumbo in our search results, I think the best way is to check if the search term is of a particular format (i.e. (uuid)[https://en.wikipedia.org/wiki/Universally_unique_identifier] or something slightly looser), then we would proceed to do the ID matching

legend-engine-application-query/src/test/java/org/finos/legend/engine/application/query/api/TestQueryStoreManager.java
```diff
@@ -734,4 +734,28 @@ public void testGetLightQueriesSortWithLastUpdatedAt() throws Exception
...
+        Assert.assertEquals(3, queryStoreManager.getQueries(new TestQuerySearchSpecificationBuilder().withSearchTerm("1").build(), currentUser).size());
+        Assert.assertEquals(1, queryStoreManager.getQueries(new TestQuerySearchSpecificationBuilder().withSearchTerm("2").build(), currentUser).size());
+        Assert.assertEquals(1, queryStoreManager.getQueries(new TestQuerySearchSpecificationBuilder().withSearchTerm("13").build(), currentUser).size());
+    }
+
+    @Test
+    public void testGetQueriesSortedWithCurrentUserFirst() throws Exception
```
akphi, line 752: --> `testSearchQueriesSortedByCurrentUserFirst`

akphi (changes requested):

akphi (changes requested):

akphi (approved):

PR #1617, opened 2023-03-31
update execute to use requestContext

AFine-gs:
#### What type of PR is this?

Improvement

Use wrapper class for accessory information that comes from http requests

### DIFF
--- a/legend-engine-shared-core/pom.xml
+++ b/legend-engine-shared-core/pom.xml
@@ -79,6 +79,13 @@
         </dependency>
         <!-- For Rest Result -->
 
+        <!-- For Request Context -->
+        <dependency>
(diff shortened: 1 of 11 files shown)


legend-engine-shared-core/pom.xml
```diff
@@ -79,6 +79,13 @@
         </dependency>
         <!-- For Rest Result -->
 
+        <!-- For Request Context -->
+        <dependency>
+            <groupId>javax.servlet</groupId>
+            <artifactId>javax.servlet-api</artifactId>
+        </dependency>
```
rafaelbey, line 86: We should not expose servlet related classes to plan execution to keep the execution agnostic of who triggered it.

PR #1638, opened 2023-04-06
Deleting redundant PureIDELight

haldes:
#### What type of PR is this?
Bug Fix

This is to remove redundant PureIDE class, which was used for testing.

NA

NA
No


aziemchawdhary-gs:
Can we also remove the dependencies on Pure IDE Light from the pom for this module?

Yasirmod17 (approved):

PR #1679, opened 2023-04-14
Persistence: Nontemporal Delta: Versioning + Stage Filter + Dedup

kumuwu:
#### What type of PR is this?

Improvement

1. Support for Max Version Strategy based versioning, filtering of stage table and deduplication in Nontemporal Delta milestoning scheme
2. Bug fix: Removed the digest check in Nontemporal Delta's insert statement to prevent from duplicated rows
3. Optimization: Only check for if staging table is empty in executor mode if the ingest mode is Unitemporal Snapshot or Bitemporal Snapshot
4. Associated tests in all the modules

Fixes #

No


prasar-ashutosh:
Changes looks good to me

epsstan (approved):

PR #1681, opened 2023-04-14
Generate freemarker functions for included enumeration mappings

gs-ssh16:
#### What type of PR is this?

Bug Fix

Freemarker functions are currently generated for only enumeration mappings at the top level. This PR extends it to those from included mappings

### DIFF
--- a/legend-engine-pure-code-compiled-core/src/main/resources/core/pure/mapping/mappingExtension.pure
+++ b/legend-engine-pure-code-compiled-core/src/main/resources/core/pure/mapping/mappingExtension.pure
@@ -131,6 +131,10 @@ function <<access.private>> meta::pure::mapping::updatePropertyMappingIDWithSuff
      );
 }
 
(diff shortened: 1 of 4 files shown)


legend-engine-pure-runtime-execution/pom.xml
```diff
@@ -147,6 +147,11 @@
             <groupId>org.finos.legend.engine</groupId>
             <artifactId>legend-engine-pure-platform-dsl-mapping-java</artifactId>
         </dependency>
+        <dependency>
```
hardikmaheshwari, line 150: I believe runtime-execution module is expected to be light weight. Should this dependency be added to ide-light module? or where this test executes in pipeline?

gs-ssh16: Moved this dependency to ide light

hardikmaheshwari (approved):

PR #1697, opened 2023-04-18
Sourcetree Calc: Fix unions with differing source property levels

aziemchawdhary-gs:
#### What type of PR is this?

Bugfix

Fix source tree calculation for union mappings where one union element is starting from the source and another from property access


davidharte-gs (approved):

aziemchawdhary-gs:
Fixed a missing case in a following commit: https://github.com/finos/legend-engine/pull/1751

PR #1708, opened 2023-04-20
Persistence: Implement CLUSTER BY clause for snowflake

kumuwu:
#### What type of PR is this?

Improvement

To allow the user to create a table in snowflake with "cluster by"

Fixes #

no


legend-engine-xt-persistence-component/legend-engine-xt-persistence-component-logical-plan/src/main/java/org/finos/legend/engine/persistence/components/logicalplan/datasets/ClusterKeyAbstract.java
```diff
@@ -0,0 +1,35 @@
...
+    typeAbstract = "*Abstract",
+    typeImmutable = "*",
+    jdkOnly = true,
+    optionalAcceptNullable = true,
+    strictBuilder = true
+)
+public interface ClusterKeyAbstract extends LogicalPlanNode
+{
+    @Parameter(order = 0)
+    Value clusterKeyValue();
```
prasar-ashutosh, line 34: Let's just call it: key

legend-engine-xt-persistence-component/legend-engine-xt-persistence-component-relational-core/src/main/java/org/finos/legend/engine/persistence/components/relational/sqldom/constraints/table/ClusterKey.java
```diff
@@ -0,0 +1,39 @@
...
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+package org.finos.legend.engine.persistence.components.relational.sqldom.constraints.table;
+
+import org.finos.legend.engine.persistence.components.relational.sqldom.SqlDomException;
+import org.finos.legend.engine.persistence.components.relational.sqldom.SqlGen;
+import org.finos.legend.engine.persistence.components.relational.sqldom.schemaops.values.Value;
+
+public class ClusterKey implements SqlGen
```
prasar-ashutosh, line 21: Should we call it ClusteringKeyConstraint

legend-engine-xt-persistence-component/legend-engine-xt-persistence-component-logical-plan/src/main/java/org/finos/legend/engine/persistence/components/logicalplan/datasets/ClusterKeyAbstract.java
kumuwu: Sure, fixed!

legend-engine-xt-persistence-component/legend-engine-xt-persistence-component-relational-core/src/main/java/org/finos/legend/engine/persistence/components/relational/sqldom/constraints/table/ClusterKey.java
kumuwu: Sure, fixed!

legend-engine-xt-persistence-component/legend-engine-xt-persistence-component-relational-core/src/main/java/org/finos/legend/engine/persistence/components/relational/api/RelationalIngestorAbstract.java
```diff
@@ -428,4 +451,33 @@ private Optional<Long> getNextBatchId(Datasets datasets, Executor<SqlGen, Tabula
...
+                    .findFirst()
+                    .map(TabularData::getData)
+                    .flatMap(t -> t.stream().findFirst())
+                    .orElseThrow(IllegalStateException::new);
+                // Put into map only when not null
+                Object lower = resultMap.get(MIN_OF_FIELD);
+                Object upper = resultMap.get(MAX_OF_FIELD);
+                if (lower != null && upper != null)
+                {
+                    map.put(filter, Arrays.asList(lower, upper));
```
rafaelbey, line 476: Contract will be cleaner if we put this on a Tuple, as we always will have two values rather than a list.  On the caller, it will be tuple.first, tuples.second rather than 'magic' index values. for the array.

legend-engine-xt-persistence-component/legend-engine-xt-persistence-component-relational-core/src/main/java/org/finos/legend/engine/persistence/components/relational/api/RelationalIngestorAbstract.java
```diff
@@ -387,10 +391,29 @@ private Map<String, String> extractPlaceHolderKeyValues(Datasets datasets, Execu
...
         {
             placeHolderKeyValues.put(BATCH_ID_PATTERN, nextBatchId.get().toString());
         }
+        if (optimizationFilters.isPresent())
+        {
+            for (OptimizationFilter filter : optimizationFilters.get().keySet())
+            {
+                Object lowerBound = optimizationFilters.get().get(filter).get(0);
+                Object upperBound = optimizationFilters.get().get(filter).get(1);
+                if (lowerBound instanceof Date)
```
rafaelbey, line 405: I was trying to fin a test case for this, and not sure I saw one.  I was hoping to get more details into what these quotes do, and why just Date does not required quoting.  Can you point me to a test or add one?  Also explain the rational of the if/else?

kumuwu: The test that I added in UnitemporalDeltaTest.java in h2 module tests both the if branch and else branch. In the test there are two optimization filters, "expiryDate" of date type and "id" of integer type.

The rationale of having this block is as follows:

Firstly, we always produce the placeholders text surrounded by single quotes, example below:
`(sink."column" >= **'{ID_LOWER_BOUND}'**) AND (sink."column" <= **'{ID_UPPER_BOUND}'**)`

When we replace the placeholders with actual values, we want to do different things with different types (we only support date type or number type for optimization filters):

- For date type, we want to retain the single quotes, such that after replacement it will look like this: `(sink."column" >= '2023-01-01') AND (sink."column" <= '2023-01-02')`
- For number type, we do not want to retain the single quotes, such that after replacement it will look like this: `(sink."column" >= 1) AND (sink."column" <= 2)`

rafaelbey: But the code tells a different story.  If you have a Date (look at the if condition `if (lowerBound instanceof Date)`), don't add quotes, hence why I was confuse, and trying to understand the rational for this change.  

Do Im missing something or is the code buggy and we should look into it?  I suspect a better instacne of will be `if (lowerbound instanceof Number)` so that non-numbers keep the quotes.

davidsunglee: Think the code is okay. I think the confusion is `placeHolderKeyValues` appears to be a map where the key is the 'find' term and the value is the 'replace' term.

So for the if clause (Date), the pattern to search for is just `{LOWER_BOUND}` with no quotes, so the quotes are kept upon replacement.

Whereas for the else clause, the pattern to search for is `'{LOWER_BOUND}'` with quotes, so the quotes will be removed upon replacement.

Test does look like it covers this.

davidsunglee: Separate question though: should we switch from `else` to something like `else if (lowerBound instanceof Number)` and add a separate `else` clause that throws?

prasar-ashutosh: These are the supported data types for lower bound and Upper Bound values:
INT, INTEGER, INT64, BIGINT, FLOAT, DOUBLE, DECIMAL, DATE
Out of them , only DATE needs a quote and other data types do not need a quote.
That's why it was written with if and a common else block.

I think we can change this to switch case to be more cleaner.

legend-engine-xt-persistence-component/legend-engine-xt-persistence-component-relational-core/src/main/java/org/finos/legend/engine/persistence/components/relational/api/RelationalIngestorAbstract.java
kumuwu: Sure, fixed this!

legend-engine-xt-persistence-component/legend-engine-xt-persistence-component-relational-core/src/main/java/org/finos/legend/engine/persistence/components/relational/api/RelationalIngestorAbstract.java
kumuwu: > Separate question though: should we switch from `else` to something like `else if (lowerBound instanceof Number)` and add a separate `else` clause that throws?

Sure, fixed this!

rafaelbey: Thanks for the explanation!  This make sense now, but I feel still too convoluted :) 

While not a feedback item to address now, we should change the template query to avoid having the quotes there, and let the value we want to replace with come with the quotes, if needed.

For example:  `where COL >= {LOWER_BOUND}`

then the replacement pair will provide the quotes to the value, rather than the other way around:

```java
placeHolderKeyValues.put(filter.lowerBoundPattern(), SINGLE_QUOTE + lowerBound.toString() + SINGLE_QUOTE);
```

rafaelbey (approved):

PR #1729, opened 2023-04-25
Use byte type with internalize instead of ByteStream

hardikmaheshwari:
#### What type of PR is this?
Improvement

Use byte type with internalize instead of ByteStream

Fixes #


aziemchawdhary-gs (approved):

legend-engine-language-pure-grammar/src/main/java/org/finos/legend/engine/language/pure/grammar/to/DEPRECATED_PureGrammarComposerCore.java
```diff
@@ -881,9 +882,9 @@ public String visit(CLatestDate cLatestDate)
     }
 
     @Override
-    public String visit(CByteStream cByteStream)
+    public String visit(CByteArray cByteArray)
     {
-        return "byteStream(" + HelperValueSpecificationGrammarComposer.renderString(cByteStream.value, this) + ")";
+        return "byteArray(" + HelperValueSpecificationGrammarComposer.renderString(new String(cByteArray.value, StandardCharsets.UTF_8), this) + ")";
```
kevin-m-knight-gs, line 887: Are we sure we want to render the bytes in this way? If the bytes originally came from a string, it will be fine; but if they didn't, it might well yield a problematic string (possibly including newlines and special characters and such). Maybe it would be better to encode them in base 64 or something.

legend-engine-xt-flatdata-pure/src/main/resources/core_external_format_flatdata/executionPlan/tests/utils.pure
```diff
@@ -17,9 +17,9 @@ import meta::legend::*;
 import meta::pure::extension::*;
 import meta::pure::runtime::*;
 
-function meta::external::format::flatdata::executionPlan::test::conertStringToByteStream(data:String[1]): InstanceValue[1]
+function meta::external::format::flatdata::executionPlan::test::convertStringToByteArray(data:String[1]): InstanceValue[1]
```
kevin-m-knight-gs, line 20: I think convertStringToBytes is a better name, since in some contexts we won't be interpreting Byte[*] as an array.
```suggestion
function meta::external::format::flatdata::executionPlan::test::convertStringToBytes(data:String[1]): InstanceValue[1]

```

legend-engine-xt-flatdata-pure/src/main/resources/core_external_format_flatdata/executionPlan/tests/utils.pure
```diff
@@ -17,9 +17,9 @@ import meta::legend::*;
 import meta::pure::extension::*;
 import meta::pure::runtime::*;
 
-function meta::external::format::flatdata::executionPlan::test::conertStringToByteStream(data:String[1]): InstanceValue[1]
+function meta::external::format::flatdata::executionPlan::test::convertStringToByteArray(data:String[1]): InstanceValue[1]
 {
-  ^InstanceValue(genericType=^GenericType(rawType=ByteStream), multiplicity=PureOne, values=$data)
+  ^InstanceValue(genericType=^GenericType(rawType=Byte), multiplicity=ZeroMany, values=$data)
```
kevin-m-knight-gs, line 22: Don't we actually need to convert the data here?

legend-engine-language-pure-grammar/src/main/antlr4/org/finos/legend/engine/language/pure/grammar/from/antlr4/core/M3LexerGrammar.g4
```diff
@@ -10,7 +10,7 @@ LET:                                        'let';
 ALL_VERSIONS:                               'allVersions';
 ALL_VERSIONS_IN_RANGE:                      'allVersionsInRange';
 
-BYTE_STREAM_FUNCTION:                       'byteStream';
+BYTE_ARRAY_FUNCTION:                        'byteArray';
```
kevin-m-knight-gs, line 13: I think it would be better to call this "toBytes" or something like that. Calling it "byteArray" seems to imply that Byte[*] is implemented as an array, which is not always going to be the case.
```suggestion
BYTE_ARRAY_FUNCTION:                        'toBytes';
```

kevin-m-knight-gs (changes requested):

legend-engine-executionPlan-execution/src/main/java/org/finos/legend/engine/plan/execution/validation/FunctionParameterTypeValidator.java
```diff
@@ -192,7 +192,7 @@ protected boolean canParse(String parameterValue)
         }
     };
 
-    private static final FunctionParameterTypeValidator byteStreamValidator = new FunctionParameterTypeValidator("ByteStream")
+    private static final FunctionParameterTypeValidator byteValidator = new FunctionParameterTypeValidator("Byte")
```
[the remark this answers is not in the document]
hardikmaheshwari: hmm I am not sure how would that be used so don't want to allow it until we have a clear picture

legend-engine-xt-flatdata-pure/src/main/resources/core_external_format_flatdata/executionPlan/tests/utils.pure
hardikmaheshwari: Do we have a functionality to get bytes from a string in pure?

legend-engine-xt-flatdata-pure/src/main/resources/core_external_format_flatdata/executionPlan/tests/utils.pure
hardikmaheshwari: makes sense done

legend-engine-language-pure-grammar/src/main/antlr4/org/finos/legend/engine/language/pure/grammar/from/antlr4/core/M3LexerGrammar.g4
hardikmaheshwari: done

kevin-m-knight-gs (approved):

legend-engine-language-pure-grammar/src/main/java/org/finos/legend/engine/language/pure/grammar/to/DEPRECATED_PureGrammarComposerCore.java
hardikmaheshwari: We essentially use this in context of service tests data so we can't have an encoding else it won't be possible for users to work with tests.
Regarding special characters like new-line our antlr parser only allow STRING token so new-line or anything special should already have been handled appropriately. 
Test : testBindingWithService()

PR #1730, opened 2023-04-25
fixed reverse function (alloy) and added test for it

michalmacu:
#### What type of PR is this?

Bug Fix

If you call reverse() function in Alloy Studio it will throw Java error "void cannot be converted to java.util.List<>"

The one described above.

The issue was cause by the fact that someone assumed that Collections.reverse() returns the reversed collection, but it acutally returns void and modifies the collection passed as an argument.

No.


davidharte-gs (approved):

aziemchawdhary-gs (approved):

aziemchawdhary-gs:
Build passed internally

PR #1732, opened 2023-04-25
entitlements - fix a bug in retrieving tables from a circular database

YannanGao-gs:
#### What type of PR is this?

Bug Fix

fix a bug in retrieving tables from a circular database

Fixes #


akphi (approved):
I think we can merge this change as the workaround to mitigate the current issue, we should be more concerned about the fact that circular dependencies of store does not throw off compiler

PR #1735, opened 2023-04-26
feature: added modulo function with test (alloy)

michalmacu:
#### What type of PR is this?

Improvement

The mod_Integer__Integer_Integer_ function was avalible in PURE, but had no implementation for Alloy. This PR adds this function and test for it.

It's a trivial function using Javas modulo operator.

No

### DIFF
--- a/legend-engine-executionPlan-dependencies/src/main/java/org/finos/legend/engine/plan/dependencies/util/Library.java
+++ b/legend-engine-executionPlan-dependencies/src/main/java/org/finos/legend/engine/plan/dependencies/util/Library.java
@@ -521,11 +521,6 @@ public class Library
         }
     }
 
-    public static long mod(long left, long right)
-    {
-        return left % right;
-    }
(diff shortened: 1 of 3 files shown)


aziemchawdhary-gs:
Instead of adding and calling mod in the Library, could we not just call j_remainder on the inputs since we know that the type of mod are both Pure Integers? 
Usually the math functions in Library code are written to handle different types of numbers but I don't think you need to handle that in the generation of mod

michalmacu:
> Instead of adding and calling mod in the Library, could we not just call j_remainder on the inputs since we know that the type of mod are both Pure Integers? Usually the math functions in Library code are written to handle different types of numbers but I don't think you need to handle that in the generation of mod, hence no need to add an implementation of mod to Library.

That is a valid point. I will re-write it as you've suggested. I though that this might be the case, but I didn't manage to notice `j_remainder()`

PR #1739, opened 2023-04-27
Add referral header to queryTag

mrudula-gs:
#### What type of PR is this?
Improvement

Adds referral header info to the session variable QUERY_TAG for snowflake executions along with the executionTraceID and user info in a json format.

Post execution, snowflake queries will populate the QUERY_TAG of query history with some sql execution metadata info and this change adds referral header to the existing info.


legend-engine-executionPlan-execution/src/main/java/org/finos/legend/engine/plan/execution/PlanExecutor.java
```diff
@@ -285,7 +286,7 @@ public Result executeWithArgs(ExecuteArgs executeArgs)
         }
     }
 
-    private void setUpState(SingleExecutionPlan singleExecutionPlan, ExecutionState state, MutableList<CommonProfile> profiles, String user)
+    public static void setUpState(SingleExecutionPlan singleExecutionPlan, ExecutionState state, MutableList<CommonProfile> profiles, String user)
```
rafaelbey, line 289: Can we keep this private and make sure the test executes properly so that it goes thru this code the same way production does it?

legend-engine-executionPlan-execution/src/test/java/org/finos/legend/engine/plan/execution/nodes/helpers/freemarker/TestFreeMarkerExecutor.java
```diff
@@ -47,4 +56,29 @@ public static String collectionSizeTemplate()
                 "<#return collection?size> " +
                 "</#function>";
     }
+
+    @Test
+    public void testQueryTagReferralHeaderPlaceHolder()
```
rafaelbey, line 61: This class does not seem a good place for this test.  This should be part of plan execution.  
The issue here is that you are not really testing what happens in production, and actually you end duplicating that flow.

AFine-gs (approved):

PR #1741, opened 2023-04-28
M2M feature parity: equals method

tanujgirdhar:

Fixes #


legend-engine-executionPlan-dependencies/src/main/java/org/finos/legend/engine/plan/dependencies/util/Library.java
```diff
@@ -1416,4 +1417,96 @@ public static boolean pureAssert(boolean test, Supplier<String> message)
         }
         return true;
     }
+
+    public static boolean equal(Object left, Object right)
```
rafaelbey, line 1421: Not sure how this is tested on this change.  I expected a change to the `TestLibrary`, unless this is tested elsewhere.

PR #1743, opened 2023-04-28
Legend SQL - Add Kerberos authentication and delegation

gs-manvig:
#### What type of PR is this?
Improvement

Introduce Kerberos authentication and delegation to the Legend SQL engine

Fixes #

### DIFF
--- a/legend-engine-xt-sql-postgres-server/src/main/java/org/finos/legend/engine/postgres/auth/KerberosIdentityProvider.java
+++ b/legend-engine-xt-sql-postgres-server/src/main/java/org/finos/legend/engine/postgres/auth/KerberosIdentityProvider.java
@@ -18,6 +18,8 @@ import org.finos.legend.engine.shared.core.identity.Identity;
 import org.finos.legend.engine.shared.core.identity.factory.DefaultIdentityFactory;
 import org.finos.legend.engine.shared.core.kerberos.SubjectTools;
 
+import javax.security.auth.Subject;
+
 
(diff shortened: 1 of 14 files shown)


shtirlets:
Looks good

pierredebelen (dismissed):

legend-engine-xt-sql-postgres-server/src/main/java/org/finos/legend/engine/postgres/auth/KerberosIdentityProvider.java
```diff
@@ -26,4 +28,9 @@ public Identity getIdentityForPassword(String userName, SecureString passwd)
     {
         return DefaultIdentityFactory.INSTANCE.makeIdentity(SubjectTools.getSubjectFromUsernamePassword(userName, passwd.getChars()));
     }
+
+    public static Identity getIdentityForSubject(Subject subject)
+    {
+        return DefaultIdentityFactory.INSTANCE.makeIdentity(subject);
```
rafaelbey, line 34: We should not use `DefaultIdentityfactory` but rather use `IdentityFactoryProvider`.  This will respect whatever implementation is set by the application.

gs-manvig: Makes sense. This has been updated

rafaelbey (approved):

PR #1757, opened 2023-05-03
reverse combine order of filters

PrateekGarg-gs:
#### What type of PR is this?

Improvement

earlier the order of and/or filter condition combination was reversed. Now keep the order as natural order described by user in pure query. As some databases process sql filters in order and only process certain condtions if previous condition is processed

Fixes #


finos-admin:

finos-admin:

PR #1759, opened 2023-05-03
 M2M feature parity: equals method

tanujgirdhar:
#### What type of PR is this?
- Improvement

Same behavior for equals method between PURE and Alloy

Fixes #

No


tanujgirdhar:
many collection functions use .equals internally. must override the equals method in the class implementation to handle that scenario.

PR #1765, opened 2023-05-04
Router: enable preeval for marked functions

aziemchawdhary-gs:
#### What type of PR is this?

Improvement 

Enables the router to call preeval to normalize functions prior to being passed to the router. 
Only functions marked with NormalizeRequiredFunction stereotype will be processed by preeval

This change removes the need to explicitly call preeval prior to making calls to execute for queries using marked functions.

The change is dependent on https://github.com/finos/legend-pure/pull/694


legend-engine-pure-code-compiled-core/src/main/resources/core/pure/router/routing/router_routing.pure
```diff
@@ -162,22 +162,38 @@ function <<access.private>> meta::pure::router::routing::routeValueSpecification
...
      f:Function<Any>[1]       | print(if($debug.debug,|$debug.space+if($f->instanceOf(NativeFunction),|'NativeFunction',|'FunctionDefinition')+' : '+$f->asString()+'\n',|''));
                                 let specializedFunctionProcessor = specializedFunctionExpressionRouterProcessor($extensions)->filter(e|$e.first->eval($fe)).second;
                                 if($specializedFunctionProcessor->isNotEmpty(),
                                   | assert($specializedFunctionProcessor->size() == 1, | 'Only one match is authorized for specializedFunctionProcessor');
                                     $specializedFunctionProcessor->toOne()->evaluate([list($f), list($fe), list($state), list($executionContext), list($vars), list($inScopeVars), list($extensions), list($debug)])->toOne()->cast(@RoutingState);,
-                                  | $f->routeFunctionExpressionFunctionDefinition($fe, $state, $executionContext, $vars, $inScopeVars->putAll($f->openVariableValues()), $extensions, $debug));
+                                  | let newFe = prevalFunctionExpressionIfRequired($fe,$inScopeVars, $extensions, $debug);
```
hardikmaheshwari, line 188: What's you view on doing preeval processing for all function whether specialized processor exists or not? I see 2 views : 
1. If we do it before we can centralize that logic call from one place.
2. If we do it after we say someone has specified specialized processor and they should have full control

aziemchawdhary-gs: After discussing I think we want to go with 1 - have modified the change to process prior to specialized processors

hardikmaheshwari (approved):

PR #1772, opened 2023-05-08
Set Mongo host from container host

ghub-real:
#### What type of PR is this?
- Bug Fix

Fix failing test - with the mongo hostname also parameterized in addition to port number.

### DIFF
--- a/legend-engine-xt-nonrelationalStore-mongodb-executionPlan-test/src/main/resources/core_mongodb_execution_test/test_setup/testSetupDSL.txt
+++ b/legend-engine-xt-nonrelationalStore-mongodb-executionPlan-test/src/main/resources/core_mongodb_execution_test/test_setup/testSetupDSL.txt
@@ -89,7 +89,7 @@ MongoDBConnection test::mongodb::testConnection
 {
     database: userDatabase;
     store: test::mongodb::mydatabase;
-    serverURLs: [localhost:__PORT_TOKEN__];
+    serverURLs: [__HOST_TOKEN__:__PORT_TOKEN__];
     authentication: # UserPassword {
         username: 'sa';
         password: SystemPropertiesSecret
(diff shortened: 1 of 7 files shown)


rafaelbey (dismissed):

rafaelbey:
We are missing replacing localhost here: https://github.com/finos/legend-engine/blob/master/legend-engine-xt-nonrelationalStore-mongodb-executionPlan-test/src/main/java/org/finos/legend/engine/plan/execution/stores/mongodb/test/shared/MongoDBCommands.java#L132

legend-engine-xt-nonrelationalStore-mongodb-executionPlan-test/src/main/resources/core_mongodb_execution_test/test_setup/testSetupDSL.txt
```diff
@@ -89,7 +89,7 @@ MongoDBConnection test::mongodb::testConnection
 {
     database: userDatabase;
     store: test::mongodb::mydatabase;
-    serverURLs: [localhost:__PORT_TOKEN__];
+    serverURLs: [__HOST_TOKEN__:__PORT_TOKEN__];
```
rafaelbey, line 92: All this works from connectivity perspective, but when the hostname have characters like '-', the parser fails.  This is a limitation on what kind of hostname you will be supporting, and one that we should fill the gap to fix some of the e2e testing we do.

ghub-real: Fixed the parser to accept - as well, but we might have to support a wider set of values potentially.

rafaelbey (approved):

rafaelbey (approved):

PR #1774, opened 2023-05-09
fix: divide scale issue

michalmacu:
#### What type of PR is this?

Bug Fix

Previous divide() function implementation which used Java.math.BigDecimal.divide() failed to specify scale for the operation which is a big issue as the default scale will be scale of the first number in the division which causes unpredictable behaviour. For example dividing 1 by 2 will just result in 1.

|left|right|pre fix|post fix|
|----|-----|-------|--------|
|1   |2    |0.5    |0.5     |
|1D  |2D   |1      |0.5     |
|1.0D|2D   |0.5    |0.5     |
|1D  |2.0D |1      |0.5     |
|1D  |2    |1      |0.5     |
|1D  |2.0  |1      |0.5     |
|1   |2D   |1      |0.5     |
|1.0 |2D   |0.5    |0.5     |
|1.0 |2.0  |0.5    |0.5     |

I came up with the 34 value for scale by running few tests in PURE IDE, but it would be prudent to consult this value with someone more aware of the real needs of PURE language in Alloy context.

No.


aziemchawdhary-gs (changes requested):
I think we should take a different approach here: we should enable divide_Decimal_1__Decimal_1__Integer_1__Decimal_1_ in Engine and leave the implementation of divide as is
Users using divide with decimal should use the above function to set the scale

aziemchawdhary-gs:
PR to add the missing decimal function is here: https://github.com/finos/legend-engine/pull/1824

aziemchawdhary-gs:
Closing this PR as the better solution is to add the missing divide with scale function: https://github.com/finos/legend-engine/pull/1824

PR #1806, opened 2023-05-16
Skip executing service tests that are always true

pbt-santos:
Avoids using resources on tests that always return true. Returns empty tests indicating tests were not actually run.

Fixes tests due to skipped assertion.

#### What type of PR is this?

Improvement

Skips legacy tests whose assertions simply return true. This will avoid using resources on tests that do not add value.

Fixes #

Skipped test results will be displayed as empty.


gs-ssh16:
Though the test doesn't assert anything, test running - generating SQL, running the SQL on empty tables successfully is by itself an important part

rafaelbey:
I dont think there is much value on executing against H2 since the user real database is of different type.  All this will be doing is testing that we can generate 'some' H2 flavor SQL syntax and that this is executable.   By no means this equates to 1\ valid translation from pure to sql (given that we are not asserting output), and 2\ valid sql for the user db type.  

My understanding of these test cases are for functional purpose - is the user business logic correct and we translate correctly. 
 For that we need to compare against an expected output.

For testing H2 equivalence and correctness vs other DB types, the platform itself can do this as part thru unit testing and automated builds.

As for generating SQL, the plan will always generated SQL for the intended database type regardless of testing too.

gs-ssh16:
> I dont think there is much value on executing against H2 since the user real database is of different type. All this will be doing is testing that we can generate 'some' H2 flavor SQL syntax and that this is executable. By no means this equates to 1\ valid translation from pure to sql (given that we are not asserting output), and 2\ valid sql for the user db type.
> 
> My understanding of these test cases are for functional purpose - is the user business logic correct and we translate correctly. For that we need to compare against an expected output.
> 
> For testing H2 equivalence and correctness vs other DB types, the platform itself can do this thru unit testing and automated builds.
> 
> As for generating SQL, the plan will always generated SQL for the intended database type regardless of testing too.

At present, in most cases, SQLs across databases are structurally same with syntax differences. Given this flow is already deprecated in favor of the new service testing framework, I don't see any reason why we should fork more within it and remove some guards it provides against structurally invalid SQLs. Project build would have failed before if a platform upgrade generates structurally invalid SQLs, but it wouldn't if this change goes in (which involves some risk).

Also, in the cases where the lambda is something like {res | 2 == 2}, tests still run - which will raise a question why we are doing something special only in limited cases.

hardikmaheshwari:
To add to @gs-ssh16 point. If we really want to stop running such tests and are removing some safeguards we possibly should let users know of such changes. What are your comments on this approach - 
When we compile asserts we check if user is using the result variable in lambda body or not. If not we can be sure that such asserts are noOp and can fail registration for such case.
Then users have option to either remove their test and opt-in into no safety net approach or write a test using strategic approach.
I feel this way we would be moving towards platform strategic approach and clean users tests actually.

I realize that users can write noOp test using result variable in lambda body too but I haven't see such a case yet

finos-admin:

finos-admin:

PR #1816, opened 2023-05-18
Add preeval Store test case

rafaelbey:

Fixes #

### DIFF
--- a/legend-engine-pure-code-compiled-core/src/main/resources/core/pure/router/preeval/tests.pure
+++ b/legend-engine-pure-code-compiled-core/src/main/resources/core/pure/router/preeval/tests.pure
@@ -1499,7 +1499,7 @@ function meta::pure::router::preeval::tests::getTestRuntimeWithModelConnection(s
 
 function <<access.private>> meta::pure::router::preeval::tests::assertRoundTrip(input : FunctionDefinition<Any>[1], expectedOrig : FunctionDefinition<Any>[1]):Boolean[1]
 {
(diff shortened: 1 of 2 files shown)


legend-engine-pure-code-compiled-core/src/main/resources/core/pure/router/preeval/tests.pure
```diff
@@ -1499,7 +1506,7 @@ function meta::pure::router::preeval::tests::getTestRuntimeWithModelConnection(s
 
 function <<access.private>> meta::pure::router::preeval::tests::assertRoundTrip(input : FunctionDefinition<Any>[1], expectedOrig : FunctionDefinition<Any>[1]):Boolean[1]
 {
-  assertRoundTrip($input, $expectedOrig, debug())
```
aormerod-gs, line 1502: As discussed offline, this is being switched off to reduce output in the automated builds (and is probably more useful to investigation, so can be switched on then)

aormerod-gs (dismissed):

legend-engine-pure-code-compiled-core/src/main/resources/core/pure/router/preeval/tests.pure
rafaelbey: As discussed - this clutter logs when running thru CICD.  This to me is more a tool developers have to fix things locally, and we should avoid outputting this to github action logs.

aormerod-gs (approved):

PR #1821, opened 2023-05-19
Storeexupdate

AFine-gs:

allow cancellation by either session or request id


legend-engine-executionPlan-execution/src/main/java/org/finos/legend/engine/plan/execution/stores/StoreExecutableManager.java
```diff
@@ -56,48 +61,77 @@ public void addExecutable(String requestID, StoreExecutable execution)
 
     public void addExecutable(RequestContext context, StoreExecutable execution)
     {
-        String requestID = RequestContext.getRequestToken(context);
+        String requestID = RequestContext.getRequestToken(context);  //header if its there else it's the session
+
+        if (isRegistered && requestID != null)
+        {
+            List<String> foo = sessionIDToProvidedID.computeIfAbsent(RequestContext.getSessionID(context), x -> Collections.synchronizedList(new ArrayList<>()));
```
rafaelbey, line 68: can we name this variable on a more descriptive way?

legend-engine-executionPlan-execution/src/main/java/org/finos/legend/engine/plan/execution/stores/StoreExecutableManager.java
```diff
@@ -34,6 +36,8 @@
 {
     INSTANCE;
     private final ConcurrentHashMap<String, List<StoreExecutable>> requestExecutableMap = new ConcurrentHashMap<>();
+
+    private final ConcurrentHashMap<String, List<String>> sessionIDToProvidedID = new ConcurrentHashMap<>();
```
rafaelbey, line 40: The way things are been use, seems better to leverage a Set rather than a List. 

This will help on the remove statements, while also avoiding the need of checking if contains, as duplicates will have no effect.  

Thoughts?

rafaelbey (approved):

PR #1833, opened 2023-05-23
Add mongo dependency and override to core config

hardikmaheshwari:
#### What type of PR is this?
Bug Fix

Add mongo dependency and override to core config. This is blocking tests execution from IDE

Fixes #

### DIFF
--- a/legend-engine-pure-code-compiled-core-configuration/pom.xml
+++ b/legend-engine-pure-code-compiled-core-configuration/pom.xml
@@ -101,6 +101,16 @@
                     </dependency>
                     <dependency>
                         <groupId>org.finos.legend.engine</groupId>
+                        <artifactId>legend-engine-xt-nonrelationalStore-mongodb-pure</artifactId>
+                        <version>${project.version}</version>
+                    </dependency>
+                    <dependency>
+                        <groupId>org.finos.legend.engine</groupId>
+                        <artifactId>legend-engine-xt-nonrelationalStore-mongodb-javaPlatformBinding-pure</artifactId>
+                        <version>${project.version}</version>
+                    </dependency>
+                    <dependency>
+                        <groupId>org.finos.legend.engine</groupId>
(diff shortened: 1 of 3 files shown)


rafaelbey:
I don't think this is the right solution since this create a "super" module in production, that affects downstream projects that depend on this module (and they might not need mongo).  Conceptually, this also goes against the idea that the system is modular.

To test on IDE, we need an IDE solution.  Maybe a local binder that overrides everything inside the IDE pure code?

hardikmaheshwari:
> I don't think this is the right solution since this create a "super" module in production, that affects downstream projects that depend on this module (and they might not need mongo). Conceptually, this also goes against the idea that the system is modular.
> 
> To test on IDE, we need an IDE solution. Maybe a local binder that overrides everything inside the IDE pure code?
> 
> Otherwise, we might want to rollback the "local execution" from using this local binder and hard code java platform until a proper solution is brought that allow local binder while also respecting modularization.

@rafaelbey core-configuration module was meant to be a module aligned with server deployment and hence it is meant to be a super module which brings together dependencies for all features for a particular deployment.

rafaelbey:
I dont agree with that design approach. If you consider all other design choices on how extensions are built, this repo just minimize if not invalidate them. 

For example, all extensions define their parsers, composers, compilers, executions, etc extensions in ways that are discovered, and everywhere we make usage of this design choice. 

This means, that with just adding these extensions as runtime dependencies on classpath, like in the Server, the process can operate with these extensions without further lines of code. 

Hence, we should solve the local binder problem in a way that, well, is actually local, while further embrace the independence of the extensions in ways that they are "plug & play". 

So, unless Im missing what value this coreconfig repo provide that the extension design does not, and Im happy to be corrected, our direction should be phase it out with solutions that work properly with the distributed design.

ghub-real:
We don't need this anymore, do we?
(The local IDE tests need some additions - is that what we are looking to resolve?)

finos-admin:

PR #1861, opened 2023-05-29
Enable executions from Legend Studio & prod version support

ghub-real:
…… (#30)

* added getExtraElementForPathToElementRegisters in MongoDBCompilerExtension

* added v1_31_0 protocol

#### What type of PR is this?
Improvement

Add PathToElementRegisters for Legend studio to load the Mongo execution plan serializers. (via compiler extension)
Add a fixed version (1_31_0) - to enable production version support in engine for Mongo


ghub-real:
Need to remove protocol files (v1_31_0)

PR #1868, opened 2023-05-31
Dataspaces can be included and localized into relational stores.

haroonsherjan:
#### What type of PR is this?

improvement

Consumers of a dataspace can localize the underlying mapping into the consumer's store.

New DSL for users


legend-engine-language-pure-compiler/src/main/java/org/finos/legend/engine/language/pure/compiler/toPureGraph/PackageableElementFourthPassBuilder.java
```diff
@@ -154,6 +157,11 @@ public PackageableElement visit(Measure measure)
     public PackageableElement visit(Mapping mapping)
     {
         final org.finos.legend.pure.m3.coreinstance.meta.pure.mapping.Mapping pureMapping = this.context.pureModel.getMapping(this.context.pureModel.buildPackageString(mapping._package, mapping.name), mapping.sourceInformation);
+        if (mapping.classMappings != null && pureMapping._classMappings().isEmpty())
+        {
+            RichIterable<Pair<SetImplementation, RichIterable<EmbeddedSetImplementation>>> setImplementations = ListIterate.collect(mapping.classMappings, cm -> cm.accept(new ClassMappingFirstPassBuilder(this.context, pureMapping)));
```
aziemchawdhary-gs, line 162: Do you need to do this? The PackageableElementThirdPassBuilder already does this so the pureMapping should have the class mappings processed and populated by the time this pass runs. 
Is there a flow where this doesn't happen?

PR #1875, opened 2023-06-02
Add Prometheus JMX agent to expose JVM platform jmx metrics for monitoring

shtirlets:
#### What type of PR is this?
Improvement

Add Prometheus JMX agent to expose JVM platform jmx metrics for monitoring

Run configurations / scripts will need to be updated to activate / expose the JVM agent. See https://github.com/prometheus/jmx_exporter#running-the-java-agent for more details.

### DIFF
--- a/legend-engine-xt-sql-postgres-server/pom.xml
+++ b/legend-engine-xt-sql-postgres-server/pom.xml
@@ -204,6 +204,11 @@
             <artifactId>logback-core</artifactId>
             <scope>runtime</scope>
         </dependency>
+        <dependency>
+            <groupId>io.prometheus.jmx</groupId>
+            <artifactId>jmx_prometheus_javaagent</artifactId>
+            <scope>runtime</scope>
+        </dependency>
 
         <dependency>
             <groupId>javax.ws.rs</groupId>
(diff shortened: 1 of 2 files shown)


gs-manvig:
Looks good to me

epsstan (changes requested):
Let's take a look at how engine and the other server components expose metrics. 

I don't think they use JMX. They use a Prometheus SDK to create metrics which are then exposed via a HTTP metrics servlet/endpoint.

Can we use the same approach here ?

shtirlets:
@epsstan we are looking to expose the JVM/Environment metrics. My understanding we are getting for free Dropwizard in other use cases.

aormerod-gs (approved):

aormerod-gs:
The intention of this change is for exposing the JVM platform metrics (that are already in JMX) in to Prometheus (rather than exposing application metrics in to JMX)

epsstan:
> The intention of this change is for exposing the JVM platform metrics (that are already in JMX) in to Prometheus (rather than exposing application metrics in to JMX)

Aah ok. My bad. I didn't read the PR carefully.  So the java agent scrapes JMX and exposes it via a HTTP server.

epsstan (approved):

PR #1897, opened 2023-06-07
add custom deserializer for modelStoreData

MauricioUyaguari:

Fixes #


legend-engine-protocol-pure/src/main/java/org/finos/legend/engine/protocol/pure/v1/model/data/ModelStoreData.java
```diff
@@ -14,11 +14,79 @@
...
+import java.io.IOException;
+import java.util.Iterator;
 import java.util.List;
+import java.util.Map;
 
+@JsonDeserialize(using = ModelStoreData.ModelStoreDeserializer.class)
 public class ModelStoreData extends EmbeddedData
 {
 
+    private static ObjectMapper objectMapper = PureProtocolObjectMapperFactory.getNewObjectMapper();
```
hardikmaheshwari, line 38: ModelStoreData is a protocol class, right? We should move object mapper to nested class rather

PR #1910, opened 2023-06-08
Handle empty queries

gs-manvig:
#### What type of PR is this?
Bug Fix

Add support for empty queries

`Connection.isValid` in Postgres JDBC driver invokes an empty query and expects an empty response.
This PR forwards the empty query to the empty session handler


shtirlets:
looks good

aormerod-gs (approved):

PR #1916, opened 2023-06-09
Add support for SHOW TRANSACTION queries

gs-manvig:
#### What type of PR is this?
Improvement

Add support for `SHOW TRANSACTION` queries.
These queries are executed during the pool creation of HikariCP to validate the connection.

`SHOW TRANSACTION` queries are temporary routed to metadata handler until it can be supported in Alloy SQL
No


shtirlets:
Looks good

PR #1923, opened 2023-06-12
Add ability to generate OpenAPI spec from PureModelContext

kranthikiran01:
#### What type of PR is this?

Improvement
Current openapi generation uses file generation to create openapi specification for legend service(s) inside a studio project. This file generation is invoked during builds. We need a feature to generate openapi specification on demand when pure model context is supplied. This is useful in various cases one of which is showing openapi on studio.

OpenAPIGenerationService endpoint needs to be added which addresses the above requirement.


AFine-gs:
Please take a look at the  /pure/v1/external/format/generateSchema  . You may need to extend it to work with this use case.

MauricioUyaguari:
@kranthikiran01 we shouldnt be doing this. The OpenAPI falls into the artifact generation api. We have the api already in engine. we should do the work to expose this in studio. Lets discuss this.

finos-admin:

finos-admin:

PR #1929, opened 2023-06-13
Fixed parsing of TDS results with null values

shtirlets:
#### What type of PR is this?

Improvement

- Enhanced to extract correct data type from JSON response, previously always retrieved text
- Add handling for null values
- Add parsing unit tests for strings, enums and nulls


legend-engine-xt-sql-postgres-server/src/main/java/org/finos/legend/engine/postgres/handler/legend/LegendResultSet.java
```diff
@@ -80,13 +84,13 @@ public Object getObject(int i) throws Exception
                     return ((LocalDate) temporalAccessor).atStartOfDay().toInstant(ZoneOffset.UTC).toEpochMilli();
                 }
             case "Integer":
```
aormerod-gs, line 86: Can we declare some constants for these type?

aormerod-gs (approved):

PR #1946, opened 2023-06-20
fix schema exploration endpoint when pattern is empty

akphi:
fix an issue with schema exploration endpoint when the pattern is empty, the generated database would be empty


MauricioUyaguari:
can we add tests for wildcards inputs ?

PR #1957, opened 2023-06-22
GraphQL - adding memorization to types traversal

abhishoya-gs:
#### What type of PR is this?
Improvement

For large models our current algorithm takes a lot of time, traversing each path. Refactoring it to memorize past results.

Fixes #

Test fails currently as seen here https://github.com/finos/legend-engine/pull/1965 due to timeout

### DIFF
--- a/legend-engine-xt-graphQL-pure/src/main/resources/core_external_query_graphql/binding/fromPure/introspection/fromPure_Introspection.pure
+++ b/legend-engine-xt-graphQL-pure/src/main/resources/core_external_query_graphql/binding/fromPure/introspection/fromPure_Introspection.pure
@@ -5,10 +5,10 @@ import meta::external::query::graphQL::binding::fromPure::introspection::*;
 
 function meta::external::query::graphQL::binding::fromPure::introspection::scanTypes(t:Type[1], processed:Type[*]):Type[*]
 {
-  let np = if(!$processed->contains($t), | $processed->concatenate($t), | $processed);
-  if (!$processed->contains($t) && $t->instanceOf(Class),
-      | 
-        let propTypes = $t->cast(@Class<Any>)
+  let processedContainsType = $processed->contains($t);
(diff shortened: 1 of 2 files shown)


legend-engine-xt-graphQL-pure/src/main/resources/core_external_query_graphql/binding/fromPure/introspection/fromPure_Introspection.pure
```diff
@@ -5,13 +5,24 @@ import meta::external::query::graphQL::binding::fromPure::introspection::*;
...
 {
-   if (!$processed->contains($t),
-       |$t->concatenate(if($t->instanceOf(Class),
-                         |$t->cast(@Class<Any>)->allProperties().genericType.rawType->distinct()->concatenate($t->cast(@Class<Any>)->meta::pure::functions::meta::findAllSpecializations())->filter(t|($t->instanceOf(Class) || $t->instanceOf(meta::pure::metamodel::type::Enumeration)) && $t != Any)->map(c|$c->meta::external::query::graphQL::binding::fromPure::introspection::scanTypes($processed->concatenate($t)))->distinct(),
-                         |[]
-                       )),
-       |[]
-   );
+  let np = if(!$processed->contains($t), | $processed->concatenate($t), | $processed);
+  if (!$processed->contains($t) && $t->instanceOf(Class),
```
aziemchawdhary-gs, line 9: Would swapping the conjuncts in this condition make this more efficient? If you check the instance of $t first before doing the contains check you avoid having to go through processed if you don't need to. Might make a difference if the size of $processed gets very large.

aziemchawdhary-gs: yes doing the contains only once is better

epsstan (approved):

PR #1961, opened 2023-06-22
Introducing paginate function

abhishoya-gs:
#### What type of PR is this?
Improvement

Introduces paginate function with extended capability to return boolean flag 'hasMore' in JSON results, set to true when there are more pages.

Fixes #


rafaelbey:
Im not sure that we need a function like this, or if we do, this seems something that need further discussion.

Among concerns:

- overloading the from/to on the pure to sql
- json contract wont be extensible, and even further, could break some client that were not aware of this pagination 
- Assume pagination a core feature of graph fetch 

Also, can we expand on the use case here?  I think a user can already paginate thru the existing functionality (`sort()->drop()->take()`) and if no result are returned, no more result.  Is not necessarily the most efficient, but could be a immediate solution while we ensure the system support pagination cleanly and clearly.

PR #2014, opened 2023-07-10
Introduce `in` dynaFunc snowflake specific behavior

gs-manvig:
#### What type of PR is this?
Bug Fix

`In` clause without brackets along with an `is not null` currently fails in Snowflake due to SQL Compilation error. Adding brackets around `in` clause fixes the queries.

Previous SQL that fails: 
```sql
select
  case
    when "root"."COL" in ('val1', 'val2') is not null then 'true'
    else 'false'
  end as "(derivation)"
from
  table as "root"
```

New SQL introduced by this change
```sql
select
  case
    when ("root"."COL" in ('val1', 'val2')) is not null then 'true'
    else 'false'
  end as "(derivation)"
from
  table as "root"
```
No

### DIFF
--- a/.github/workflows/database-trino-integration-test.yml
+++ b/.github/workflows/database-trino-integration-test.yml
@@ -68,4 +68,4 @@ jobs:
       - name: Run Connection Protocol Integration Tests
         env:
           MAVEN_OPTS: -Xmx4g
-        run: mvn -pl legend-engine-xt-relationalStore-trino-execution-tests  -Dtest="ExternalIntegration*Trino*" test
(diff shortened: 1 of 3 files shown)


rafaelbey:
Does this query even make sense tho? `when "root"."COL" in ('val1', 'val2') is not null then 'true'`

If I got this right, this is always 'true' regardless of the in clause, since `"root"."COL" in ('val1', 'val2')` always returns a boolean, and then the `is not null` always will be true.

Do we have some integration test to confirm proper behavior?   I suspect the problem is not the parenthesis but the is not null...

PR #2023, opened 2023-07-11
Change scope of dependencies to better support this test module on ID…

rafaelbey:
…E light

#### What type of PR is this?

Bug Fix

When this module is imported on IDE Light, it should provide all its dependencies.

Given the module itself is a 'test' module and one module others should not depend on, changing the scope does not impact anything else.

Fixes #


finos-admin:

PR #2041, opened 2023-07-18
Update loggers to use class name

gs-jp1:
#### What type of PR is this?
Improvement

Uses class name in logger.

No


rafaelbey:
What is the rational for this?  

While I think make sense, we should check with SRE, as I think this single logger is assume on some logging scraping scripts.

gs-jp1:
> What is the rational for this?
> 
> While I think make sense, we should check with SRE, as I think this single logger is assume on some logging scraping scripts.

it was exactly this SRE scraping/analysis that lead us to making this change.

aormerod-gs (approved):

PR #2052, opened 2023-07-20
SQL integraton test, mvn path fixes

gs-jp1:
#### What type of PR is this?
Bug Fix

updating mvn paths

### DIFF
--- a/.github/workflows/database-athena-integration-test.yml
+++ b/.github/workflows/database-athena-integration-test.yml
@@ -72,4 +72,4 @@ jobs:
           MAVEN_OPTS: -Xmx4g
           AWS_ACCESS_KEY_ID: ${{ secrets.AWS_INTEGRATION_USER1_ACCESS_KEY_ID }}
           AWS_SECRET_ACCESS_KEY: ${{ secrets.AWS_INTEGRATION_USER1_SECRET_ACCESS_KEY }}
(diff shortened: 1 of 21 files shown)


aziemchawdhary-gs (approved):

.github/workflows/database-athena-integration-test.yml
```diff
@@ -72,4 +72,4 @@ jobs:
           MAVEN_OPTS: -Xmx4g
           AWS_ACCESS_KEY_ID: ${{ secrets.AWS_INTEGRATION_USER1_ACCESS_KEY_ID }}
           AWS_SECRET_ACCESS_KEY: ${{ secrets.AWS_INTEGRATION_USER1_SECRET_ACCESS_KEY }}
-        run: mvn -pl legend-engine-xt-relationalStore-athena-execution-tests -Dtest=ExternalIntegration*Athena test
+        run: mvn -pl legend-engine-xts-relationalStore/legend-engine-xt-relationalStore-dbExtension/legend-engine-xt-relationalStore-athena/legend-engine-xt-relationalStore-athena-execution-tests -Dtest=ExternalIntegration*Athena test
```
gs-manvig, line 75: Would be less verbose to use module names like so `mvn -pl :legend-engine-xt-relationalStore-athena-execution-tests` instead of having to use directory names

PR #2057, opened 2023-07-21
Legend SQL - Update logging

gs-manvig:
#### What type of PR is this?
Improvement

- Rename logger name to use Class name
- Bump netty version
- Use native netty handler when available 
- Add client IP to MDC
- Add username to MDC

- Previous addition of username to MDC only occurred for non-Kerberos modes of authentication. Proposed implementation moves it up so username is included for all modes of authentication
- Client IP is added on connection establishment (i.e. pre-authentication)
No


shtirlets:
Looks good

legend-engine-xts-sql/legend-engine-xt-sql-postgres-server/src/main/java/org/finos/legend/engine/postgres/PostgresWireProtocol.java
```diff
@@ -574,6 +575,7 @@ private void handleAuthSuccess(Channel channel, Identity authenticatedUser) thro
     {
         String database = properties.getProperty("database");
         session = sessions.createSession(database, authenticatedUser);
+        MDC.put("user", authenticatedUser.getName());
```
aormerod-gs, line 578: should we put a session ID in here as well?

aormerod-gs (approved):

PR #2063, opened 2023-07-24
Introduce `paginated` API

abhishoya-gs:
#### What type of PR is this?
Improvement - Feature

Introducing `paginated` API which will serve as an alternate to users currently using slice, limit, drop for pagination. 

`paginated<T>(set:T[*], pageNumber: Integer[1], pageSize: Integer[1]): T[*]`

It returns the n-th (n = pageNumber) page for `T` after paginating as per pageSize.

Additionally the relational implementation mandates presence of a sort function being invoked before the API. 

Fixes #

Yes


AFine-gs:
Should the sort function be a parameter of pagination ?   This would always ensure that a sort is specified

gs-ssh16 (approved):

PR #2080, opened 2023-07-27
XStore Association can be used for localized project queries

haroonsherjan:
#### What type of PR is this?
Improvement

Users can create a XStore association mapping against an included mapping without including the matching store in their own store. This will still generate the appropriate SQL even if the physical join has not been defined. Works for joins that leverage relational mapped class attributes on both sides of equality. Not currently for equality that depends upon a relational class attribute against a static value. (ex. $this.name = 'personName' && $this.id == $that.ID_FK)


finos-admin:

finos-admin:

PR #2088, opened 2023-07-28
Test Data Generation I -  support a new API to generate skeleton schema/table from query

YannanGao-gs:
#### What type of PR is this?

Improvement

Test Data Generation I - support a new API to generate skeleton schema/table from query

Fixes #

### DIFF
--- a/legend-engine-test-data-generation/src/main/java/org/finos/legend/engine/testData/generation/model/TestDataGenerationInput.java
+++ b/legend-engine-test-data-generation/src/main/java/org/finos/legend/engine/testData/generation/model/TestDataGenerationInput.java
@@ -1,0 +1,51 @@
+/**
+ * Copyright (c) 2020-present, Goldman Sachs
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package org.finos.legend.engine.testData.generation.model;
+
+import com.fasterxml.jackson.annotation.JsonProperty;
+import org.finos.legend.engine.protocol.pure.v1.model.context.PureModelContext;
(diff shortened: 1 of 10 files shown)


legend-engine-test-data-generation/src/main/java/org/finos/legend/engine/testData/generation/service/TestDataGenerationService.java
```diff
@@ -0,0 +1,63 @@
...
+import org.finos.legend.pure.m3.coreinstance.meta.pure.metamodel.function.LambdaFunction;
+import org.finos.legend.pure.m3.coreinstance.meta.relational.metamodel.Column;
+import org.finos.legend.pure.m3.coreinstance.meta.relational.metamodel.relation.Table;
+
+import java.util.Collections;
+import java.util.List;
+
+public class TestDataGenerationService
+{
+    public static List<EmbeddedData> generateEmbeddedData(Lambda query, Root_meta_pure_runtime_Runtime runtime, Mapping mapping, PureModel pureModel)
```
MauricioUyaguari, line 38: can we return a pmcd instead to make it flow better with other generation flows. we can read the embedded data on the studio side.

legend-engine-test-data-generation/src/main/java/org/finos/legend/engine/testData/generation/model/TestDataGenerationInput.java
```diff
@@ -0,0 +1,48 @@
...
+package org.finos.legend.engine.testData.generation.model;
+
+import com.fasterxml.jackson.annotation.JsonProperty;
+import org.finos.legend.engine.protocol.pure.v1.model.context.PureModelContext;
+import org.finos.legend.engine.protocol.pure.v1.model.valueSpecification.raw.Lambda;
+
+public class TestDataGenerationInput
+{
+    @JsonProperty
+    public Lambda query;
```
MauricioUyaguari, line 26: lets add a comment that query should not be required but for now we will require it.

MauricioUyaguari (changes requested):
minor comments.

legend-engine-test-data-generation/src/main/java/org/finos/legend/engine/testData/generation/service/TestDataGenerationService.java
YannanGao-gs: Hello, would you mind being more specific about returning a pmcd?

MauricioUyaguari: we should be returning a pmcd in the api.

MauricioUyaguari (approved):

PR #2125, opened 2023-08-09
add tests for the API which is to generate Skeleton schema/table

YannanGao-gs:
#### What type of PR is this?

Improvement 

add tests for the API which is to generate Skeleton schema/table

Fixes #

### DIFF
--- a/legend-engine-test-data-generation/src/test/resources/models/relationalModelTestDataGenerationInput.json
+++ b/legend-engine-test-data-generation/src/test/resources/models/relationalModelTestDataGenerationInput.json
@@ -1,0 +1,641 @@
+{
+  "query": {
+    "_type": "lambda",
+    "body": [
+      {
+        "_type": "func",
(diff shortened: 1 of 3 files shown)


legend-engine-test-data-generation/src/test/resources/models/relationalModelTestDataGenerationInput.json
```diff
@@ -0,0 +1,641 @@
+{
```
MauricioUyaguari, line 1: Can we change to use pure grammar in the test so we can test and can we also just keep only the required elements for the test. thanks!

YannanGao-gs: The reason why i used the input json is that query is needed for now. 
If I just used the  pure grammar, it's hard to get query.

MauricioUyaguari (approved):

PR #2136, opened 2023-08-11
feat: backend support for code fix suggestions

mihir1021:
#### What type of PR is this?

Improvement

During code composition, if an element's type cannot be found, then the user is prompted with a code fix suggestion in Studio. In order to support the Studio functionality (described in this [Studio PR](https://github.com/finos/legend-studio/pull/2501) and this [Studio PR](https://github.com/finos/legend-studio/pull/2502)), Engine must return a set of candidates.

This PR creates a new type of exception that extends and EngineException, and it contains a new class variable that contains the candidates that will help users resolve the can't find element error in Studio. This list is populated during composition by the Pure Model class and eventually returned to Studio through the compile API.

Fixes #

When changes are released, they should be sent out along with one of the Studio PRs hyperlinked in the summary above.


finos-admin:

PR #2190, opened 2023-08-27
Update TestExtensions with new db extensions

gs-kotang:
#### What type of PR is this?

- Improvement

Improve existing tests. Extensions-collection modules to have access to all database extensions

Fixes #

No


SaiSriharsha:
Need to add pure modules as runtime deps for those DBs which do not have grammar modules.

Also, collections-execution module should be updated as well

gs-ssh16:
Need to add -pure modules as runtime deps for those DBs which do not have grammar modules. Basically, we need to see all new repos.

Also, collections-execution module should be updated with those DBs which have execution modules

gs-kotang:
> Need to add -pure modules as runtime deps for those DBs which do not have grammar modules. Basically, we need to see all new repos.
> 
> Also, collections-execution module should be updated with those DBs which have execution modules

Added both

PR #2196, opened 2023-08-29
Ignore selectSubClauses and namingTests packages for BigQuery tests

abhishoya-gs:
#### What type of PR is this?
Ignore selectSubClauses and namingTests packages for BigQuery tests

**What:** Ignore test suites explicitly.

**Why:** In pure IDE light when we run these tests they fail as setup function is called before each test. We have setup checking on DDL support in BQ and fails with unsupported api. However in java run, setup is added as first test case with assumption that it will pass. Though it fails it doesn't stop suite from running. So explicitly ignoring tests for BQ where setup fails.

Fixes #

No


finos-admin:

finos-admin:

PR #2213, opened 2023-08-31
Prepend suite name to test name in Relational_DbSpecific

abhishoya-gs:
#### What type of PR is this?
Bug Fix

- While running Db Specific our main test suite contains test suites nested inside at different levels
- These nested test suites have duplicate test names
- Surefire while generating report flattens out the structure and only takes test name
- While reading these reports during SQL Summary Generator, we end up overwriting no duplicate test names

This change will make names unique by prepending suite name to them.

example - 
| when | name|
|--|--|
| before | testNotEndsWith |
| after | dynaFunctions::endsWith::testNotEndsWith |

Fixes #

https://github.com/abhishoya-gs/legend-engine/actions/runs/6084838006

No


hardikmaheshwari:
Thanks @abhishoya-gs for writing a good description

hardikmaheshwari (approved):

PR #2252, opened 2023-09-11
Improve mapping coverage analysis algo to support analysis for complex auto mapped

YannanGao-gs:
#### What type of PR is this?

Improvement

Improve mapping coverage analysis algo to support analysis for binding queries

after: 

https://github.com/finos/legend-engine/assets/73408381/6b9a59b0-e595-4fb8-b565-987cace85440

Fixes #

### DIFF
--- a/legend-engine-xts-analytics/legend-engine-xts-analytics-mapping/legend-engine-xt-analytics-mapping-pure/src/main/resources/core_analytics_mapping/modelCoverage/mappedEntityBuilder.pure
+++ b/legend-engine-xts-analytics/legend-engine-xts-analytics-mapping/legend-engine-xt-analytics-mapping-pure/src/main/resources/core_analytics_mapping/modelCoverage/mappedEntityBuilder.pure
@@ -12,6 +12,7 @@
 // See the License for the specific language governing permissions and
 // limitations under the License.
 
+import meta::relational::mapping::*;
 import meta::analytics::mapping::modelCoverage::*;
 import meta::pure::mapping::*;
 import meta::pure::mapping::modelToModel::*;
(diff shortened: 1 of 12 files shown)


legend-engine-xts-analytics/legend-engine-xts-analytics-mapping/legend-engine-xt-analytics-mapping-pure/src/main/resources/core_analytics_mapping/modelCoverage/mappedEntityBuilder.pure
```diff
@@ -281,18 +290,47 @@ function meta::analytics::mapping::modelCoverage::buildEntity(
          ); //TODO check how types work with qp's.we n eed to filter on whether the type is root supported currently
    );
 
-   let autoMappedProperties = if($setImplementation->instanceOf(PureInstanceSetImplementation),
+   // handle auto mapped properties
```
gs-jp1, line 293: this naturally feels wrong. I would expect the binding to be exposed in the classMappings as every other target type is. This way it would be a very standard class mapping handler on the SemiStructuredEmbeddedRelationalInstanceSetImplementation and these functions keep their initial contract

YannanGao-gs: Hello John, thanks for reviewing this PR.  SemiStructuredEmbeddedRelationalInstanceSetImplementation needs to be handled as an auto-mapped mapping due to this newly added commit
https://github.com/finos/legend-engine/pull/2141

YannanGao-gs:
Found an ongoing issue:  forget to handle the case where there might be a cycle in the model

MauricioUyaguari (approved):

PR #2269, opened 2023-09-14
Fix `is not distinct from` sql-to-pure

gs-manvig:
Remove redundant `isTrue` for an `or` expression

#### What type of PR is this?
Bug Fix

The generated sql for `is not distinct from` fails to compile snowflake 
Issue does not occur when the redundant `isTrue` is removed from the pure expression for `is not distinct from` 

No


legend-engine-xts-sql/legend-engine-xt-sql-pure/src/main/resources/core_external_query_sql/binding/fromPure/fromPure.pure
```diff
@@ -2210,7 +2210,7 @@ function <<access.private>> meta::external::query::sql::transformation::queryToP
       sfe(isEmpty_Any_$0_1$__Boolean_1_, $left),
       sfe(isEmpty_Any_$0_1$__Boolean_1_, $right)
     ]),
-    sfe(isTrue_Boolean_$0_1$__Boolean_1_, sfe(equal_Any_MANY__Any_MANY__Boolean_1_, [$left, $right]))
```
gs-jp1, line 2213: looks good to me, thanks

aormerod-gs (approved):

PR #2301, opened 2023-09-23
Escape single quote present in values for temp table

abhishoya-gs:
#### What type of PR is this?
Bug Fix

#### What does this PR do / why is it needed?
Escape single quote present in values for temp table by replacing with two single quotes. 

Fixes #

From https://docs.snowflake.com/en/sql-reference/data-types-text#single-quoted-string-constants

`To include a single quote character within a string constant, type two adjacent single quotes (e.g. '').`

yes


gs-kotang (commented):
Lgtm. This only affects snowflake executions. We should follow up to see if we can have the ResultNormalizer handle this and test it more broadly.

abhishoya-gs:
> Lgtm. This only affects snowflake executions. We should follow up to see if we can have the ResultNormalizer handle this and test it more broadly.

Will do

PR #2336, opened 2023-10-02
Update Snowflake App Generation Extension

sprisha:
#### What type of PR is this?

 Adding artifact generation extension for snowflakeApp

 Needed for Function Activator deployment efforts 

Fixes #

 No

### DIFF
--- a/legend-engine-xts-snowflakeApp/legend-engine-xt-snowflakeApp-api/src/main/java/org/finos/legend/engine/language/snowflakeApp/deployment/SnowflakeAppArtifactGenerationExtension.java
+++ b/legend-engine-xts-snowflakeApp/legend-engine-xt-snowflakeApp-api/src/main/java/org/finos/legend/engine/language/snowflakeApp/deployment/SnowflakeAppArtifactGenerationExtension.java
@@ -27,7 +27,7 @@ import java.util.List;
 
 public class SnowflakeAppArtifactGenerationExtension implements ArtifactGenerationExtension
 {
-    private static final String ROOT_PATH = "Snowflake App ";
+    private static final String ROOT_PATH = "snowflakeApp";
     private static final Logger LOGGER = org.slf4j.LoggerFactory.getLogger(SnowflakeAppArtifactGenerationExtension.class);
 
     @Override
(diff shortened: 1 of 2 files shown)


legend-engine-xts-snowflakeApp/legend-engine-xt-snowflakeApp-api/src/main/java/org/finos/legend/engine/language/snowflakeApp/deployment/SnowflakeAppArtifactGenerationExtension.java
```diff
@@ -21,12 +21,14 @@
...
 import org.finos.legend.pure.m3.coreinstance.meta.pure.metamodel.PackageableElement;
-
+import org.slf4j.Logger;
+import java.util.Collections;
 import java.util.List;
 
 public class SnowflakeAppArtifactGenerationExtension implements ArtifactGenerationExtension
 {
     private static final String ROOT_PATH = "Snowflake App ";
+    private static final Logger LOGGER = org.slf4j.LoggerFactory.getLogger(SnowflakeAppArtifactGenerationExtension.class);
```
MauricioUyaguari, line 31: @sprisha 
the root path for your artifact generation should be a valid file path. Therefore a space would not be valid, please fix.
private static final String ROOT_PATH = "Snowflake App ";

MauricioUyaguari (changes requested):

legend-engine-xts-snowflakeApp/legend-engine-xt-snowflakeApp-api/src/main/java/org/finos/legend/engine/language/snowflakeApp/deployment/SnowflakeAppArtifactGenerationExtension.java
sprisha: got it, thanks

legend-engine-xts-snowflakeApp/legend-engine-xt-snowflakeApp-api/src/main/java/org/finos/legend/engine/language/snowflakeApp/deployment/SnowflakeAppArtifactGenerationExtension.java
```diff
@@ -38,15 +40,25 @@ public String getKey()
...
 
 
     @Override
     public List<Artifact> generate(PackageableElement element, PureModel pureModel, PureModelContextData data, String clientVersion)
     {
-        return null;
+       try
+        {
+            /* add content logic here */
+            Artifact output = new Artifact(null, element.getName() + "_snowflakeAppArtifact.json", "json");
```
rafaelbey, line 52: My understanding is that each artifact is already generated inside a directory for the element.

Hence, I don't thin we should do this concatenation of element name with the suffix, as it will make the name more complex it needs to be.

MauricioUyaguari (changes requested):
Don't really get the point of this change. We are returning an artifact with no content `new Artifact(null, element.getName() + "_snowflakeAppArtifact.json", "json")`.
In the test all we are doing is checking if the outputs has one artifact?

PR #2348, opened 2023-10-06
Normalize with timezone

PrateekGarg-gs:

Fixes #


aziemchawdhary-gs:
Can we add tests for uses with and without the flag? You should  be able to add tests to TestResultNormalizer to test both flows

aziemchawdhary-gs:
Change not needed - timezone was not the issue in date formatting

PR #2360, opened 2023-10-11
Compare JSON Values for Integral vs Non-Integral

mrudula-gs:
#### What type of PR is this?
Bug Fix

On service test execution when we compare the JSON, the assert fails when there is an Integral Node and a Non-Integral Node even though the actual node value is exactly same (with no decimal value on the Non-Integral node).
The changeset is to also to compare the actual value and not just node types.

We see this issue for values with exponent notation that round of to proper Integral numbers and this is because we are enabling the Deserialization Feature USE_BIG_DECIMAL_FOR_FLOATS.
We can enable the ACCEPT_FLOAT_AS_INT but USE_BIG_DECIMAL_FOR_FLOATS takes precedence over USE_BIG_DECIMAL_FOR_FLOATS and so its of no use.

One approach is to preProcess before Jackson handles but we decided not to do that as it could significantly increase service test execution time.

### DIFF
--- a/legend-engine-core/legend-engine-core-test/legend-engine-test-runner-shared/src/main/java/org/finos/legend/engine/test/runner/shared/JsonNodeComparator.java
+++ b/legend-engine-core/legend-engine-core-test/legend-engine-test-runner-shared/src/main/java/org/finos/legend/engine/test/runner/shared/JsonNodeComparator.java
@@ -150,7 +150,7 @@ public class JsonNodeComparator implements Comparator<JsonNode>
     {
         if (node1.isIntegralNumber())
         {
-            if (!node2.isIntegralNumber())
+            if (!node2.isIntegralNumber() && !(node1.asText().equals(node2.asText())))
             {
                 return -1;
             }
@@ -166,7 +166,7 @@ public class JsonNodeComparator implements Comparator<JsonNode>
         }
         if (node1.isFloatingPointNumber())
         {
-            if (!node2.isFloatingPointNumber())
+            if (!node2.isFloatingPointNumber() && !(node1.asText().equals(node2.asText())))
             {
                 return -1;
             }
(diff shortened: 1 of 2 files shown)


rafaelbey:
Are gonna allow 1.0 to be equal to 1?  I think this would highlight some other problems, either on the model or on the store...

Still, if that is the case, wont be better to just compare the values as double always rather than as text?

mrudula-gs:
1.0 is not considered as equal to 1 as of now, and this change is not going to break that. That is the case where one is Integral node and one is non-integral node with different text values so there are evaluated as not equal.

rafaelbey:
Can you provide a more concreate  example on what was not working and now will work?  That whole comparison logic seems too complex already, and seems we are adding more complexity, and the test case does not seem to provide clear evidence on what is that we are fixing either.

aziemchawdhary-gs:
This is fixing an edge case involving exponent values. For example, suppose we have a class with type Float[1]. 
Test input data provides a value 6.2292853E7 and suppose we just do a straight mapping of the value from source to target in the mapping/query.

Jackson is set to with WRITE_BIGDECIMAL_AS_PLAIN in legend-engine, so when we serialize/deserialize such values we end up with 62292853.
As a result when writing tests the user has an expected value of 62292853. We parse the expected test data using Jackson which then stores this in an IntNode with value 62292853.
However, the execution of the mapping/query knows that the model has type Float so the execution puts 62292853 into a DecimalNode.

The current JsonNodeComparator therefore ends up comparing a IntNode and DecimalMode which both have the value 62292853 and currently says they are not equivalent due to the type of the node (ignoring the value itself)
This change is handling the above case but *not* switching the entire comparator to mathematical equality (e.g. 1 == 1.0)

rafaelbey:
That make sense. Seems as we should care less about the node type, and more about the node content.  

Like, in that case, should we compare these first as big decimal regardless if the node type?

`node.decimalValue()` works for numbers (double, float, integer, decimal, etc), and in that scenario, both should return a decimal with scale 0.

Thoughts?

mrudula-gs:
I tried this out Rafael, the check node2.isDecimal() && !node2.canConvertToLong() 
This work for the exponent case IntNode(62292853) and DecimalNode(62292853)
But goes False for IntNode(1) and DecimalNode(1.0) and so it won't fail the precision comparison

davidharte-gs (approved):

PR #2379, opened 2023-10-17
Publish to sandbox only support for BigQuery Function Activator

horbe:
Basic support for BigQuery Function Activator


Yasirmod17 (commented):
Currently reviewing

PR #2389, opened 2023-10-19
Allow parsing sql expressions

rafaelbey:
#### What type of PR is this?

Improvement

Allow SQL parser to just parse expressions.  This will allow grammars to contain partial sql expressions like where clauses
 
Fixes #


gs-jp1:
looks good to me

akphi (approved):

PR #2437, opened 2023-11-04
Enhance model manager to support fetching pmcd from workspaces

rafaelbey:
#### What type of PR is this?

Improvement

Enhance model manager to support loading PMCD from Legend SDLC workspaces.  

This reduce duplication of code while allow multiple existing core APIs to be able to operate on PMCDs from workspaces, rather than bespoke places like GraphQL or Legend SQL.

Fixes #


akphi (dismissed):

akphi:
@rafaelbey Nice!

aormerod-gs (dismissed):

epsstan (approved):

PR #2461, opened 2023-11-15
Translate XStore into Relational based on locality

haroonsherjan:
#### What type of PR is this?
Feature

Enable model level joins. Users no longer need to define relational joins and include producer stores.


finos-admin:

PR #2368, opened 2023-10-13
data-push: POC server

akphi:
- [x] Refactor the models
- [x] Add basic support for data-push server (to s3->bulk-copy (bcp) to snowflake) using the new connection framework
- [x] Create demo flow for connection framework: `SnowflakeWithKeyPair`
- [x] Document the design/architecture/implementation and migration strategy of the new framework - preview at https://github.com/akphi/legend-engine/blob/psh/docs/connection/new-connection-framework.md - @pierredebelen @epsstan 

## NOTE

This PR only includes the model change and the push-server glue code, the migration change will be done in a different PR, but this is the rough demo:

> A Snowflake connection using KeyPair is automatically converted/deserialized to use the new connection models and therefore connection is acquired using the new framework

https://github.com/finos/legend-engine/assets/13574879/ce953d76-d68f-4cb1-890f-b424159327e1

PR #2464, opened 2023-11-16
connection manager: migration-strategy demo

akphi:
This is a continuation of #2368 

See detailed documentation at https://github.com/finos/legend-engine/blob/master/docs/connection/new-connection-framework.md#migration-strategy

To test out the changes: set the environment variable `org.finos.legend.engine.enableNewConnectionFramework=true` when starting the server. To execute, since at the time of speaking, we haven't included the new connection models in the latest production protocol, choose `clientVersion=vX_X_X`

### Demo
A Snowflake connection using KeyPair is automatically converted/deserialized to use the new connection models, and therefore, the connection is acquired using the new framework.

https://github.com/finos/legend-engine/assets/13574879/ce953d76-d68f-4cb1-890f-b424159327e1


legend-engine-xts-connection/legend-engine-xt-connection-factory/src/main/java/org/finos/legend/connection/ConnectionExtension.java
```diff
@@ -28,4 +29,9 @@ default List<AuthenticationMechanismType> getExtraAuthenticationMechanismTypes()
     {
         return Collections.emptyList();
     }
+
+    default Map<String, DatabaseType> getExtraDatabaseTypeMapping()
```
rafaelbey, line 33: This is leaking relational concepts into the core.  This is an area we need to address long term, hence not sure we want to add new functionality that assume relational is part of core.  

For example - we should avoid a single enum for all db types since that conflicts with the modularization of the system.  

Anyway this can be part of the relational extension?

akphi: @rafaelbey this is actually `DatabaseType` here is from `org.finos.legend.connection`, not the one from `storeRelational`, and it's designed with extensibility in mind: if you look at `RelationalDatabaseType`, that's an enum that extends this interface, hence, demonstrating we can further modularize this if we want into `SnowflakeRelationalDBType`, `BigQueryRelationalDBType`, etc.

I think perhaps I run out of names, but `database` to me applies to more than just relational DB, so when I name it `DatabaseType`, I envision we can do `DatabaseType.Mongo`, `DatabaseType.ElasticSearch` (also see `DatabaseSupport`, that's generic for all kinds of stores)

Thanks for reviewing :), but please check with @epsstan and @pierredebelen when/if we want to merge this.

finos-admin:

akphi:
Closing as this approach will be undone by https://github.com/finos/legend-engine/pull/2506

PR #2470, opened 2023-11-22
Fix Boolean literal processor for MemSQL

gs-kotang:
#### What type of PR is this?

- Bug Fix

#### What does this PR do / why is it needed?

Override the default boolean literal processor in the MemSQL extension

Fixes #

testSqlTrue and testSqlFalse in Sql Generation tests of MemSQL should be green after this change

No


gs-kotang:
> Is this backward compatible? Have we identify any potential area where this can break? I can see people compensating on mappings, leading to failures with this change. (I still think is a good change, but we should identify any area we can break to communicate to users accordantly)

Makes sense. I think we are already aware of a few projects where people use `case(condition, 'true', 'false')` in their mappings. And it won't be backward compatible. I will start a conversation with them, and open this pr once we reach some conclusion.

finos-admin:

finos-admin:

PR #2476, opened 2023-11-24
otherwise embedded mapping support 

tanujgirdhar:
* no changes needed for relationalGraphFetch

* always route to otherwiseMapping for graphFetchFlow

* fix otherwise embedded mapping for graphfetch

* behave as if embedded mapping were empty

Fixes #

### DIFF
--- a/legend-engine-pure/legend-engine-pure-code/legend-engine-pure-code-compiled-core/src/main/resources/core/pure/graphFetch/graphFetch_routing.pure
+++ b/legend-engine-pure/legend-engine-pure-code/legend-engine-pure-code-compiled-core/src/main/resources/core/pure/graphFetch/graphFetch_routing.pure
@@ -12,6 +12,7 @@
 // See the License for the specific language governing permissions and
 // limitations under the License.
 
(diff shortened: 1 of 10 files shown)


legend-engine-pure/legend-engine-pure-code/legend-engine-pure-code-compiled-core/src/main/resources/core/pure/graphFetch/graphFetch_routing.pure
```diff
@@ -12,6 +12,7 @@
 // See the License for the specific language governing permissions and
 // limitations under the License.
 
+import meta::relational::mapping::*;
```
rafaelbey, line 15: How is this import used?  Can it be removed?

Seems that this is breaking encapsulation - core libraries depending on the implementation of relational.

tanujgirdhar: yes, this can be removed.

PR #2478, opened 2023-11-27
Legend SQL assortment of updates

gs-jp1:
#### What type of PR is this?
Improvement

- handle standard aggregation functions in window context
- bool_and/bool_or support
- extra tracing
- do not fail on duplicate sources
- explicit fail handling of function orderby


legend-engine-xts-relationalStore/legend-engine-xt-relationalStore-generation/legend-engine-xt-relationalStore-pure/src/main/resources/core_relational/relational/sqlQueryToString/extensionDefaults.pure
```diff
@@ -159,6 +159,13 @@ function meta::relational::functions::sqlQueryToString::default::convertGreatest
   )->joinStrings('case ', ' ', ' else null end'));
 }
 
+function meta::relational::functions::sqlQueryToString::default::convertBoolAndOrToCaseStatement(param:String[1], and:Boolean[1]):String[1]
+{
+  let fn = if ($and, | 'min', | 'max');
+
+  'case when ' + $fn + '(' + $param + ') = \'true\' then \'true\' else \'false\' end';
```
rafaelbey, line 166: hardcoding true/false as string seems wrong, and disregard the DB extension boolean support handling.

legend-engine-pure/legend-engine-pure-code/legend-engine-pure-code-compiled-core/src/main/resources/core/pure/corefunctions/collectionExtension.pure
```diff
@@ -123,6 +123,11 @@ function meta::pure::functions::collection::and(vals:Boolean[*]):Boolean[1]
    $vals->fold({i,a|$i && $a}, true)
 }
 
+function meta::pure::functions::collection::or(vals:Boolean[*]):Boolean[1]
+{
+   $vals->fold({i,a|$i || $a}, false)
```
rafaelbey, line 128: seems duplicated from the code below... can one delegate to the other?

legend-engine-xts-relationalStore/legend-engine-xt-relationalStore-generation/legend-engine-xt-relationalStore-pure/src/main/resources/core_relational/relational/sqlQueryToString/extensionDefaults.pure
gs-jp1: valid concern which was considered. We do not actually have access to that flag at the required point. This function is designed to be more a shared helper which can be parameterized accordingly for the dialect if required. (E.g min/max, comparison etc). We need true/false return to ensure correct type handling on resultset streaming.

legend-engine-pure/legend-engine-pure-code/legend-engine-pure-code-compiled-core/src/main/resources/core/pure/corefunctions/collectionExtension.pure
gs-jp1: a simple delegation will not work due to recursion, the alternative would be to introduce a different statement such as
if ($vals->isEmpty(), | false, | $values->toOneMany()->or()) which feels less optimal than this simple duplication.
The idea would be to remove the 1..* function but requires some further analysis first.

legend-engine-xts-relationalStore/legend-engine-xt-relationalStore-generation/legend-engine-xt-relationalStore-pure/src/main/resources/core_relational/relational/sqlQueryToString/extensionDefaults.pure
gs-jp1: 1. that would introduce a completely asymmetric flow to any other dynafunc transform. There are many considerably more complex dynafunc transformw than this which are done in a similar manner.
2. you do not have all the context at the time to do that
3. this would add significant complexity for which is fundamentally a simple transform.

This function is a simple shared helper using common pattern to be used where needed for the dialects. It can either be modified if needed in future to handle special cases or simply coded directly in the specific dialect handler

aormerod-gs (approved):

PR #2509, opened 2023-12-13
Relax multiplicity requirement

Yasirmod17:
Relax multiplicity requirement

Fixes #


Yasirmod17:
This is no longer required

PR #2547, opened 2024-01-16
Uplift Service Test framework to accommodate new Runtime DSL

haroonsherjan:
#### What type of PR is this?
Service Tests can accomodate new Runtime DSL

### DIFF
--- a/legend-engine-core/legend-engine-core-language-pure/legend-engine-protocol-pure/src/main/java/org/finos/legend/engine/protocol/pure/v1/extension/ConnectionFactoryExtension.java
+++ b/legend-engine-core/legend-engine-core-language-pure/legend-engine-protocol-pure/src/main/java/org/finos/legend/engine/protocol/pure/v1/extension/ConnectionFactoryExtension.java
@@ -51,7 +51,7 @@ public interface ConnectionFactoryExtension
         return Optional.empty();
     }
 
-    default Optional<Pair<Connection, List<Closeable>>> tryBuildTestConnection(Connection sourceConnection, EmbeddedData data)
+    default Optional<Pair<Connection, List<Closeable>>> tryBuildTestConnection(Connection sourceConnection, List<EmbeddedData> data)
     {
         return Optional.empty();
     }
--- a/legend-engine-core/legend-engine-core-test/legend-engine-testable/src/main/java/org/finos/legend/engine/testable/connection/TestConnectionBuilder.java
+++ b/legend-engine-core/legend-engine-core-test/legend-engine-testable/src/main/java/org/finos/legend/engine/testable/connection/TestConnectionBuilder.java
@@ -19,10 +19,10 @@ import org.eclipse.collections.api.tuple.Pair;
 import org.eclipse.collections.impl.factory.Lists;
 import org.eclipse.collections.impl.tuple.Tuples;
 import org.eclipse.collections.impl.utility.Iterate;
+import org.eclipse.collections.impl.utility.ListIterate;
 import org.finos.legend.engine.protocol.pure.v1.extension.ConnectionFactoryExtension;
 import org.finos.legend.engine.protocol.pure.v1.model.context.PureModelContextData;
 import org.finos.legend.engine.protocol.pure.v1.model.data.EmbeddedData;
-import org.finos.legend.engine.protocol.pure.v1.model.data.ExternalFormatData;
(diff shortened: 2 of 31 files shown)


legend-engine-core/legend-engine-core-language-pure/legend-engine-protocol-pure/src/main/java/org/finos/legend/engine/protocol/pure/v1/extension/ConnectionFactoryExtension.java
```diff
@@ -51,7 +51,7 @@ default Optional<Connection> tryBuildFromConnection(Connection connection, Strin
         return Optional.empty();
     }
 
-    default Optional<Pair<Connection, List<Closeable>>> tryBuildTestConnection(Connection sourceConnection, EmbeddedData data)
+    default Optional<Pair<Connection, List<Closeable>>> tryBuildTestConnection(Connection sourceConnection, List<EmbeddedData> data)
```
MauricioUyaguari, line 54: Let's not break the existing contract. Let's add another method maybe somethign like `tryBuildTestConnectionWithMultiData`
If the size of the list is 1, we should still default to `tryBuildTestConnection` as the defaullt implementation. 
This would prevent us from doing ->get(0) on the implementations

legend-engine-core/legend-engine-core-test/legend-engine-testable/src/main/java/org/finos/legend/engine/testable/connection/TestConnectionBuilder.java
```diff
@@ -67,7 +59,7 @@ public Pair<Connection, List<Closeable>> visit(Connection connection)
                 .getFirstOptional()
                 .orElseThrow(() ->
                 {
-                    String errorMessage = embeddedData == null
+                    String errorMessage = embeddedData.isEmpty()
```
MauricioUyaguari, line 62: `embeddedData == null || embeddedData.isEmpty()`

legend-engine-xts-service/legend-engine-test-runner-service/src/test/java/org/finos/legend/engine/test/runner/service/TestRuntimeGenerationForServiceTests.java
```diff
@@ -24,8 +24,10 @@
 import org.finos.legend.engine.protocol.pure.v1.model.packageableElement.store.relational.connection.RelationalDatabaseConnection;
 import org.finos.legend.engine.shared.core.deployment.DeploymentMode;
 import org.junit.Assert;
+import org.junit.Ignore;
 import org.junit.Test;
 
+@Ignore
```
MauricioUyaguari, line 30: why are we ignoring this ?

legend-engine-xts-service/legend-engine-test-runner-service/src/test/java/org/finos/legend/engine/test/runner/service/TestServiceTestRunner.java
```diff
@@ -33,6 +34,7 @@
 
 import static org.finos.legend.pure.generated.core_relational_java_platform_binding_legendJavaPlatformBinding_relationalLegendJavaPlatformBindingExtension.Root_meta_relational_executionPlan_platformBinding_legendJava_relationalExtensionsWithLegendJavaPlatformBinding__Extension_MANY_;
 
+@Ignore
```
MauricioUyaguari, line 37: ????

MauricioUyaguari (changes requested):

legend-engine-core/legend-engine-core-language-pure/legend-engine-protocol-pure/src/main/java/org/finos/legend/engine/protocol/pure/v1/extension/ConnectionFactoryExtension.java
haroonsherjan: I think we need to break the contract because we do an extension collect for that function and it expects a list. I can do a size 1 check to prevent any funkines with the get(0) or make it iterative

MauricioUyaguari (approved):

epsstan (approved):

PR #2556, opened 2024-01-18
Changes for dataspace constraints api

lisaknbur:

Fixes #

### DIFF
--- a/legend-engine-xts-analytics/legend-engine-xts-analytics-quality/legend-engine-xt-analytics-quality-api/src/main/java/org/finos/legend/engine/api/analytics/DataspaceQualityAnalytics.java
+++ b/legend-engine-xts-analytics/legend-engine-xts-analytics-quality/legend-engine-xt-analytics-quality-api/src/main/java/org/finos/legend/engine/api/analytics/DataspaceQualityAnalytics.java
@@ -1,0 +1,242 @@
+//  Copyright 2022 Goldman Sachs
+//
+//  Licensed under the Apache License, Version 2.0 (the "License");
+//  you may not use this file except in compliance with the License.
+//  You may obtain a copy of the License at
+//
+//       http://www.apache.org/licenses/LICENSE-2.0
+//
+//  Unless required by applicable law or agreed to in writing, software
+//  distributed under the License is distributed on an "AS IS" BASIS,
+//  WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+//  See the License for the specific language governing permissions and
+//  limitations under the License.
+
+package org.finos.legend.engine.api.analytics;
+
+import com.fasterxml.jackson.databind.ObjectMapper;
+import io.swagger.annotations.Api;
+import io.swagger.annotations.ApiOperation;
+import org.eclipse.collections.api.RichIterable;
+import org.finos.legend.engine.language.pure.compiler.toPureGraph.PureModel;
+import org.finos.legend.engine.language.pure.modelManager.ModelManager;
+import org.finos.legend.engine.protocol.pure.v1.model.context.PureModelContextData;
+import org.finos.legend.engine.protocol.pure.v1.model.packageableElement.PackageableElement;
+import org.finos.legend.engine.shared.core.ObjectMapperFactory;
+import org.finos.legend.pure.generated.Root_meta_analytics_quality_model_CheckResult;
+import org.finos.legend.pure.generated.Root_meta_pure_functions_collection_List_Impl;
+import org.finos.legend.pure.generated.core_analytics_quality_associationChecks;
+import org.finos.legend.pure.generated.core_analytics_quality_classChecks;
+import org.finos.legend.pure.generated.core_analytics_quality_enumerationChecks;
+import org.finos.legend.pure.generated.core_analytics_quality_functionChecks;
+import org.finos.legend.pure.generated.core_analytics_quality_propertyChecks;
+import org.finos.legend.pure.m3.coreinstance.meta.pure.metamodel.function.ConcreteFunctionDefinition;
+import org.finos.legend.pure.m3.coreinstance.meta.pure.metamodel.function.property.Property;
+import org.finos.legend.pure.m3.coreinstance.meta.pure.metamodel.type.Class;
+import org.finos.legend.pure.m3.coreinstance.meta.pure.metamodel.type.Enum;
+import org.finos.legend.pure.m3.coreinstance.meta.pure.metamodel.type.Enumeration;
+import org.finos.legend.pure.runtime.java.compiled.execution.CompiledExecutionSupport;
+
+import javax.ws.rs.Consumes;
+import javax.ws.rs.POST;
+import javax.ws.rs.Path;
+import javax.ws.rs.core.MediaType;
+import javax.ws.rs.core.Response;
+import java.util.ArrayList;
+import java.util.List;
+import java.util.stream.Collectors;
+
+@Api(tags = "Analytics - Quality")
+@Path("pure/v1/analytics/quality")
+public class DataspaceQualityAnalytics
+{
+    private final ModelManager modelManager;
+    private List<String> messages;
+
+    public DataspaceQualityAnalytics(ModelManager modelManager)
+    {
+        this.modelManager = modelManager;
+    }
+
+    @POST
+    @Path("check")
+    @ApiOperation("Checks the quality of provided Data space")
+    @Consumes(MediaType.APPLICATION_JSON)
+    public Response checkDataSpaceConstraints(PureModelContextData pureModelContextData)
+    {
+        messages = new ArrayList<>();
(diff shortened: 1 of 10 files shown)


davidharte-gs (approved):

legend-engine-xts-analytics/legend-engine-xts-analytics-quality/legend-engine-xt-analytics-quality-api/src/test/java/org/finos/legend/engine/api/analytics/DataspaceQualityAnalyticsTest.java
```diff
@@ -0,0 +1,97 @@
...
+import org.junit.Assert;
+import org.junit.Test;
+
+import javax.ws.rs.core.Response;
+import java.io.IOException;
+import java.util.Objects;
+
+public class DataspaceQualityAnalyticsTest
+{
+    private final DataspaceQualityAnalytics api = new DataspaceQualityAnalytics(new ModelManager(DeploymentMode.TEST));
```
MauricioUyaguari, line 32: @lisaknbur
Can you change these tests to use grammar instead of json files. 
It makes it much easier for us to see what each test case holds and exactly what you are testing. 
You can build pmcd from grammar using the composers then run your tests.

legend-engine-xts-analytics/legend-engine-xts-analytics-quality/legend-engine-xt-analytics-quality-api/src/main/java/org/finos/legend/engine/api/analytics/DataspaceQualityAnalytics.java
```diff
@@ -0,0 +1,242 @@
...
+    public DataspaceQualityAnalytics(ModelManager modelManager)
+    {
+        this.modelManager = modelManager;
+    }
+
+    @POST
+    @Path("check")
+    @ApiOperation("Checks the quality of provided Data space")
+    @Consumes(MediaType.APPLICATION_JSON)
+    public Response checkDataSpaceConstraints(PureModelContextData pureModelContextData)
```
gs-jp1, line 65: the definition of this API feels a little unclear. It is called "checkDataSpaceConstraints" but consumes a pmcd which could be anything. Should we be passing a pointer to a dataspace or a Dataspace itself. 

The implementation also feels a little unclear, we are checking the constraints on the elements in the pmcd, rather than actually what is part of the dataspace. What is the definition of "part of dataspace"? is it the elements? is it all packageable elements linked from the mappings? etc

legend-engine-xts-analytics/legend-engine-xts-analytics-quality/legend-engine-xt-analytics-quality-api/src/main/java/org/finos/legend/engine/api/analytics/DataspaceQualityAnalytics.java
```diff
@@ -0,0 +1,242 @@
...
+    {
+        messages = new ArrayList<>();
+        PureModel pureModel = this.modelManager.loadModel(pureModelContextData, null, null, null);
+
+        List<PackageableElement> classes = pureModelContextData.getElements().stream().filter(e -> e instanceof org.finos.legend.engine.protocol.pure.v1.model.packageableElement.domain.Class).collect(Collectors.toList());
```
gs-jp1, line 70: should we be just compiling the dataspcae and then call a pure function for "check constraints" on that dataspace, rather than embedding the logic in what is an API class, which ideally has no logic at all

legend-engine-xts-analytics/legend-engine-xts-analytics-quality/legend-engine-xt-analytics-quality-api/src/main/java/org/finos/legend/engine/api/analytics/DataspaceQualityAnalytics.java
```diff
@@ -0,0 +1,242 @@
...
+
+        CompiledExecutionSupport es = pureModel.getExecutionSupport();
+
+        if (classChecks(classes, pureModel, es) && associationChecks(associations, pureModel, es) && enumerationChecks(enumerations, pureModel, es) && functionChecks(functions, pureModel, es))
+        {
+            return Response.ok().build();
+        }
+        else
+        {
+            return Response.status(400).entity("Dataspace does not match the constraints - " + messages).build();
```
gs-jp1, line 83: it would be good to have a modelled response here instead of an arbitrary String. String is unclear from a client perspective, and really does not allow any useful processing from client applications.
maybe something like
{
    errors: []
}
where error has a type (class, enum, assocation etc) and a message. In Pure there is also a concept of a quality check group to be able to classify the issue type

legend-engine-xts-analytics/legend-engine-xts-analytics-quality/legend-engine-xt-analytics-quality-api/src/main/java/org/finos/legend/engine/api/analytics/DataspaceQualityAnalytics.java
```diff
@@ -0,0 +1,242 @@
...
+        }
+    }
+
+    private boolean classChecks(List<PackageableElement> classes, PureModel pureModel, CompiledExecutionSupport es)
```
gs-jp1, line 87: as per comment above, all the code from line 87 to 241 feels like it should belong in Pure code as it is effectively all working on the compiled code so no benefit to having to having this in java

legend-engine-xts-analytics/legend-engine-xts-analytics-quality/legend-engine-xt-analytics-quality-api/src/main/java/org/finos/legend/engine/api/analytics/DataspaceQualityAnalytics.java
```diff
@@ -0,0 +1,242 @@
...
+import java.util.ArrayList;
+import java.util.List;
+import java.util.stream.Collectors;
+
+@Api(tags = "Analytics - Quality")
+@Path("pure/v1/analytics/quality")
+public class DataspaceQualityAnalytics
+{
+    private final ModelManager modelManager;
+    private List<String> messages;
```
gs-jp1, line 54: resource must not have any globally modifiable state, This is completely non thread safe and the API response will be wrong in any multiple concurrent requests.

legend-engine-xts-analytics/legend-engine-xts-analytics-quality/legend-engine-xt-analytics-quality-api/src/main/java/org/finos/legend/engine/api/analytics/DataspaceQualityAnalytics.java
lisaknbur: This is calling pure functions already and then just aggregating all the responses from the individual checks

gs-jp1: yes it is calling pure functions, but the entire logic for calling them should also reside in Pure. 
We already have this, so adding a runChecksForDataSpace funtion can be added in similar fashion which then calls the runChecks here.
https://github.com/finos/legend-engine/blob/7677747c788b6bf517e6baea7ab6237d3f220761/legend-engine-xts-analytics/legend-engine-xts-analytics-quality/legend-engine-xt-analytics-quality-pure/src/main/resources/core_analytics_quality/checksEngine.pure#L101


Having code co-located between languages/contexts (resource impl vs funtional impl) vs  is far from ideal
1. The logic has been localized to the implementation of an opinionated rest API whereas the functional implementation is completely generic. e.g. if I wanted to calculate the analytics for a dataspace for another context I cannot do that with this implementation.
2. The tests are mixing concerns/language. json -> rest -> pure -> java. The tests here are really just testing the impl of the check functions themselves and the concat logic. Neither of which make sense in the context of this API really. The check functions already have tests so these tests are not adding anything beneficial. One simple tests for checking instrumentation is fine, but comprehensive logical checking is not required here.

In general
1. try to keep Resource classes to exactly that, to handle the interface of the web call and nothing else. A resource class should be very slim and have very limited/no logic
3. Any instrumentation of pure code from java should be singular in nature. If you are chaining calls together, it is a sign that the implementation likely lives in Pure also.

lisaknbur: Ah makes sense. Thanks, i will make the changes

PR #2628, opened 2024-02-14
Review temptable snowflake

gs-kotang:
#### What type of PR is this?

- Improvement

Allow users to specify the Db and schema where temp tables can be created in Snowflake. Add two optional properties - tempTableDb, tempTableSchema - to Snowflake Datasource specification. If these aren't specified we fall back to the defaults mentioned in [the documentation](https://github.com/finos/legend-engine/blob/master/docs/store/extensions/Relational/Databases/snowflake/README.md)

Fixes #

This needs a protocol release
Yes


rafaelbey:
Do we need this Snowflake logic anymore?  Can we just remove this handling and move to proper temp tables and eliminate risk associated with our custom handling?

https://docs.snowflake.com/en/user-guide/tables-temp-transient

PR #2633, opened 2024-02-15
Introduce a new API to update provided query fields and update query protocol to record originalVersionId

YannanGao-gs:
#### What type of PR is this?

improvement

Introduce a new API to update query version and update query protocol to record originalVersionId

Fixes #

### DIFF
--- a/legend-engine-application-query/src/main/java/org/finos/legend/engine/application/query/api/ApplicationQuery.java
+++ b/legend-engine-application-query/src/main/java/org/finos/legend/engine/application/query/api/ApplicationQuery.java
@@ -193,6 +193,8 @@ public class ApplicationQuery
     @Consumes({MediaType.APPLICATION_JSON})
     public Response patchQuery(@PathParam("queryId") String queryId, Query query, @ApiParam(hidden = true) @Pac4JProfileManager ProfileManager<CommonProfile> profileManager)
     {
+        MutableList<CommonProfile> profiles = ProfileManagerHelper.extractProfiles(profileManager);
+        Identity identity = IdentityFactoryProvider.getInstance().makeIdentity(profiles);
         try (Scope scope = GlobalTracer.get().buildSpan("Patch Query - update selected query fields").startActive(true))
         {
             return Response.ok().entity(this.queryStoreManager.patchQuery(queryId, query, getCurrentUser(profileManager))).build();
@@ -203,7 +205,7 @@ public class ApplicationQuery
             {
                 return ((ApplicationQueryException) e).toResponse();
             }
-            return ExceptionTool.exceptionManager(e, LoggingEventType.UPDATE_QUERY_ERROR, ProfileManagerHelper.extractProfiles(profileManager));
+            return ExceptionTool.exceptionManager(e, LoggingEventType.UPDATE_QUERY_ERROR, identity.getName());
         }
     }
 
--- a/legend-engine-application-query/src/main/java/org/finos/legend/engine/application/query/api/ApplicationQuery.java
+++ b/legend-engine-application-query/src/main/java/org/finos/legend/engine/application/query/api/ApplicationQuery.java
@@ -188,14 +188,14 @@ public class ApplicationQuery
     }
 
     @PUT
-    @Path("{queryId}/updateSelectedFields")
-    @ApiOperation(value = "Update query versionId")
+    @Path("{queryId}/patchQuery")
+    @ApiOperation(value = "Patch Query - update selected query fields")
     @Consumes({MediaType.APPLICATION_JSON})
-    public Response updateQuerySelectedFields(@PathParam("queryId") String queryId, Query query, @ApiParam(hidden = true) @Pac4JProfileManager ProfileManager<CommonProfile> profileManager)
+    public Response patchQuery(@PathParam("queryId") String queryId, Query query, @ApiParam(hidden = true) @Pac4JProfileManager ProfileManager<CommonProfile> profileManager)
     {
-        try (Scope scope = GlobalTracer.get().buildSpan("Query: Update Query Version Id").startActive(true))
+        try (Scope scope = GlobalTracer.get().buildSpan("Patch Query - update selected query fields").startActive(true))
         {
-            return Response.ok().entity(this.queryStoreManager.updateQuerySelectedFields(queryId, query, getCurrentUser(profileManager))).build();
+            return Response.ok().entity(this.queryStoreManager.patchQuery(queryId, query, getCurrentUser(profileManager))).build();
         }
         catch (Exception e)
         {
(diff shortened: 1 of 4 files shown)


legend-engine-application-query/src/main/java/org/finos/legend/engine/application/query/api/QueryStoreManager.java
```diff
@@ -386,13 +387,50 @@ else if (matchingQueries.size() == 0)
...
+    {
+        Query currentQuery = this.getQuery(queryId);
+        // Make sure only the owner can update the query
+        // NOTE: if the query is created by an anonymous user previously, set the current user as the owner
+        if (currentQuery.owner != null && !currentQuery.owner.equals(currentUser))
+        {
+            throw new ApplicationQueryException("Only owner can update the query", Response.Status.FORBIDDEN);
+        }
+
+        Class<? extends Query> queryClazz = currentQuery.getClass();
```
travisstebbins, line 408: Is this a typo?
```suggestion
        Class<? extends Query> queryClass = currentQuery.getClass();
```

legend-engine-application-query/src/test/java/org/finos/legend/engine/application/query/api/TestQueryStoreManager.java
```diff
@@ -611,6 +615,12 @@ public void testUpdateQuery() throws Exception
         queryStoreManager.createQuery(TestQueryBuilder.create("1", "query1", currentUser).build(), currentUser);
         queryStoreManager.updateQuery("1", TestQueryBuilder.create("1", "query2", currentUser).build(), currentUser);
         Assert.assertEquals("query2", queryStoreManager.getQuery("1").name);
+        Query queryWithSelectedFields = new Query();
+        queryWithSelectedFields.id = "1";
+        queryWithSelectedFields.versionId = "1.0.0";
+        queryStoreManager.updateQuerySelectedFields("1", queryWithSelectedFields, currentUser);
+        Assert.assertEquals("1.0.0", queryStoreManager.getQuery("1").versionId);
+        Assert.assertEquals("0.0.0", queryStoreManager.getQuery("1").originalVersionId);
```
travisstebbins, line 623: Could we put this in a separate unit test so we're not testing `updateQuery` and `updateQuerySelectedFields` in the same test?

legend-engine-application-query/src/main/java/org/finos/legend/engine/application/query/api/ApplicationQuery.java
```diff
@@ -180,6 +180,26 @@ public Response updateQuery(@PathParam("queryId") String queryId, Query query, @
         }
     }
 
+    @PUT
+    @Path("{queryId}/updateSelectedFields")
+    @ApiOperation(value = "Update query versionId")
+    @Consumes({MediaType.APPLICATION_JSON})
+    public Response updateQuerySelectedFields(@PathParam("queryId") String queryId, Query query, @ApiParam(hidden = true) @Pac4JProfileManager ProfileManager<CommonProfile> profileManager)
```
travisstebbins, line 187: Is the point of this endpoint to update only the fields of the query with `queryId` that are part of the `query` parameter? If so, I think it would make sense to make this endpoint a `PATCH` endpoint instead of a `PUT` endpoint, since we're only trying to update the fields present in the payload `query` object and not replace the entire query (https://developer.mozilla.org/en-US/docs/Web/HTTP/Methods/PATCH, https://www.geeksforgeeks.org/difference-between-put-and-patch-request/). Then, we could maybe rename this endpoint and rename the `queryStoreManager` function from `updateQuerySelectedFields` to `patchQuery`.
```suggestion
    @PATCH
    @Path("{queryId}")
    @ApiOperation(value = "Patch query")
    @Consumes({MediaType.APPLICATION_JSON})
    public Response patchQuery(@PathParam("queryId") String queryId, Query query, @ApiParam(hidden = true) @Pac4JProfileManager ProfileManager<CommonProfile> profileManager)
```

YannanGao-gs: Screenshot 2024-03-04 at 11 30 38 AM
Thank you! However, it seems like Patch is not supported in engine's current dependency.

YannanGao-gs: Thank you! With current Engine's dependency. PATCH operation is not supported. Since we wrote our own update logic, do you think if it's okay with keeping use PUT? 
I've updated the function name to be patch instead

travisstebbins: Yeah if our dependency doesn't support `PATCH` operation then `PUT` is probably the next best option, but we can make the function called `patch` still since that's what it's actually doing.
Hopefully in the future if we update our dependencies and `PATCH` is supported, then we could update the method type.

legend-engine-application-query/src/main/java/org/finos/legend/engine/application/query/api/QueryStoreManager.java
YannanGao-gs: 👀 I think it's just a naming convention per each developer's preference loll 👀😅😅

travisstebbins: Oh interesting. Feels weird to me haha, but if it's used elsewhere then should be fine

legend-engine-application-query/src/main/java/org/finos/legend/engine/application/query/api/QueryStoreManager.java
```diff
@@ -386,13 +387,50 @@ else if (matchingQueries.size() == 0)
...
+        for (java.lang.reflect.Field field : queryClazz.getDeclaredFields())
+        {
+            try
+            {
+                field.setAccessible(true);
+                Object updatedValue = field.get(updatedQuery);
+                if (updatedValue != null)
```
MauricioUyaguari, line 415: this might require some rethinking about if we have the api take from `Query updatedQuery` to something ese
but one use case this current flow does not support is if we want to set  a value to null if it has already been set to something. 
woud using the mongo set operation help here, and update only those fields?

legend-engine-application-query/src/main/java/org/finos/legend/engine/application/query/api/QueryStoreManager.java
MauricioUyaguari: not a big deal but I think this modue is off in naming convention, we should probably start being more reader friendly and start naming things from clazz to class.

legend-engine-application-query/src/main/java/org/finos/legend/engine/application/query/api/QueryStoreManager.java
YannanGao-gs: If the patchQuery function still takes in a parameter of type Query that we are going to update fields upon, to generate the payload of mongo set operation, I think we'll still need to compose a document by iterating the provided parameter to get non-null fields, then call set? 😅😅

Or,  could you please be more specific about what the parameter type should be? 👀😅😅
Thanks

MauricioUyaguari (approved):

PR #2641, opened 2024-02-20
Introduce PureModelContextCollection type and use it in WorkspaceSDLCLoader

abhishoya-gs:
#### What type of PR is this?
Improvement

As of today, WorkspaceSDLCLoader calls ModelLoader for each dependency. This introduces as many numbers of HTTP requests and doesn't account for dependency override. 

Solution

- Introduce PureModelContextCollection type
- Modify WorkspaceSDLCLoader to call ModelLoader with a PureModelContextCollection 
- Support fetching PureModelContextCollection using AlloySDLCLoader

Additionally, PureModelContextCollection can be used further to send smaller payloads from clients like Studio.

Fixes #

For testing, I rely on change in test setup and manual testing.

No


MauricioUyaguari:
@abhishoya-gs 
can we not change the model loader and add another type of PureModelContext. 
I think @rafaelbey  started something. see : https://github.com/finos/legend-engine/pull/2446

abhishoya-gs:
> @abhishoya-gs can we not change the model loader and add another type of PureModelContext. I think @rafaelbey started something. see : #2446

Okay sure, seems like a better approach

legend-engine-core/legend-engine-core-language-pure/legend-engine-language-pure-modelManager-sdlc/src/main/java/org/finos/legend/engine/language/pure/modelManager/sdlc/SDLCLoader.java
```diff
@@ -148,30 +162,38 @@ private boolean isCacheableAlloySDLC(SDLC sdlc)
...
+            {
+                return this.alloyLoader.getCacheKey(context);
+            }
         }
-        else
+        else if (context instanceof PureModelContextCollection)
         {
-            return this.alloyLoader.getCacheKey(context);
+            PureModelContextCollection pureModelContextCollection = (PureModelContextCollection) context;
+            Assert.assertTrue(pureModelContextCollection.getContexts().stream().allMatch(pureModelContext -> pureModelContext instanceof PureModelContextPointer), () -> "Invalid type of PureModelContext in PureModelContextCollection for cacheKey");
```
rafaelbey, line 181: This constrains defeats one of the main purposes of the collection I wanted to introduce - improve studio flow. For example - studio can manage a collection, where an entry is a PMCD of the current project, and the other entries are pointer to the project dependencies.

abhishoya-gs: I will remove this constraint so it will help improve studio flow. I will also change the constraint added on shouldCache but since we cache compiled model only for PureModelContextPointer, we won't gain much there in case of studio.

Also, I have placed another check in SDLCFetcher [see here](https://github.com/finos/legend-engine/pull/2641/files#diff-5e7885c22761b0fcb1da9a3fb4d667c3c60def193a97f369f9e5aca01227af74R96). It's required as my implementation only uses alloy loaders and thus only AlloySDLC. If there's a use case to support sourcing from multiple loaders the logic would have to be changed accordingly.

finos-admin:

finos-admin:

PR #2649, opened 2024-02-21
BigQuery Extension - Adjust pom.xml to fix build errors

abhishoya-gs:

Fixes #


epsstan:
This PR is temporarily removing the BigQuery wrapper module from the build. 

The module harcodes a version number which is breaking the build. This will be fixed properly in a follow-up PR.

epsstan (approved):

PR #2659, opened 2024-02-27
Identity refactoring

abhishoya-gs:
#### What type of PR is this?
Refactoring

As of today we have tight coupling between pac4j profiles and execution code. This makes it difficult to execute queries without pac4j context such as using REPL layer. This PR refactors all non-HTTP functions to accept Identity instead of pac4j CommonProfile.

Fixes #

No


legend-engine-xts-graphQL/legend-engine-xt-graphQL-query/src/main/java/org/finos/legend/engine/query/graphQL/api/GraphQL.java
```diff
@@ -55,23 +55,23 @@ protected PureModel loadSDLCProjectModel(MutableList<CommonProfile> profiles, Ht
         PureModelContextPointer pointer = new PureModelContextPointer();
         pointer.sdlcInfo = sdlcInfo;
 
-        Subject subject = ProfileManagerHelper.extractSubject(profiles);
+        Subject subject = identity.getSubjectFromIdentity();
```
epsstan, line 58: We should fix this in the next pass. There's no guarantee we have a subject here

legend-engine-xts-mongodb/legend-engine-xt-nonrelationalStore-mongodb-executionPlan/src/main/java/org/finos/legend/engine/plan/execution/stores/mongodb/plugin/MongoDBExecutionNodeExecutor.java
```diff
@@ -64,12 +60,12 @@
...
 {
-    private final MutableList<CommonProfile> profiles;
+    Identity identity;
     private ExecutionState executionState;
 
-    public MongoDBExecutionNodeExecutor(MutableList<CommonProfile> profiles, ExecutionState executionState)
+    public MongoDBExecutionNodeExecutor(Identity identity, ExecutionState executionState)
     {
-        this.profiles = profiles;
+        this.identity = identity;
```
epsstan, line 68: Is this instance used ? If not, lets remove

PR #2672, opened 2024-03-01
Dataspace can be used as interface for stores in Runtime and Function Tests

haroonsherjan:
#### What type of PR is this?
improvement


rafaelbey (approved):

rafaelbey (approved):

legend-engine-core/legend-engine-core-language-pure/legend-engine-protocol-pure/src/main/java/org/finos/legend/engine/protocol/pure/v1/model/packageableElement/function/StoreTestData.java
```diff
@@ -28,7 +29,7 @@
 public class StoreTestData
 {
     public String doc;
-    public String store;
+    public StoreProviderPointer store;
```
MauricioUyaguari, line 32: this is breaking existing use case, we need to make sure we have custom deserialization in place for string values already coming in. 

https://github.com/finos/legend/blob/master/showcases/data/Store/Relational%20Store/Function/code.pure#L184

PR #2676, opened 2024-03-05
Make EngineUrlStreamHandlerFactory safe wrt to changes in JDK15+

aziemchawdhary-gs:
TreeMap implementation has been changed in Java 15 to optimize some functions. As a result running embedded service executions concurrently in the same JVM can now throw concurrent modification exceptions when running on JVM 15 onwards. 
Note that TreeMap has never been threadsafe - just that exceptions are now thrown on Java 15 onwards. 

Switching to a threadsafe map implementation to avoid exceptions being thrown for concurrent executions of the same service on the same JVM. 

References:
https://www.oracle.com/java/technologies/javase/15-relnote-issues.html
https://bugs.java.com/bugdatabase/view_bug.do?bug_id=JDK-8227666
https://hg.openjdk.org/jdk/jdk/rev/65b345254ca3


davidharte-gs (commented):
Can we add a failing concurrency test?

rafaelbey (approved):

PR #2692, opened 2024-03-12
Use native git with git-commit-id-maven plugin

aziemchawdhary-gs:
Support development flow for using git work-tree on this project. 
We use the git-commit-id Maven plugin which by default uses jgit. When using legend-engine with git work-tree the build fails because jgit does not support git work-trees and cannot identify a commit when running. 

This fixes the project to instead use the git command line rather than jgit in this plugin. 
This will have potential impact downstream, particularly in build pipelines which will need to have git installed. 
The Github runners we use have git installed by default but other build infrastructure will have to ensure they have git installed.


aziemchawdhary-gs:
Marking as draft for now whilst I investigate impact downstream

finos-admin:

finos-admin:

PR #2697, opened 2024-03-13
Minor fixes

lisaknbur:
#### What type of PR is this?

Small fix

Fixes #


legend-engine-config/legend-engine-server/pom.xml
```diff
@@ -654,7 +654,16 @@
             <artifactId>legend-engine-xt-relationalStore-store-entitlement-analytics</artifactId>
             <scope>runtime</scope>
         </dependency>
-
+        <dependency>
+            <groupId>org.finos.legend.engine</groupId>
+            <artifactId>legend-engine-xt-analytics-quality-api</artifactId>
+            <version>${project.version}</version>
```
MauricioUyaguari, line 660: shouldnt need project.version. the module should be in the main pom.

legend-engine-config/legend-engine-server/pom.xml
```diff
@@ -1031,12 +1040,12 @@
         <dependency>
             <groupId>org.finos.legend.engine</groupId>
             <artifactId>legend-engine-xt-relationalStore-executionPlan-connection</artifactId>
-            <scope>test</scope>
+<!--            <scope>test</scope>-->
```
MauricioUyaguari, line 1043: we should probably remove this entirely. why just comment out?

MauricioUyaguari (changes requested):

PR #2716, opened 2024-03-19
[Legend SQL] Add ability to externalize in arrow format

gs-manvig:
#### What type of PR is this?
Improvement

Allow Alloy SQL clients to request data externalized to arrow format.

This allows users to use a memory efficient serialization format than JSON.

### DIFF
--- a/legend-engine-xts-json/legend-engine-xt-json-runtime/pom.xml
+++ b/legend-engine-xts-json/legend-engine-xt-json-runtime/pom.xml
@@ -110,6 +110,10 @@
             <groupId>org.finos.legend.engine</groupId>
             <artifactId>legend-engine-shared-core</artifactId>
         </dependency>
+        <dependency>
+            <groupId>org.finos.legend.engine</groupId>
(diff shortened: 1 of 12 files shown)


legend-engine-xts-json/legend-engine-xt-json-runtime/pom.xml
```diff
@@ -110,6 +110,10 @@
             <groupId>org.finos.legend.engine</groupId>
             <artifactId>legend-engine-shared-core</artifactId>
         </dependency>
+        <dependency>
+            <groupId>org.finos.legend.engine</groupId>
+            <artifactId>legend-engine-xt-relationalStore-executionPlan</artifactId>
+        </dependency>
```
rafaelbey, line 116: Per my comment on the other PR, we need to avoid bringing specific stores to "core" functionality like externalize.  The goal is that externalize works with "any" store, and we need to find a way to dispatch to respective store modules, if we need to.

finos-admin:

finos-admin:

PR #2736, opened 2024-03-27
Make all identifiers for enumSourceValue in EnumMappings as single-quoted References

sprisha:
For Enum Mappings, if a user passes enum references as a source value using form Mode like this: 
image
it converts to text mode like this: 
image
And this fails compiling because the latter is unable to parse properly with the existing antlr rules. 

So now, we will make it so we single-quote each identifier in a enum reference, like this: 
image
This compiles properly and fixes the roundtrip issue. 
#### What type of PR is this?

 Bug fix for studio 

 Described above.

Fixes #

 No.

### DIFF
--- a/legend-engine-core/legend-engine-core-base/legend-engine-core-language-pure/legend-engine-language-pure-grammar/src/main/java/org/finos/legend/engine/language/pure/grammar/to/PureGrammarComposerUtility.java
+++ b/legend-engine-core/legend-engine-core-base/legend-engine-core-language-pure/legend-engine-language-pure-grammar/src/main/java/org/finos/legend/engine/language/pure/grammar/to/PureGrammarComposerUtility.java
@@ -150,7 +150,7 @@ public class PureGrammarComposerUtility
         }
         else
         {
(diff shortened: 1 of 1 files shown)


legend-engine-core/legend-engine-core-base/legend-engine-core-language-pure/legend-engine-language-pure-grammar/src/main/java/org/finos/legend/engine/language/pure/grammar/to/PureGrammarComposerUtility.java
```diff
@@ -150,7 +150,7 @@ public static String convertIdentifier(String val, boolean doubleQuotes)
         }
         else
         {
-            return UNQUOTED_IDENTIFIER_PATTERN.matcher(val).matches() ? val : convertString(val, true, doubleQuotes);
+            return convertString(val, true, doubleQuotes);
```
rafaelbey, line 153: This is not a good change, as will trigger many breaks on test cases, and will impact users that dont need quote.

I though the change is either to enhance the regex, or enhance the grammar definition so we can parse the value properly.

finos-admin:

finos-admin:

PR #2745, opened 2024-04-02
Support server side push down operations for REPL using Ag-grid

gayathrir11:


legend-engine-config/legend-engine-repl/legend-engine-repl-relational/pom.xml
```diff
@@ -26,6 +26,36 @@
     <artifactId>legend-engine-repl-relational</artifactId>
     <name>Legend Engine - REPL - Relational</name>
 
+    <properties>
+        <skip.yarn>false</skip.yarn>
+        <repl.web-application.version>11.29.0</repl.web-application.version>
+        <repl.web-application.url>https://registry.npmjs.org/@finos/legend-application-repl-deployment/-/legend-application-repl-deployment-${repl.web-application.version}.tgz</repl.web-application.url>
```
aziemchawdhary-gs, line 32: Is it possible to make the initial part of the URL a Maven property? We have two places where we download dependencies from a NPM repository (here and also Pure IDE Light HTTP server pom). 
If we could extract out https//registry.npmjs.org into a central place then that makes it easier to change the URL when building in a corporate enviroment where registry.npmjs.org may be blocked.

rafaelbey (approved):

rafaelbey (changes requested):
The build failed on the dependency enforcer

> [INFO] --- maven-dependency-plugin:2.10:analyze-only (dependency-analyze) @ legend-engine-repl-relational ---
Warning:  Used undeclared dependencies found:
Warning:     com.fasterxml.jackson.core:jackson-annotations:jar:2.10.5:compile

pierredebelen (approved):

PR #2756, opened 2024-04-05
add handling for larger Float and Decimal numbers

shtirlets:
#### What type of PR is this?
Improvement

Add handling for larger Float and Decimal numbers. Currently the larger numbers get truncating providing incorrect results
Add trace details to the error response

Fixes #


gs-manvig:
Looks good to me

aormerod-gs (approved):

PR #2766, opened 2024-04-10
REPL - pass identity to plan executor

abhishoya-gs:
#### What type of PR is this?
Improvement

This PR allows users to connect to databases using local subject identity

Fixes #

Yes


epsstan (approved):
@abhishoya-gs - In the next PR we should remove the compile time dependency on javax.Subject

PR #2783, opened 2024-04-16
Java Generation: Generate classes for concrete function parameters 

aziemchawdhary-gs:
M2M execution relies on source fetch tree computation to generate the list of Pure classes to generate Java code for. 
However, if a function is passed an empty argument then there is no property access and hence nothing is present in the source tree informing the platform that Java code should be generated for the corresponding parameter type. As a result the compilation of Java code present in the execution plan fails due to missing classes. 

This change fixes that by processing the parameters and generating code for parameter types that are not part of the Java standard library or engine conventions library.  If we happen to have already generated code for the parameter type earlier in the plan generation (e.g the class is mentioned in the source fetch tree) then the Java generation merging code will handle duplicates that may be generated by this changeset.

### DIFF
--- a/legend-engine-xts-java/legend-engine-xt-javaGeneration-pure/src/main/resources/core_external_language_java/metamodel_factories.pure
+++ b/legend-engine-xts-java/legend-engine-xt-javaGeneration-pure/src/main/resources/core_external_language_java/metamodel_factories.pure
@@ -2874,6 +2874,40 @@ function meta::external::language::java::factory::knownPackage(pkg: meta::extern
    )))))))));
 }
 
+function meta::external::language::java::factory::isKnownPackage(pkg: meta::external::language::java::metamodel::Package[1]):Boolean[1]
+{
+     let pkgName = $pkg->packageToString();
+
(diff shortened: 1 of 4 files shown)


legend-engine-xts-java/legend-engine-xt-javaGeneration-pure/src/main/resources/core_external_language_java/metamodel_factories.pure
```diff
@@ -2874,6 +2874,40 @@ function meta::external::language::java::factory::knownPackage(pkg: meta::extern
    )))))))));
 }
 
+function meta::external::language::java::factory::isKnownPackage(pkg: meta::external::language::java::metamodel::Package[1]):Boolean[1]
```
davidharte-gs, line 2877: Can probably clean this up with Pierre's new-ish multi-if support:

https://github.com/finos/legend-pure/blob/master/legend-pure-functions/legend-pure-functions-base/legend-pure-m2-functions-base-pure/src/main/resources/platform_functions/lang/if.pure#L15

davidharte-gs (approved):

PR #2823, opened 2024-05-02
Persistence Component: Fix Unitemporal snapshot partition to support multiple partition key values

prasar-ashutosh:
#### What type of PR is this?
- Bug Fix

Fix Unitemporal snapshot partition to support multiple partition key values
The current implementation does not support multiple values for multiple partition keys.
This implementation will allow users to provide a list of partition specs. Each partition spec is a map of partition key and corresponding value.

Fixes #

NO


legend-engine-xts-persistence/legend-engine-xt-persistence-component/legend-engine-xt-persistence-component-logical-plan/src/main/java/org/finos/legend/engine/persistence/components/util/LogicalPlanUtils.java
```diff
@@ -130,14 +130,24 @@ public static Condition getPartitionColumnsMatchCondition(Dataset mainDataSet, D
...
+    public static Condition getPartitionSpecMatchCondition(Dataset dataSet, List<Map<String, String>> partitionSpecList)
     {
-        return getColumnValueMatchInCondition(dataSet, partitionFilter);
+        return Or.of(partitionSpecList.stream()
+                 .map(partitionSpec -> And.of(
+                         partitionSpec.entrySet().stream()
+                                 .map(columnValuePair ->
+                                         Equals.of(
+                                                 FieldValue.builder().datasetRef(dataSet.datasetReference()).fieldName(columnValuePair.getKey()).build(),
+                                                 StringValue.of(columnValuePair.getValue()))
```
kumuwu, line 147: May be a good chance to fix this one: here it may not always be a StringValue, it could be an ObjectValue when the field is a Number (so that we do not print the quotes around the value); similarly we can fix this one for partitionValuesByField case too

legend-engine-xts-persistence/legend-engine-xt-persistence-component/legend-engine-xt-persistence-component-logical-plan/src/main/java/org/finos/legend/engine/persistence/components/ingestmode/UnitemporalSnapshotAbstract.java
```diff
@@ -89,6 +98,29 @@ default void validate()
...
+        if (!partitionSpecList().isEmpty())
+        {
+            for (Map<String, String> partitionSpec : partitionSpecList())
+            {
+                if (partitionFields().size() != partitionSpec.size())
+                {
+                    throw new IllegalStateException("Can not build UnitemporalSnapshot, size of each partitionSpec must be same as size of partitionFields");
+                }
+            }
         }
```
kumuwu, line 124: So we are still expecting the users to pass both partitionFields and partitionSpecList in this use case - so shall we validate the the partition keys in the spec list match the ones specified in partitionFields? (we are doing this for the partitionValuesByField case, so just  wondering if we should do this here too for consistency)

kumuwu:
We also need some changes in IngestModeCaseConverter, may be good to add a test for upper case after that too

legend-engine-xts-persistence/legend-engine-xt-persistence-component/legend-engine-xt-persistence-component-logical-plan/src/main/java/org/finos/legend/engine/persistence/components/ingestmode/UnitemporalSnapshotAbstract.java
```diff
@@ -89,6 +98,29 @@ default void validate()
...
+            for (Set<String> partitionValues: partitionValuesByField().values())
+            {
+                if (partitionValues.size() > 1)
+                {
+                    partitionKeysWithMoreThanOneValues++;
+                }
+            }
+            if (partitionKeysWithMoreThanOneValues > 1)
+            {
+                throw new IllegalStateException(String.format("Can not build UnitemporalSnapshot, partitionValuesByField does not support more than 1 value for more than one partition key"));
```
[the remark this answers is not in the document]
prasar-ashutosh: Sure will do as suggested.

legend-engine-xts-persistence/legend-engine-xt-persistence-component/legend-engine-xt-persistence-component-logical-plan/src/main/java/org/finos/legend/engine/persistence/components/util/LogicalPlanUtils.java
prasar-ashutosh: In that case the interface needs to be a List<Map<String, Object>>. 
Don't think we can fix for partitionValuesByField as the type of value is String there
Map<String, Set<String>> partitionValuesByField();

legend-engine-xts-persistence/legend-engine-xt-persistence-component/legend-engine-xt-persistence-component-logical-plan/src/main/java/org/finos/legend/engine/persistence/components/ingestmode/UnitemporalSnapshotAbstract.java
prasar-ashutosh: Good point, will add it.

prasar-ashutosh:
> We also need some changes in IngestModeCaseConverter, may be good to add a test for upper case after that too
Sure, Will make this change as well add test for it as well

PR #2832, opened 2024-05-07
Enable CORS requests for REPLGridServer

gayathrir11:

Fixes #


akphi (changes requested):
Let's try using Filter rather than having to call `enableCors` for every end points

PR #2838, opened 2024-05-08
Lineage: Assert on type of property mapping

aziemchawdhary-gs:
Add an assert in lineage code and throw a useful error message when processing a unsupported property mapping type


aziemchawdhary-gs:
Fixing properly for M2M rather than just the assert on  unsupported property mappings.

PR #2842, opened 2024-05-10
REPL - Cache command

gs-ssh16:
#### What type of PR is this?

Improvement

Add cache command to REPL client. Usage: `cache <connection> <pure expression>`


akphi (approved):

akphi:
Looks great 🚀 ! If I understand correctly, this keeps adding more table with prefix `test` to save snapshot of the data as the query being built up; if that's correct then perhaps one small improvement we can make is to print out the query corresponding to the snapshot

akphi:
actually, I think I misunderstood the feature, please disregard my comment.

PR #2858, opened 2024-05-20
Stereotypes support for Relational Databases

mrudula-gs:
#### What type of PR is this?
- Improvement

Ability to tag stores with profiles to be able to identify and route during plan generation

### DIFF
--- a/legend-engine-core/legend-engine-core-base/legend-engine-core-language-pure/legend-engine-language-pure-compiler/pom.xml
+++ b/legend-engine-core/legend-engine-core-base/legend-engine-core-language-pure/legend-engine-language-pure-compiler/pom.xml
@@ -260,6 +260,10 @@
             <artifactId>hamcrest-core</artifactId>
             <scope>test</scope>
         </dependency>
+        <dependency>
+            <groupId>org.finos.legend.engine</groupId>
+            <artifactId>legend-engine-xt-relationalStore-protocol</artifactId>
+        </dependency>
         <!-- TEST -->
     </dependencies>
 </project>
--- a/legend-engine-core/legend-engine-core-base/legend-engine-core-language-pure/legend-engine-language-pure-compiler/pom.xml
+++ b/legend-engine-core/legend-engine-core-base/legend-engine-core-language-pure/legend-engine-language-pure-compiler/pom.xml
@@ -260,10 +260,6 @@
             <artifactId>hamcrest-core</artifactId>
             <scope>test</scope>
         </dependency>
-        <dependency>
-            <groupId>org.finos.legend.engine</groupId>
-            <artifactId>legend-engine-xt-relationalStore-protocol</artifactId>
-        </dependency>
         <!-- TEST -->
     </dependencies>
 </project>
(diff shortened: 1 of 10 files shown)


legend-engine-core/legend-engine-core-base/legend-engine-core-language-pure/legend-engine-language-pure-compiler/pom.xml
```diff
@@ -260,6 +260,10 @@
             <artifactId>hamcrest-core</artifactId>
             <scope>test</scope>
         </dependency>
+        <dependency>
```
aziemchawdhary-gs, line 263: This feels like the wrong thing to do: we shouldn't be adding a dependency from the core compiler to an xt module.

legend-engine-core/legend-engine-core-base/legend-engine-core-language-pure/legend-engine-language-pure-compiler/src/main/java/org/finos/legend/engine/language/pure/compiler/toPureGraph/PackageableElementFirstPassBuilder.java
```diff
@@ -85,6 +86,10 @@ public PackageableElementFirstPassBuilder(CompileContext context)
     @Override
     public PackageableElement visit(org.finos.legend.engine.protocol.pure.v1.model.packageableElement.PackageableElement element)
     {
+        if (element instanceof Database)
```
aziemchawdhary-gs, line 89: Shouldn't this happen in the compiler for the relational grammar? The grammar for the relational parser has been changed in this PR - I'd expect the compiler change to also happen in that module.

aziemchawdhary-gs (changes requested):
Perhaps its better to add stereotype to Store instead of Database? That way you could make the change you want in the core language compiler but without needing to add a dependency to an xt-protocol module in the core language compiler

gs-ssh16:
If stereotypes are being supported, I see no reason why tagged values shouldn't be. Its extra metadata which can be added on the stores

legend-engine-xts-relationalStore/legend-engine-xt-relationalStore-generation/legend-engine-xt-relationalStore-grammar/src/main/antlr4/org/finos/legend/engine/language/pure/grammar/from/antlr4/RelationalParserGrammar.g4
```diff
@@ -44,6 +44,11 @@ database:                                   DATABASE qualifiedName
 include:                                    INCLUDE qualifiedName
 ;
 
+// -------------------------------------- STEREOTYPE --------------------------------------
+
+stereotypes:                        LESS_THAN LESS_THAN stereotype (COMMA stereotype)* GREATER_THAN GREATER_THAN;
+
+stereotype:                         qualifiedName DOT identifier;
```
gs-ssh16, line 51: Rather than duplicating this, we should reuse from M3Grammar/DomainGrammar

mrudula-gs: Adding that import might introduce complications due to potential keyword conflicts and ambiguous references as DomainParserGrammar -> M3ParserGrammar -> CoreParserGrammar and RelationalParserGrammar -> CoreParserGrammar

PR #2859, opened 2024-05-21
Add support for Pivot

akphi:


akphi:
```
# Summary

We are bringing the magic of SecView `DataCube` in `Alloy`!

Unlike the existing query builder tools in Alloy like Databrowser and Alloy Query, where the focus is the form editor which is used to incrementally build the query, `DataCube` focus on allowing users incrementally build up the grid view of their data.

[screenshot]

One can think of `DataCube` as a data report builder tool, the report built by `DataCube` contains 2 parts: the data-fetching layer, i.e. the data query, and the layout customization. These are also the 2 main pieces of work: improving data query capabilities in Alloy and bringing in the `DataCube`'s rich set of layout features.

# Layout Features


# Query Capabilities

Alloy has supported most query functions that `DataCube` support, with exceptions for a few, such as `pivot` (also known as `h-pivot`), `group by rollup`.
https://docs.snowflake.com/en/sql-reference/constructs/pivot
https://docs.snowflake.com/en/sql-reference/constructs/group-by-rollup

- [ ] TDS source:
  - [x] project()
  - [x] Import CSV
  - [ ] Service
  - [ ] DataSpace
- [x] extend()
- [x] distinct()
- [x] drop()
- [x] filter()
- [x] groupBy(): aka v-pivot
- [ ] groupByRollUp()/groupByCube()
- [ ] limit
- [ ] pivot(): aka h-pivot
- [ ] pivotWithTotal()
- [x] rename()
- [ ] slice
- [ ] sort
- [x] select(): for show/hide columns

The general form of the query being built/supported by this UI is

```
TDS
  ->groupBy() // v-pivot (form the tree map on the left)
  ->pivot() // h-pivot
  ->extend() // allow creating new columns
  ->filter()
  ->select()
  ->rename()
  ->sort()
  ->limit()
  (->slice()) // an optimization we can do to lessen memory usage
```

> We must confirm the form of the query as the more things coming after `pivot()` the more complicated/brittle the query is potentially

# Application Architecture

- [ ] Support different query edition workflows (this potentially allow us to merge back into `Alloy Query`)
- [ ] Support saving the query with layout configuration (consider saving this to local disk for now or MongoDB store)
- [ ] Do not auto-apply changes like `Alloy Query` so we can better support undo/redo

```

akphi:
# Future

- `FUTURE` Split View (2 queries or more and you can use the row as the input for the bottom one)
- `FUTURE` Local caching
- `FUTURE` Support Paremters
- `FUTURE` Plotting/Graphing?
- `FUTURE` Streaming data!
- `FUTURE` When specifying the filter, we can search for value: think about this, this can be turned on/off as it potentially stresses the DB
- `FUTURE` Support undo/redo
- `FUTURE` Show result set summary (time it takes, number of rows, etc.)

- Horizontal Pivot (aka pivot):
  - The total columns (--> this will require group by rollup)
  - Arrangement of the total columns (last or first)
  - Default aggregation per data type: do not aggregate on strings
  - Columns can be excluded from h-pivot
- Vertical Pivot (aka group-by):
  - Root aggregation (basically we need group-by rollup)
  - Tree representation: that supports drilldown (drilling down will fire sub-query)
- Extended Column (aka Derivation)
  - Create derivation by right-clicking the grid
  - Concern: what is the order of operation: can we filter on extended columns, can we pivot, etc.?
- `LAYOUT` Column pinning: you can drag and drop the vertical rule next the v-pivot to freeze the column when scrolling horizontally
- Column Rename
- Column Selection
- Column Rearrange
- In-grid Interaction:
  - `FUTURE` Filter out/in like in query grid
  - Column Auto-Sizing: double click to
- Filter
  - Add new filter condition for selected value(s)
- Sort
  - In grid sorting
  - Absolute sorting
- Layout:
  - `LAYOUT` Grid
    - Hide/Show root aggregation
    - Row limit: affects all level (even for the drilldown -> show a warning about the row limit to user at the bottom of the screen)
    - Show selection stats: the stats of the selected columns
    - Show lines: Tree, Horizontal, Vertical
    - Scale Number: basis point (0.01%), percentage (%), million (m), billion (b), thousands (k), auto (b/m/k)
    - Display warning when truncated (show as ######), or display elipsis when truncated
    - Font: select font, size, italic, bold
    - Color: fore/background: normal, negative, zero, error
    - Alternatate color every ? row, choose color
  - `LAYOUT` Column
    - Blur: make secret/redact?
    - Font: see above
    - For numeric only: decimal points (how many digits after comma), display comma, negative value in parenthesis, scale (see above)

gayathrir11:
[result.csv](https://github.com/finos/legend-engine/files/15402460/result.csv)
Northwind data for testing

gayathrir11:
[result (2).csv](https://github.com/finos/legend-engine/files/15496512/result.2.csv)
Northwind data

PR #2875, opened 2024-05-29
External Formats: Add support for serialization/deserialization config

aziemchawdhary-gs:


legend-engine-core/legend-engine-core-pure/legend-engine-pure-code-compiled-core/src/main/resources/core/pure/binding/executionPlan/executionPlan_generation.pure
```diff
@@ -170,6 +180,13 @@ function <<access.private>> meta::external::format::shared::executionPlan::inter
    let enableConstraints  = $state.exeCtx.enableConstraints->isEmpty() || $state.exeCtx.enableConstraints->toOne();
    let bindingArg         = $fe.parametersValues->at(1);
    let inScopeVars        = $state.inScopeVars;
+   let config             = if($fe.parametersValues->size()==4,
```
davidharte-gs, line 183: would it be better to check against the func? i.e if($fe.func->in([meta::external::format::shared::functions::internalize_Class_1__Binding_1__String_1__ExternalFormatInternalizeConfig_1__T_MANY_,...] ...

PR #2876, opened 2024-05-30
Introduce PCT (& upgrade to Legend 5.8.1)

pierredebelen:


.github/workflows/build.yml
```diff
@@ -82,7 +82,7 @@ jobs:
       - name: Build (PR)
         if: github.ref != 'refs/heads/master'
         run: |
-          mvn -B -e -DskipTests=true install
+          mvn -T3 -B -e -DskipTests=true install -DargLine="-XX:MaxRAMPercentage=90.0"
```
kevin-m-knight-gs, line 85: Setting -DargLine here isn't useful, since we're skipping tests.
```suggestion
          mvn -T3 -B -e -DskipTests=true install"
```

.github/workflows/build.yml
```diff
@@ -100,7 +100,7 @@ jobs:
 
       - name: Test
         run: |
-          mvn -B -e surefire:test -DargLine="-XX:MaxRAMPercentage=70.0" -Dsurefire.reports.directory=${GITHUB_WORKSPACE}/surefire-reports-aggregate
+          mvn -T3 -B -e surefire:test -DargLine="-XX:MaxRAMPercentage=90.0" -Dsurefire.reports.directory=${GITHUB_WORKSPACE}/surefire-reports-aggregate
```
kevin-m-knight-gs, line 103: If we're going to be running 3 test VMs simultaneously, do we really want to set the max RAM percentage for each to be 90%?

kevin-m-knight-gs (dismissed):

gs-ssh16 (approved):

PR #2878, opened 2024-05-31
add lastOpenAt in Query

YannanGao-gs:
#### What type of PR is this?

Improvement

add lastOpenAt in Query

Fixes #

### DIFF
--- a/legend-engine-application-query/src/main/java/org/finos/legend/engine/application/query/api/QueryStoreManager.java
+++ b/legend-engine-application-query/src/main/java/org/finos/legend/engine/application/query/api/QueryStoreManager.java
@@ -23,6 +23,7 @@ import com.mongodb.client.MongoDatabase;
 import com.mongodb.client.model.Aggregates;
 import com.mongodb.client.model.Field;
 import com.mongodb.client.model.Filters;
+import com.mongodb.client.model.Updates;
 import com.mongodb.client.model.Projections;
 import com.mongodb.client.model.Sorts;
 import org.bson.Document;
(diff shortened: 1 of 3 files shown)


legend-engine-application-query/src/main/java/org/finos/legend/engine/application/query/api/QueryStoreManager.java
```diff
@@ -309,7 +309,10 @@ else if (matchingQueries.size() == 0)
         {
             throw new ApplicationQueryException("Can't find query with ID '" + queryId + "'", Response.Status.NOT_FOUND);
         }
-        return matchingQueries.get(0);
+        Query query = matchingQueries.get(0);
+        query.lastOpenAt = Instant.now().toEpochMilli();
+        this.getQueryCollection().findOneAndReplace(Filters.eq("id", queryId), queryToDocument(query));
```
MauricioUyaguari, line 314: can we not do a full find and replace here and just update the lastOpenAt property instead ?

MauricioUyaguari (changes requested):

legend-engine-application-query/src/main/java/org/finos/legend/engine/application/query/api/QueryStoreManager.java
YannanGao-gs: this getQuery()  is not created by this PR.
What I am doing here is just updating the `lastOpenAt `

YannanGao-gs: sry May I know how this new information will be stored back in Mongo if not doing this 
` this.getQueryCollection().findOneAndReplace(Filters.eq("id", queryId), queryToDocument(query));`

MauricioUyaguari: couldnt we just set the lastOpenAt property, something like -> https://www.mongodb.com/docs/manual/tutorial/update-documents/#update-a-single-document. 
I think it might be safer to just update that single property because really you shouldnt be updating anything else in a get request.

MauricioUyaguari (approved):

PR #2882, opened 2024-06-03
Persistence Component: Introduce ignore case for schema evolution

kumuwu:
#### What type of PR is this?

- Improvement
- Bug Fix

1. Small bug fix: added batch id in IngestModeCaseConverter
2. Introduced new flag to allow ignoring case for schema evolution (always enabled for metadata table)

Fixes #


legend-engine-xts-persistence/legend-engine-xt-persistence-component/legend-engine-xt-persistence-component-physical-plan/src/main/java/org/finos/legend/engine/persistence/components/schemaevolution/SchemaEvolution.java
```diff
@@ -301,14 +304,38 @@ private SchemaDefinition evolveSchemaDefinition(SchemaDefinition schema, Set<Fie
...
-                .filter(f -> !modifiedFieldNames.contains(f.name()))
+                .filter(f -> !containsWithOptionallyIgnoreCase(modifiedFieldNames, f.name()))
                 .collect(Collectors.toList());
 
         evolvedFields.addAll(modifiedFields);
 
         return schema.withFields(evolvedFields);
     }
 
+    private boolean equalsWithOptionallyIgnoreCase(String name1, String name2)
```
prasar-ashutosh, line 315: Logic seems correct. I think the name equals and contains we need to change a bit as they were methods called on String and Collection. Now we are passing them as parameters.
So suggesting these names:
- areEqual(String str1, String str2)
- areEqual(Set<String> set1, Set<String> set2)
- collectionContainsElement(Collection<String> list, String target)

kumuwu: Sure, changed!

epsstan (approved):

PR #3005, opened 2024-08-07
Optimize source tree calculation

siaka-Akash:
#### What type of PR is this?

- Improvement


legend-engine-core/legend-engine-core-pure/legend-engine-pure-code-compiled-core/src/main/resources/core/pure/graphFetch/graphExtension.pure
```diff
@@ -492,7 +491,13 @@ function <<access.private>> meta::pure::graphFetch::enrichSourceTreeNodeForPrope
                );
             },
             {pms:PurePropertyMapping[*] |
-               $pms->map(pm| $pm.transform.expressionSequence->at(0)->evaluateAndDeactivate()->scanProperties(^List<PropertyPathNode>(), [], [], ^ScanConfig(scanClasses=true, explodeMilestonedProperties=false)));
+              let result = $pms->fold({pm, prevRes | let result = $pm.transform.expressionSequence->at(0)->evaluateAndDeactivate()->scanProperties(noDebug(), ^ScanConfig(scanClasses=true, explodeMilestonedProperties=false), $prevRes->last()->getVisitedFunctions());  //$prevRes->last()->getVisitedFunctionMap()
```
aziemchawdhary-gs, line 494: Would it better to expose processExpressionSequence from scanProperties and use that instead of traversing the expression sequence and calling scanProperties on each expression? Is that equivalent to doing this?

aziemchawdhary-gs, line 494: Should the commented code be removed? //$prevRes->last()->getVisitedFunctionMap()?

siaka-Akash: processExpressionSequence() is a private function so could not be accessed from here, Only scanproperteis() is a public API

siaka-Akash: I think we shouldn't expose the internal functionalities to maintain a standard flow where everything goes through scanProperties rather than calling different APIs for different ValueSpec.

AFine-gs (approved):

PR #3008, opened 2024-08-08
repl: make forceExit() explicit

akphi:
This will facilitate easier testing, due to difficulties around setting up JUnit and System.out - see https://stackoverflow.com/questions/309396/how-to-test-methods-that-call-system-exit

Will help with https://github.com/finos/legend-engine-ide-lsp/pull/214

### DIFF
--- a/legend-engine-config/legend-engine-repl/legend-engine-repl-client/src/main/java/org/finos/legend/engine/repl/client/Client.java
+++ b/legend-engine-config/legend-engine-repl/legend-engine-repl-client/src/main/java/org/finos/legend/engine/repl/client/Client.java
@@ -53,15 +53,18 @@ public class Client
     private final LegendInterface legendInterface = new LocalLegendInterface();
     private final Terminal terminal;
     private final LineReader reader;
-    private boolean debug = false;
-    private MutableList<ReplExtension> replExtensions = Lists.mutable.empty();
-    private MutableList<CompleterExtension> completerExtensions = Lists.mutable.empty();
-    private ModelState state;
+    private final MutableList<ReplExtension> replExtensions;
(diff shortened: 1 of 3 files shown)


MauricioUyaguari (approved):

legend-engine-config/legend-engine-repl/legend-engine-repl-client/src/main/java/org/finos/legend/engine/repl/client/Client.java
```diff
@@ -344,14 +347,12 @@ public String getLastCommand(int skip)
...
     }
 
-    // Separate the call to System.exit() to a separate method to make it
-    // easier to mock in test, as handling System.exit() is non-trivial
-    // See https://stackoverflow.com/questions/309396/how-to-test-methods-that-call-system-exit
-    public void exit()
+    // Force exit (using System.exit()) will make sure the REPL quit completely
+    // when "exit" command is called or terminate signal with Ctrl+D/Ctrl+C in invoked
+    // otherwise, after the REPL input loop is exited, it will get stuck at a state where
+    // it still accepts input, but do nothing.
```
kevin-m-knight-gs, line 353: I would suggest making this a proper javadoc.
```suggestion
    /**
     * Force exit using System.exit(0). This will make sure the REPL quits completely
     * when the "exit" command is called or terminate signal with Ctrl+D/Ctrl+C is invoked.
     * Otherwise, after the REPL input loop is exited, and it will get stuck at a state where
     * it still accepts input but does nothing.
     */
```

akphi: I'll make this change in a another PR. Thank you!

PR #3031, opened 2024-08-21
Support to pass home directory to REPL + duckdb to persist on that directory

rafaelbey:

Fixes #


legend-engine-config/legend-engine-repl/legend-engine-repl-client/src/main/java/org/finos/legend/engine/repl/client/Client.java
```diff
@@ -70,12 +71,18 @@ public static void main(String[] args) throws Exception
     public MutableList<Command> commands;
 
     public Client(MutableList<ReplExtension> replExtensions, MutableList<CompleterExtension> completerExtensions, PlanExecutor planExecutor) throws Exception
+    {
+        this(replExtensions, completerExtensions, planExecutor, FileUtils.getUserDirectory().toPath());
```
akphi, line 75: The default directory should not be the user home folder, but scoped to a folder like `.legend` or `.legend/repl` so the REPL somewhat mimick standard behavior of tools like `maven`, `npm`. We already persisted the command history and might potentially persist to disk a lot more things, so we should avoid overcrowding the user home dir.

```suggestion
        this(replExtensions, completerExtensions, planExecutor, FileUtils.getUserDirectory().toPath().resolve(".legend/repl"));
```

legend-engine-config/legend-engine-repl/legend-engine-repl-client/src/main/java/org/finos/legend/engine/repl/client/Client.java
```diff
@@ -289,8 +296,7 @@ private void persistHistory()
 
     public Path getHomeDir()
     {
-        String homeDir = System.getProperty("legend.repl.storageDir");
-        return FileUtils.getUserDirectory().toPath().resolve(homeDir != null ? homeDir : ".legend/repl");
+        return this.homeDirectory;
```
akphi, line 299: I think there's benefit in allowing configuring the home dir via a CLI config flag. Many users will boot the REPL from a JAR - they don't write a custom `main()` and invoke a `Client` with custom path, so I think let's keep the previous behavior, and the System property/CLI flag should have higher precedent than the parameter being passed in the `Client` constructor

```suggestion
        return System.getProperty("legend.repl.homeDir") != null ? System.getProperty("legend.repl.homeDir") : this.homeDirectory;
```

akphi (changes requested):

legend-engine-config/legend-engine-repl/legend-engine-repl-client/src/main/java/org/finos/legend/engine/repl/client/Client.java
rafaelbey: Make sense!  Im makign that change.  Note that you dont need the `.`.

akphi: the `.` will keep it as a hidden folder for `UNIX` environment, I think we should keep it.

akphi (approved):

PR #3040, opened 2024-08-26
Package all flavors of REPL the same way for easy access and debugging

rafaelbey:
#### What type of PR is this?

Improvement

This introduce a new assembly module for REPL that create scripts to run all the flavors of REPL: default, relational, datacube.

This is the first step to unify how to access and run the REPLs, and the next step is to have a single main, and allow users to select the features they want, reducing duplication and confusion on what REPL to use.

### DIFF
--- a/legend-engine-config/legend-engine-repl/legend-engine-repl-data-cube/pom.xml
+++ b/legend-engine-config/legend-engine-repl/legend-engine-repl-data-cube/pom.xml
@@ -257,40 +257,6 @@
                             </execution>
                         </executions>
                     </plugin>
-                    <plugin>
-                        <artifactId>maven-shade-plugin</artifactId>
-                        <executions>
-                            <execution>
-                                <phase>package</phase>
-                                <goals>
-                                    <goal>shade</goal>
-                                </goals>
-                                <configuration>
-                                    <shadedArtifactAttached>true</shadedArtifactAttached>
-                                    <createDependencyReducedPom>false</createDependencyReducedPom>
-                                    <finalName>legend-engine-repl-data-cube-${project.version}-DEV</finalName>
-                                    <filters>
-                                        <filter>
-                                            <artifact>*:*</artifact>
-                                            <excludes>
-                                                <exclude>module-info.class</exclude>
-                                                <exclude>META-INF/**/module-info.class</exclude>
-                                                <exclude>META-INF/*.SF</exclude>
-                                                <exclude>META-INF/*.DSA</exclude>
-                                                <exclude>META-INF/*.RSA</exclude>
-                                            </excludes>
-                                        </filter>
-                                    </filters>
-                                    <transformers>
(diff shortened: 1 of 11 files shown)


legend-engine-config/legend-engine-repl/legend-engine-repl-data-cube/pom.xml
```diff
@@ -74,86 +74,6 @@
...
-                                    <mainClass>org.finos.legend.engine.repl.dataCube.client.DataCubeClient</mainClass>
-                                </transformer>
-                            </transformers>
-                        </configuration>
-                    </execution>
-                </executions>
-            </plugin>
-            <plugin>
-                <groupId>org.apache.maven.plugins</groupId>
-                <artifactId>maven-enforcer-plugin</artifactId>
```
rafaelbey, line 112: @akphi - would like to understand/discuss why we need to override default enforce rules here? 

IMO - It would be best to make sure we have a set of enforce rule all modules follow since will be hard to maintain and defeats the purpose if each module defines it own rules.

legend-engine-config/legend-engine-repl/legend-engine-repl-data-cube/pom.xml
```diff
@@ -344,40 +257,6 @@
                             </execution>
                         </executions>
                     </plugin>
-                    <plugin>
```
akphi, line 347: This is under the DEV profile so it will not be automatically triggered in the normal flows, and I want to keep this for DX. Let's leave it here while the development of DataCube is still active. Thanks!

legend-engine-config/legend-engine-repl/legend-engine-repl-data-cube/pom.xml
akphi: This is a hack to make sure the shaded JAR when run does not complain about missing implementation for SLF4J, let's make sure the assembly we build does not have this issue. Ideally, I think the REPL should boot nicely without lots of Java warnings.

legend-engine-config/legend-engine-repl/legend-engine-repl-data-cube/pom.xml
rafaelbey: Make sense - I reintroduce this on the last commit.

akphi (dismissed):

legend-engine-config/legend-engine-repl/legend-engine-repl-data-cube/pom.xml
rafaelbey: My main goal right now is to avoid creating precedence on overriding the enforce rules.  

I will look into this on the next iteration, to make sure logging is properly configured without the need to bypass the enforce rules. 

With the assembly, we likely have some other options, like including a logback config to route logging to a file, etc rather than No-OP

akphi (approved):

PR #3043, opened 2024-08-26
validate tree pure constraint at runtime instead of compile time

keenkeystrokes:
#### What type of PR is this?
- Improvement

validate dataquality tree pure constraint at runtime instead of compile time

No


rafaelbey:
What is the benefits of doing it at runtime?  From a user perspective,  the later the validations happen, the worse the experience, as it means they dotn get to know what is wrong likely until the are commiting code.

keenkeystrokes:
Makes sense to validate at compile time only. Hence closing the request

PR #3080, opened 2024-09-10
update dataspace analytics result to contain executable info

YannanGao-gs:
#### What type of PR is this?

Improvement

update dataspace analytics result to contain function info and executable info

Fixes #


MauricioUyaguari:
I think we are in the right direction, but we need to be mindful of not restoring the same exact information. We should store only things we need. I think we can keep the template info inside execution context based on your filtering. But for functions, it should be stored globally in the DataSpaceAnalysisResult class. 

Furthermore, one thing I would like you to improve (separate PR is fine). 
We should not store mapped entities for the same mapping twice. 
So we should move that also to the global class. 
Major use of execution context is to have the same mapping but different runtimes. 
So we should only store one graph per mapping.

YannanGao-gs:
> I think we are in the right direction, but we need to be mindful of not restoring the same exact information. We should store only things we need. I think we can keep the template info inside execution context based on your filtering. But for functions, it should be stored globally in the DataSpaceAnalysisResult class.
> 
> Furthermore, one thing I would like you to improve (separate PR is fine). We should not store mapped entities for the same mapping twice. So we should move that also to the global class. Major use of execution context is to have the same mapping but different runtimes. So we should only store one graph per mapping.

For moving `mapped entities `, I am concerned that it will not be backward compatible if I move it to the root as this field has been there since the beginning

legend-engine-config/legend-engine-extensions-collection-generation/pom.xml
```diff
@@ -120,6 +120,11 @@
             <groupId>org.finos.legend.engine</groupId>
             <artifactId>legend-engine-xt-javaPlatformBinding-pure</artifactId>
         </dependency>
+        <dependency>
+            <groupId>org.finos.legend.engine</groupId>
```
MauricioUyaguari, line 124: why is this needed ?

legend-engine-xts-data-space/legend-engine-xt-data-space-generation/src/main/java/org/finos/legend/engine/generation/analytics/DataSpaceAnalyticsHelper.java
```diff
@@ -163,40 +264,26 @@ private static MappingModelCoverageAnalysisResult buildMappingModelCoverageAnaly
...
-                    }
-                    catch (Exception ignored)
-                    {
-
-                    }
-                    return null;
-                }).filter(c -> c != null).collect(Collectors.toList()));
-                Root_meta_analytics_function_modelCoverage_FunctionModelCoverageAnalysisResult functionCoverageAnalysisResult = core_analytics_function_modelCoverage_analytics.Root_meta_analytics_function_modelCoverage_getFunctionModelCoverage_ConcreteFunctionDefinition_MANY__FunctionModelCoverageAnalysisResult_1_(org.eclipse.collections.impl.factory.Lists.mutable.ofAll(functions), pureModel.getExecutionSupport());
+                    DataSpace updatedDataSpaceProtocol = new DataSpace(dataSpaceProtocol);
+                    updatedDataSpaceProtocol.executables = updatedDataSpaceProtocol.executables.stream().filter(ex -> ex instanceof DataSpaceTemplateExecutable).collect(Collectors.toList());
```
MauricioUyaguari, line 273: Dont think we should add dataspace to this mini graph. we should keep to pmcd of mapped elements and not add anything else. The root class should have all the information we need.

legend-engine-xts-data-space/legend-engine-xt-data-space-generation/src/main/java/org/finos/legend/engine/generation/analytics/DataSpaceAnalyticsHelper.java
```diff
@@ -124,6 +124,107 @@ private static DataSpaceExecutableResult buildExecutableResult(ResultType result
...
+        Set<String> qualifiedMappingPath = dataSpaceProtocol.executionContexts.stream().map(e -> e.mapping.path).collect(Collectors.toSet());
+        Set<String> qualifiedRuntimePath = dataSpaceProtocol.executionContexts.stream().map(e -> e.defaultRuntime.path).collect(Collectors.toSet());
+        if (dataSpaceProtocol.executables != null)
+        {
+            dataSpaceExecutableAnalysisInfos = dataSpaceProtocol.executables.stream()
+                    .filter(ex -> ex instanceof DataSpacePackageableElementExecutable)
+                    .map(executable ->
+                    {
+                        String path = ((DataSpacePackageableElementExecutable) executable).executable.path;
+                        try
```
MauricioUyaguari, line 139: dont understand the need for the try/catch herem can't we just check the type of `PackageableElement`.

legend-engine-xts-data-space/legend-engine-xt-data-space-generation/src/main/java/org/finos/legend/engine/generation/analytics/model/DataSpaceExecutableAnalysisInfo.java
```diff
@@ -0,0 +1,41 @@
...
+import org.finos.legend.engine.protocol.pure.v1.model.valueSpecification.raw.Lambda;
+
+public class DataSpaceExecutableAnalysisInfo
+{
+    public String id;
+    public String executionContextKey;
+    public String title;
+    public String description;
+    public Lambda query;
+    public String type;
```
MauricioUyaguari, line 26: lets do `PackageableElementPointer` and fill out path with type.

legend-engine-xts-data-space/legend-engine-xt-data-space-generation/src/main/java/org/finos/legend/engine/generation/analytics/model/DataSpaceExecutableAnalysisInfo.java
```diff
@@ -0,0 +1,41 @@
...
+
+import org.finos.legend.engine.protocol.pure.v1.model.valueSpecification.raw.Lambda;
+
+public class DataSpaceExecutableAnalysisInfo
+{
+    public String id;
+    public String executionContextKey;
+    public String title;
+    public String description;
+    public Lambda query;
```
MauricioUyaguari, line 25: could we store query string here instead ?

MauricioUyaguari (changes requested):

legend-engine-xts-data-space/legend-engine-xt-data-space-generation/src/main/java/org/finos/legend/engine/generation/analytics/DataSpaceAnalyticsHelper.java
```diff
@@ -124,6 +124,107 @@ private static DataSpaceExecutableResult buildExecutableResult(ResultType result
...
+                                            info.description = executable.description;
+                                            info.type = PackageableElementType.SERVICE.name();
+                                            info.query = ((PureMultiExecution) ((Service) v1Service.get()).execution).func;
+                                            return info;
+                                        }
+                                    }
+                                }
+                            }
+                        }
+                        catch (Exception ignored)
```
MauricioUyaguari, line 190: will we even ever reach here if the above does not fail i.e not a service ?

legend-engine-config/legend-engine-extensions-collection-generation/pom.xml
YannanGao-gs: This is to pass "public void testMetadataDeserializer()" at TestExtensions.java

legend-engine-xts-data-space/legend-engine-xt-data-space-generation/src/main/java/org/finos/legend/engine/generation/analytics/DataSpaceAnalyticsHelper.java
```diff
@@ -124,6 +124,107 @@ private static DataSpaceExecutableResult buildExecutableResult(ResultType result
...
+                    .map(executable ->
+                    {
+                        String path = ((DataSpacePackageableElementExecutable) executable).executable.path;
+                        try
+                        {
+                            PackageableElement service = pureModel.getPackageableElement(path);
+                            Optional<org.finos.legend.engine.protocol.pure.v1.model.packageableElement.PackageableElement> v1Service =  pureModelContextData.getElements().stream().filter(e -> e.getPath().equals(path)).findAny();
+                            if (v1Service.isPresent() && v1Service.get() instanceof Service)
+                            {
+                                if (service instanceof Root_meta_legend_service_metamodel_Service)
```
MauricioUyaguari, line 145: looking at the metamodel seems to make this a bit hard to read. could we just read the v1Service properties. 
Most of them are 1-1.

legend-engine-xts-data-space/legend-engine-xt-data-space-generation/src/main/java/org/finos/legend/engine/generation/analytics/DataSpaceAnalyticsHelper.java
YannanGao-gs: I made a mistake when resolving your comment "I think we are in the right direction, but we need to be mindful of not restoring the same exact information. We should store only things we need. I think we can keep the template info inside execution context based on your filtering. But for functions, it should be stored globally in the DataSpaceAnalysisResult class.".


I thought you wanted to keep the inline one as what it is instead of writing an info for it. 

I have added the inline one back. pleas recheck

legend-engine-xts-data-space/legend-engine-xt-data-space-generation/src/main/java/org/finos/legend/engine/generation/analytics/model/DataSpaceExecutableAnalysisInfo.java
YannanGao-gs: I have created a new enum for the type for writing 'inline'

legend-engine-xts-data-space/legend-engine-xt-data-space-generation/src/main/java/org/finos/legend/engine/generation/analytics/DataSpaceAnalyticsHelper.java
YannanGao-gs: the path used in the function pointer is not the actual function path, you will get an except if you try to get the packageelement using this path 
instead, you should generate the correct function path first, then get the v1-function

legend-engine-xts-data-space/legend-engine-xt-data-space-generation/src/main/java/org/finos/legend/engine/generation/analytics/DataSpaceAnalyticsHelper.java
YannanGao-gs: I found it's hard to get the runtime from  public Runtime runtime;

legend-engine-xts-data-space/legend-engine-xt-data-space-generation/src/main/java/org/finos/legend/engine/generation/analytics/model/DataSpaceExecutableAnalysisInfo.java
YannanGao-gs: ok if this is a string, then we could just use public List<DataSpaceExecutableAnalysisResult> executables = Collections.emptyList();

legend-engine-xts-data-space/legend-engine-xt-data-space-generation/src/main/java/org/finos/legend/engine/generation/analytics/model/DataSpaceExecutableAnalysisInfo.java
MauricioUyaguari: just reuse `PackageableElementPointer` no reason to create a new enum. 
if its inline the property will be null

legend-engine-xts-data-space/legend-engine-xt-data-space-generation/src/main/java/org/finos/legend/engine/generation/analytics/model/DataSpaceExecutableAnalysisInfo.java
MauricioUyaguari: not sure what you mean here. why would it become an empty list? 
we compose the query grammar form Lambda

legend-engine-xts-data-space/legend-engine-xt-data-space-generation/src/main/java/org/finos/legend/engine/generation/analytics/model/DataSpaceExecutableAnalysisInfo.java
MauricioUyaguari: so instead of `String type` -> `PackageableElementPointer element`
where we fill out both path and type

MauricioUyaguari (approved):

PR #3090, opened 2024-09-16
Persistence Component: Fix schema evolution for bitemp temp tables

kumuwu:
#### What type of PR is this?

- Bug Fix

Schema evolution was not working for bitemporal delta (source specifies from only), as temp tables got created with old schema. Fixed by correcting the sequence of SQL generation (schema evolution first).

Fixes #


prasar-ashutosh (approved):
Changes looks good

gs-ssh16 (approved):

PR #3173, opened 2024-10-11
datacube: use query server for storage

akphi:


legend-engine-application-query/src/main/java/org/finos/legend/engine/application/query/model/DataCubeQuery.java
```diff
@@ -0,0 +1,38 @@
...
+    public Map<String, ?> query;
+    public Map<String, ?> source;
+    public Map<String, ?> executionContext;
+
+    public Long lastUpdatedAt;
+    public Long createdAt;
+    public Long lastOpenAt;
+
+    // We make it clear that we only allow a single owner
+    public String owner;
```
MauricioUyaguari, line 37: could we make this a class structure like activator owner, in case we want to improve this. 
we may want to do another ownership type than just a single one.

legend-engine-application-query/src/main/java/org/finos/legend/engine/application/query/model/DataCubeQuery.java
```diff
@@ -0,0 +1,38 @@
...
+import java.util.Map;
+
+@JsonIgnoreProperties(ignoreUnknown = true)
+public class DataCubeQuery
+{
+    public String id;
+    public String name;
+    public String description;
+
+    public Map<String, ?> query;
```
MauricioUyaguari, line 28: shouldnt these be modeled. 
The only thing i would argue we may not want to model is the actual grid config for data cube

MauricioUyaguari (changes requested):

legend-engine-application-query/src/main/java/org/finos/legend/engine/application/query/model/DataCubeQuery.java
akphi: we could, but that's not the bit I want to focus on in this phase. Doing it as `FunctionActivator.Ownership` will entail many things such as coming up with a framework to make filter like `isMyQuery` more complicated to implement (consider case when we have deployment ID and the user comes to us with a kerberos)

Perhaps, we don't even do ownership at all?

legend-engine-application-query/src/main/java/org/finos/legend/engine/application/query/model/DataCubeQuery.java
akphi: Do we need this shape here for any reason? I don't think we do, else, we need to move `DataCubeQuery` shape into a common module (out of repl-data-cube). That's work unnecessary at the moment that we can do anytime when the need is there.

MauricioUyaguari (approved):

PR #3212, opened 2024-10-29
Review legend compile pmcd

hbalasu2:
#### What type of PR is this?

Improvement

 Adding support and exposing compile PMCD method for testing protocols from Pure light IDE


gs-jp1 (approved):
One minor thing is to adjust the copyright year
We should also consider if further test cases should be added.

PR #3218, opened 2024-10-30
Add query timeout to database connection

Paarth002:
#### What type of PR is this?
- Feature Improvement

Add query timeout to database connection. 

Add query timeout to database connection


rafaelbey:
I dont think this belong to the connection.  Seems a something that belongs to an execution context (ie ->from(runtime, mappigng, context) 

As not every query have the same execution profile requirements and doing it like this likely will lead to duplicated connections/runtimes just so users can leverage different timeouts.

gs-ssh16:
> I dont think this belong to the connection. Seems a something that belongs to an execution context (ie ->from(runtime, mappigng, context)
> 
> As not every query have the same execution profile requirements and doing it like this likely will lead to duplicated connections/runtimes just so users can leverage different timeouts.

We want to add it in connection so that timeout is enforced for all queries on top of connection (through Query/HostedService/...). It helps with cost management (avoiding long running queries)

gs-ssh16 (approved):

PR #3258, opened 2024-11-20
Persistence Component: Fix NPE in BulkLoad stats

kumuwu:
#### What type of PR is this?

- Bug Fix

The new BulkLoad stats: bytes scanned and incoming record count return null results when the file is empty. Adding null checks around.


prasar-ashutosh (approved):
LGTM

gs-ssh16 (approved):

PR #3274, opened 2024-11-25
[Legend SQL] Support second format of transaction isolation query

gs-manvig:
#### What type of PR is this?
- Improvement

##### Transaction isolation
Some SQL clients such as Hikari and ODBC driver check transaction isolation level during startup.
The current approach uses in-memory H2 to respond to those queries, and works fine for Hikari/JDBC clients.
However, H2 doesn't support the ODBC query format (`SHOW transaction_isolation`).

This PR introduces a custom handler for transaction isolation queries, so that the ODBC query format can be supported too.

##### `pg_type`
Adds support for system schema `pg_type`, which is queried during ODBC driver startup.
Routed to in-memory H2.

No


shtirlets:
looks good

aormerod-gs (approved):

PR #3295, opened 2024-12-09
Dataquality - Relation validation enhancements for optional from clause, query limit, unit tests

keenkeystrokes:
#### What type of PR is this?
- Improvement

1. make from clause optional
2. query limit - add support for limit expression in generated lambda
3. Pure unit tests
4. API enhancements to run only user query

No


rafaelbey:
Do we need to have an explicit property for runtime?  Our goal is to move towards usage of `from` function, and for newer grammars, adding this is taking steps on using deprecated patterns.

keenkeystrokes:
> Do we need to have an explicit property for runtime? Our goal is to move towards usage of `from` function, and for newer grammars, adding this is taking steps on using deprecated patterns.

hi Rafael here we are ultimately using the separate property for runtime to construct the `from` function in the final lambda instead of having the user to define the `from` function on which he is performing the quality checks.

legend-engine-xts-dataquality/legend-engine-xt-dataquality-pure/src/main/resources/core_dataquality/generation/dataquality.pure
```diff
@@ -534,3 +545,28 @@ function meta::external::dataquality::buildRelationFilterExpression(dqRelationVa
...
+                            importGroup=system::imports::coreImport)->evaluateAndDeactivate();
+}
+
+function meta::external::dataquality::buildFromRuntimeExpression(f: FunctionExpression[1], runtime: Runtime[1], returnType:GenericType[1]):FunctionExpression[1]
+{
+  ^SimpleFunctionExpression(func=from_T_m__Runtime_1__T_m_,
+                            functionName=from_T_m__Runtime_1__T_m_.name,
+                            importGroup=system::imports::coreImport,
+                            genericType=$returnType,
+                            multiplicity=ZeroMany,
```
gs-jp1, line 564: lets double check that ZeroMany is correct here. I would expect multiplicity to be the same as the paramValue given the function signature

legend-engine-xts-dataquality/legend-engine-xt-dataquality-generation/src/main/java/org/finos/legend/engine/generation/dataquality/DataQualityLambdaGenerator.java
```diff
@@ -39,46 +39,58 @@
...
+            return generateRelationValidationLambda(pureModel, (Root_meta_external_dataquality_DataQualityRelationValidation) packageableElement, validationName, runQuery, queryLimit);
         }
         throw new EngineException("Unsupported Dataquality element! " + packageableElement.getClass().getSimpleName(), ExceptionCategory.USER_EXECUTION_ERROR);
     }
 
+    private static LambdaFunction generateModelConstraintLambda(PureModel pureModel, Integer queryLimit, Root_meta_external_dataquality_DataQuality packageableElement)
+    {
+        if (Objects.isNull(queryLimit))
+        {
+            return core_dataquality_generation_dataquality.Root_meta_external_dataquality_generateDataQualityQuery_DataQuality_1__Integer_MANY__LambdaFunction_1_(packageableElement, Lists.immutable.empty(), pureModel.getExecutionSupport());
```
gs-jp1, line 80: lets change limit to 0..1

gs-jp1 (approved):
approved, but follow up required.

PR #3321, opened 2024-12-23
Persistence Component: Fix staging filters saved as String + Fix new stats for Bulk Load coming as String

kumuwu:
#### What type of PR is this?

- Bug Fix

1. In multi-dataset cases, staging filters are saved as strings (hence quoted) instead of numbers. Fixed the placeholder handlings that caused it.
2. Newly added stats for bulk load: bytes scanned and incoming records are coming as strings. Added casting to save them as numbers.

Fixes #


prasar-ashutosh (approved):
LGTM

PR #3337, opened 2025-01-07
Review jp relgraphfetch

gs-jp1:
#### What type of PR is this?
Bug Fix

For embedded mappings we want to filter out any rows that have no values associated with any of the selected embedded properties. This is to effectively mimic a left outer join which would result in [0] if no join match instead of populating an empty object which could then have false defects.


finos-admin:

finos-admin:

PR #3343, opened 2025-01-09
Add batch 2 of compiled state integrity tests

sean-borland-gs:
#### What type of PR is this?

- Improvement

Adds another set of Compiled State Integrity Tests

--

No user-facing changes.


legend-engine-xts-dataquality/legend-engine-xt-dataquality-pure/pom.xml
```diff
@@ -248,6 +248,32 @@
             <artifactId>jackson-core</artifactId>
             <scope>test</scope>
         </dependency>
+        <dependency>
+            <groupId>junit</groupId>
```
gs-jp1, line 252: should scope be test? same for other pom files

PR #3432, opened 2025-02-26
Add expected Snowflake failures prior to enabling PCT

jake-kim1:
Add expected Snowflake failures prior to enabling PCT


rafaelbey (approved):

akphi:
```sql
-- Simplified Example:
WITH column_renames AS (
  SELECT
    column_name,
    CASE
      WHEN column_name LIKE '%_%' THEN REPLACE(column_name, '_', '')  -- Remove underscore if it exists
      ELSE column_name  -- Keep the original name if no underscore
    END AS new_column_name
  FROM INFORMATION_SCHEMA.COLUMNS
  WHERE table_name = 'your_table_name' AND table_schema = 'your_schema_name'
)
SELECT
  /*  Construct the SELECT statement dynamically */
  LISTAGG(CASE
        WHEN cr.column_name != cr.new_column_name THEN '"' || cr.column_name || '" AS "' || cr.new_column_name || '"'
        ELSE '"' || cr.column_name || '"'  -- No rename needed
        END, ', ') WITHIN GROUP (ORDER BY cr.column_name) as select_statement
FROM column_renames cr;
```

PR #3441, opened 2025-03-03
Persistence Component: Fix derivation of partition spec list in multi-dataset ingestor

kumuwu:
#### What type of PR is this?

- Bug Fix

Delay derivation of partition spec list in multi-dataset ingestor to from init() time to ingest() time. This is because the staging table to query from and the batch id needed to filter only exists in ingest() time.

Fixes #


prasar-ashutosh (approved):
LGTM

gs-ssh16 (approved):

PR #3482, opened 2025-03-20
update preEval to skip parseDate only for relational

AFine-gs:


davidharte-gs (approved):

rafaelbey:
I dont think that is how preeval works since in most cases we pass all extensions to preeval (`$extensions.routerExtensions().shouldStopPreeval`).  Thru the extensions it will collect all the match functions, hence relational or not, it will be part of the universe, and the behavior likely will be the same with or without this change.

PR #3495, opened 2025-03-24
Build Tabular Function while building databases from Studio

janeenyamak1:
this will build tabular function when users try to "build database" on the connection

Fixes #


janeenyamak1:
created new PR here https://github.com/finos/legend-engine/pull/3524

PR #3553, opened 2025-04-10
fix relation mapping routing bug, better error messages

gs-rpant1729:
fix relation mapping routing bug, better error messages

#### What type of PR is this?

Bug Fix

Fix niche bug when handling union operations during relation function mapping routing. This is a temporary fix, dropping union handling altogether (unions with relation mapping isn't supported anyway). Strategic fix would be to only route set impls when actually used in the expression being routed rather than routing all sets in scope.
Also, add user-friendly error messages about unsupported functionality.

Fixes #


finos-admin:

PR #3631, opened 2025-05-03
FlatData: option for FixedWidth final column to read to EoL

aziemchawdhary-gs:
FlatData: option for FixedWidth final column to read to EoL

Allow column specification for final column to accept EOL to tell driver to read the rest of the line. 
EOL only accepted for final column to handle files with variable lengths on final field. 

Example schema, where the final column will read from index 3 upto the EoL (in this case the default EoL character since no new line character is specified in the schema definition)
```
section default: FixedWidth
{
  scope.untilEof;
  
  Record 
  {
     Title {1:2} : STRING;
     Surname: {3:EOL} : STRING;
  }
}
```

### DIFF
--- a/legend-engine-xts-flatdata/legend-engine-xt-flatdata-model/src/test/java/org/finos/legend/engine/external/format/flatdata/grammar/driver/TestFixedWidth.java
+++ b/legend-engine-xts-flatdata/legend-engine-xt-flatdata-model/src/test/java/org/finos/legend/engine/external/format/flatdata/grammar/driver/TestFixedWidth.java
@@ -115,6 +115,97 @@ public class TestFixedWidth extends AbstractDriverTest
         Assert.assertEquals("MD", p3.TITLE);
     }
 
+
+    @Test
+    public void canReadErrorFreeDynamicSizeFinalColumn()
+    {
+        FlatData flatData = parseFlatData("section default: FixedWidth\n" +
+                "{\n" +
+                "  scope.untilEof;\n" +
+                "\n" +
+                "  Record\n" +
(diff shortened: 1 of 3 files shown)


legend-engine-xts-flatdata/legend-engine-xt-flatdata-model/src/test/java/org/finos/legend/engine/external/format/flatdata/grammar/driver/TestFixedWidth.java
```diff
@@ -115,6 +118,124 @@ public void canReadErrorFree()
...
+        Assert.assertEquals(27L, p3.AGE);
+        Assert.assertEquals(1, p3.MASTER);
+        Assert.assertEquals("1.3", p3.WEIGHT);
+        Assert.assertEquals("Karl", p3.NAME);
+        Assert.assertEquals(LocalDate.parse("2011-11-26"), p3.EMPLOYED_DATE);
+        Assert.assertEquals("Master", p3.TITLE);
+    }
+
+    @Test
+    public void checkValidationErrorIncorrectFinalColumn()
```
rafaelbey, line 215: Similar to this, can we add a test where EOL appears before the last field to validate the exception behavior?

aziemchawdhary-gs: Yes - added a test for this

rafaelbey (approved):

PR #3660, opened 2025-05-12
support for backward string compatibility for date and timestamp in avro uploads in sf

sriraman-s:

Earlier version was supporting string uploads into avro and timestamp column via implicit reference of SF. 
added the epoch feature with the capability tag so that it is backward compatible.

no


kumuwu:
LGTM

prasar-ashutosh (approved):
LGTM

prasar-ashutosh (approved):
LGTM

gs-ssh16 (approved):

PR #3696, opened 2025-05-28
Bugfix: Handle delete partitions above maxPartitionSpecFilters

samkit-gs:
#### What type of PR is this?
- Bug Fix

When delete partitions are used and total number of unique partitions are above maxPartitionSpec limit, then delete partitions are getting ignored and partitions are not deleted from delete partition list.

This change fixes the above mentioned bug by moving delete partitions handling to a separate SQL execution.

Yes


prasar-ashutosh (approved):
LGTM

gs-ssh16 (approved):

PR #3710, opened 2025-05-29
Bugfix: Support for case-insensitive metadata handling for Snowflake persistence

samkit-gs:
#### What type of PR is this?
- Bug Fix

When using case-insensitive metadata on Snowflake, some parts of logic of persistence module break to understand the changes. Like idempotency check or partition column validation. This is because existing code expects exact case to be pulled from Snowflake, but when QUOTED_IDENTIFIER_IGNORE_CASE session parameter is set on Snowflake then it Snowflake keeps all column/table metadata in upper case. And this results in failure in finding right metadata.

This PR changes the breaking few places to ignore the case for metadata.

No


kumuwu:
LGTM

gs-ssh16 (approved):

PR #3713, opened 2025-05-29
Add catalogsync iceberg property

kanc23:
**Improvement**

Adding catalog sync property to include  open catalog integration while creating iceberg tables in Snowflake
https://docs.snowflake.com/en/sql-reference/sql/create-iceberg-table-snowflake#syntax


kumuwu:
LGTM

PR #3714, opened 2025-05-30
Added SnowflakeM2MUdf plan executor module

vijaylather:
#### What type of PR is this?
Improvement

Added the module for executing Snowflake M2M udf given a execution plan and the list of inputs. This module will be used for implementing SnowflakeM2MUdf.


legend-engine-xts-snowflake/legend-engine-xt-snowflake-m2mudf-plan-executor/pom.xml
```diff
@@ -0,0 +1,173 @@
...
+
+    <build>
+        <plugins>
+            <plugin>
+                <artifactId>maven-shade-plugin</artifactId>
+                <executions>
+                    <execution>
+                        <phase>package</phase>
+                        <goals>
+                            <goal>shade</goal>
```
rafaelbey, line 38: why we need to shade?

PR #3716, opened 2025-05-30
handle quoted columns in join conditions correctly

gs-rpant1729:
#### What type of PR is this?

Bug fix

Handle quoted columns in join conditions correctly

Fixes #


gs-rpant1729:
Closed in favor of #3718

PR #3717, opened 2025-05-30
Fix minor bugs in PCT relational library

jake-kim1:
* Avoid reprocessNew() call if instance value's type is of Relation
* We previously defaulted to .all() when calling reprocessNew(), but the fix is to cast it toOne() based on given args


jake-kim1:
Duplicate of https://github.com/finos/legend-engine/pull/3734

PR #3758, opened 2025-06-09
Adding SQL Dialect Tests for Relational Elements

mani-mangold:
#### What type of PR is this?

Improvement

Enhance the existing SDT suite.

Adding SDT on sql elements like select, Join, Union, Groupby.

Fixes #


mani-mangold:
Opened a new PR with corrections

PR #3764, opened 2025-06-10
update Snowflake/Databricks percentile PCT assert

gs-rpant1729:
#### What type of PR is this?

Bug fix

We recently updated the Pure percentile native implementation to align with the one supported by a majority of the relational DBs. Need to update the assert accordingly for failing Snowflake/Databricks PCT.

Fixes #


rafaelbey:
Databricks will fail too

PR #3778, opened 2025-06-13
Use CompileLegendGrammar function for snowflake tests

ADV-718:
#### What type of PR is this?
Improvement


legend-engine-xts-snowflake/legend-engine-xt-snowflake-pure-test/src/main/resources/core_snowflake_test/core_snowflakeapp/showcase/showcaseElements.txt
```diff
@@ -0,0 +1,205 @@
+
```
rafaelbey, line 1: We need to standardize these, and a file type we been agreeing on is to use `.legend`.

Yasirmod17: A subsequent PR will change this

Yasirmod17 (approved):

PR #3822, opened 2025-06-30
Add isNumeric Check for DirectionalIndex Processor

mani-mangold:
#### What type of PR is this?

Bug Fix

Adding isNumeric check for input in DirectionalIndex Processor to handle non-numeric inputs.

Fixes #


aziemchawdhary-gs (changes requested):
We need to add tests - the original change has been reverted so please fix this and add tests for this code path as part of recommitting the original set of changes.

PR #3887, opened 2025-07-21
datqaquality metamodel - deprecate mustEndWithRuntime constraint

keenkeystrokes:
#### What type of PR is this?
- Improvement

removes the mustEndWithRuntime constraint to relax the primary query on which validations are being written. For executable queries users must mention the runtime using the `from` function, however this is not mandated by spec 

Fixes #

No


finos-admin:

PR #3988, opened 2025-08-19
Delegate project function that takes in TableTDS to one with TabularDataSet

jake-kim1:
Delegate project function that takes in TableTDS to one with TabularDataSet


jake-kim1:
Duplicate of https://github.com/finos/legend-engine/pull/4000

PR #4051, opened 2025-09-08
Fix mod pure to sql

rafaelbey:
Currently, the Pure `mod` and `rem` (remainder) functions are mapped to SQL `mod` function, and this is incorrect.  In Pure, `mod` follows the Euclidean division, hence is always positive.  While `rem` is what in many languages the `%` does, including SQL.

For positive number dividend, both yield same value, but for negative that is not the case.  This PR ensure that mod account for negative dividend, ensuring the correct output.


legend-engine-core/legend-engine-core-shared/legend-engine-shared-vault/legend-engine-shared-vault-aws/src/main/java/org/finos/legend/engine/shared/core/vault/aws/AWSVaultImplementation.java
```diff
@@ -76,9 +76,7 @@ private String getValueFromAwsVault(String key)
             val = object.get(key);
             if (val == null)
             {
-                throw new RuntimeException("The key " + key + " can't be found. Available ones: " + Lists.mutable.fromStream(StreamSupport.stream(
-                        Spliterators.spliteratorUnknownSize(object.fieldNames(), Spliterator.ORDERED),
-                        false)).makeString(", "));
+                return null;
```
rafaelbey, line 79: Be able to run SF and Databricks together without failures.

pierredebelen (approved):

PR #4119, opened 2025-09-24
Add New Error Type to Pure Model

vale-paiz:

Fixes #


rafaelbey:
What is this exactly? Looks odd to be added to pure model the way it was added. 

Pure model compiles while instantiating the object (ie during constructor) 

If this should fail compilation, it should not be these added method. At that point, it's best to have this as a utility that takes a PureModel, and provides whatever change instructions message needs to be provided.

PR #4771, opened 2026-05-22
Added changes for attribute metadata in Query  and Pagination for the search API 

OmGupta-GS2038:

Fixes #


legend-engine-application-query/src/main/java/org/finos/legend/engine/application/query/api/QueryStoreManager.java
```diff
@@ -303,17 +306,58 @@ public List<Query> searchQueries(QuerySearchSpecification searchSpecification, S
...
+            builder.withSkip(offset);
+            // Fetch one extra to determine if there's a next page
+            builder.withLimit(pageSize + 1);
+
+            List<Query> queries = getQueryDao().find(filters.isEmpty() ? EMPTY_FILTER : Filters.and(filters), false, builder.build())
+                    .map(this::convertFromStoredQuery)
+                    .sorted(Comparator.comparing(query -> query.owner != null && query.owner.equals(currentUser) ? 0 : 1))
+                    .collect(Collectors.toList());
+
+            boolean hasNextPage = queries.size() > pageSize;
```
MauricioUyaguari, line 335: Items can be lost or duplicated across pages ([QueryStoreManager.java:317-330](https://github.com/finos/legend-engine/pull/4771/files))
The paginated path fetches pageSize + 1 rows (the extra row is a sentinel for hasNextPage), but then applies the in-memory "current user's queries first" sort to all pageSize+1 rows before trimming:
````
.sorted(Comparator.comparing(query -> query.owner != null && query.owner.equals(currentUser) ? 0 : 1))
````
queries = queries.subList(0, pageSize);
If the sentinel row is owned by the current user, it gets sorted into the returned page, displacing a legitimate row. The displaced row is lost forever — the next cursor is offset + pageSize, which skips past it — and the sentinel row is duplicated on the next page. The trim must happen on the Mongo-ordered list before the owner sort. The pagination tests don't catch this because they all use a single owner, which makes the owner-comparator a no-op; a test with mixed owners would expose it.

File: legend-engine-application-query/src/main/java/org/finos/legend/engine/application/query/model/ApplicationStoredQuery.java [hunk 2]
```diff
@@ -41,6 +41,7 @@ public class ApplicationStoredQuery implements StoredVersionedAssetContent<Strin
     public List<StereotypePtr> stereotypes;
     public List<QueryParameterValue> defaultParameterValues;
     public Map<String, ?> gridConfig;
+    public List<AttributeMetadata> attributeMetadata;
```
MauricioUyaguari, line 44: do we have a roundtrip serialization test for this new property?

legend-engine-application-query/src/main/java/org/finos/legend/engine/application/query/api/QueryStoreManager.java
```diff
@@ -303,17 +306,58 @@ public List<Query> searchQueries(QuerySearchSpecification searchSpecification, S
...
         }
+        else
+        {
+            builder.sortDesc(getSortByField(QuerySearchSortBy.SORT_BY_VIEW));
+        }
         builder.withExcludeFields(EXCLUDED_PROJECTION_FIELDS);
-        if (searchSpecification.limit != null && searchSpecification.limit <= 0)
+
+        // Check if pagination is requested (spec is a PaginatedQuerySearchSpecification with pageSize specified)
+        boolean isPaginated = searchSpecification instanceof PaginatedQuerySearchSpecification
```
MauricioUyaguari, line 316: we should ensure the limit the pageSize > 0 or throw a QueryException

MauricioUyaguari (changes requested):

MauricioUyaguari:
attributeMetadata not excluded from search projection. EXCLUDED_PROJECTION_FIELDS deliberately strips heavy fields (content, gridConfig, defaultParameterValues…) so search returns "light" queries — attributeMetadata should almost certainly be added there, otherwise every search hit ships the full metadata list.

legend-engine-application-query/src/main/java/org/finos/legend/engine/application/query/api/QueryStoreManager.java
```diff
@@ -303,17 +306,58 @@ public List<Query> searchQueries(QuerySearchSpecification searchSpecification, S
         {
             builder.sortDesc(getSortByField(searchSpecification.sortByOption));
         }
+        else
+        {
+            builder.sortDesc(getSortByField(QuerySearchSortBy.SORT_BY_VIEW));
```
MauricioUyaguari, line 311: . Pagination ordering is unstable
Default sort is lastOpenAt descending — a field that changes when anyone opens a query. Between page fetches the ordering shifts, causing skipped/duplicated rows. Worse, queries that were never opened have no lastOpenAt, and Mongo's skip/limit over tied values is non-deterministic without a unique tiebreaker. At minimum add a secondary sort on _id (or another unique field) so offset pagination is repeatable.

MauricioUyaguari (changes requested):

finos-admin:

finos-admin:

PR #4809, opened 2026-06-03
support to H2 connection property database_to_upper.  For testables, …

rafaelbey:
…set this to false to respect the casing of the original tables

Fixes #


mimi-tam (approved):
lgtm

MauricioUyaguari:
@rafaelbey 
this change is not backward compatible. any way we can make it backward compatible ? 
causing studio roundtrip tests to fail as new property is now always included in the protocol: 
https://github.com/finos/legend-studio/actions/runs/27101690508/job/79983459700

PR #4819, opened 2026-06-04
Add uniqueRelationName function and update SQL execution

Yasirmod17:
Add uniqueRelationName function and update SQL execution

Fixes #


rafaelbey (approved):

gs-rpant1729 (commented):
Do we also need similar handling for Views?

PR #4836, opened 2026-06-10
Refactor relation tests to assert on TDS and update copyright doc

ahallgs:
#### What type of PR is this?
- Improvement

Moves copyright doc for new files to 2026 reflecting when development was completed. Updates relation tests to assert on TDS string for platform consistency.

Fixes #

### DIFF
--- a/legend-engine-xts-relationalStore/legend-engine-xt-relationalStore-generation/legend-engine-xt-relationalStore-pure/legend-engine-xt-relationalStore-core-pure/src/main/resources/core_relational/relational/tests/mapping/relation/tests.pure
+++ b/legend-engine-xts-relationalStore/legend-engine-xt-relationalStore-generation/legend-engine-xt-relationalStore-pure/legend-engine-xt-relationalStore-core-pure/src/main/resources/core_relational/relational/tests/mapping/relation/tests.pure
@@ -31,14 +31,16 @@ function <<test.Test, test.AlloyOnly>> meta::relational::tests::mapping::relatio
     meta::external::store::relational::tests::testRuntime(meta::relational::tests::mapping::relation::testDB),
     meta::relational::extension::relationalExtensions()
   );
-  assertEquals('#TDS\n' +
-               '   name,age\n' +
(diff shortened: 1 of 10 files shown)


legend-engine-xts-relationalStore/legend-engine-xt-relationalStore-generation/legend-engine-xt-relationalStore-pure/legend-engine-xt-relationalStore-core-pure/src/main/resources/core_relational/relational/tests/mapping/relation/tests.pure
```diff
@@ -20,463 +20,853 @@ import meta::relational::metamodel::*;
...
-
-function <<meta::pure::profiles::test.Test>> meta::relational::tests::mapping::relation::testSimpleMappingWithAssociation():Boolean[1]
-{
-  testTds(|Person.all()->project(~[name:x|$x.firstName, firmName:x|$x.firm.legalName])->cast(@TabularDataSet),
-          meta::relational::tests::mapping::relation::SimpleMapping, 
-          meta::relational::tests::mapping::relation::testDB,
-          ['David | Firm D', 'Fabrice | Firm C', 'John | Firm X', 'Oliver | Firm C']
-        );
-}
-
```
gs-jp1, line 52: instead of having the expectation as a string, could this be just be toString on the TDS. That way we get the compile on the assertion.
Ideally, we wouldn't even need the toString at all and we have equals method implemented on TDS.
Totally appreciate this comment applies to many existing pct tests

ahallgs: As this test suite is still relatively small and new, will update with the refactoring to use toString on expected and actual so we can be consistent moving forward

PR #4837, opened 2026-06-10
Improve EMIT handling of atomic tests without test suites in Testables

kevin-m-knight-gs:
Improve EMIT handling of atomic tests without test suites in Testables. Previously, a fictitious test suite session would be created. Now, however, the atomic tests are run directly through the usual methods.


mimi-tam (approved):
lgtm

PR #4844, opened 2026-06-12
Support sub-aggregation queries (correlated subqueries with GROUP BY) for Relation Function Class Mappings

Vithesh-Reddy:
Relation Function Class Mappings currently lack primary key (PK) support, which is required for union mappings, graph fetch, and sub-aggregation queries (e.g., correlated subqueries with GROUP BY). This work adds end-to-end PK support across the compiler stack: Grammar, Parser, Composer, Compiler (prerequisite elements pass + third pass), and Pure inference logic.
Primary keys can be explicitly declared via ~primaryKey syntax or auto-inferred where possible — for example, when the relation function body uses a direct table accessor (#>{my::db.personTable}#), the PK is inferred from the underlying table's primary key. An extension-based architecture (RelationElementAccessorExtension) enables PK inference for SQL expressions (#SQL{...}#) and other store-specific relation sources without coupling modules.
API/Grammar Design:
Single PK: ~primaryKey: ID Multiple PKs: ~primaryKey: [ID, FIRMID] Auto-inferred (no declaration needed): ~func my::personFunc_Relation_1 where the function body resolves to a table accessor


gs-rpant1729 (dismissed):

gs-rpant1729:
LGTM

aziemchawdhary-gs (approved):

PR #4879, opened 2026-06-24
Add null check when pure multiplicity is nullable

dramala:
Today null equality check is added only in filter. This change would add it everywhere e.g in join also


finos-admin:

PR #4900, opened 2026-07-03
Null-safe equality: nullSafeEqual/nullSafeNotEqual dynaFunctions, unified semantics across all relational dialects

rafaelbey:
## Why

**The originating requirement is function activators.** A Snowflake UDTF (and function activators generally) materializes generated SQL as a standalone artifact deployed into the database — there is no engine at execution time, so the FreeMarker machinery that historically compensated for null semantics at plan-execution (e.g. `optionalVarPlaceHolderOperationSelector` swapping SQL fragments based on parameter presence) simply does not exist in that world. The platform's null-safety must be **pushed down into the SQL itself**. Working on UDTF SQL generation surfaced that null-safe equality was not expressible statically — and pulling that thread exposed broader correctness gaps affecting all execution modes.

Pure in-memory semantics treat `[] == []` as true and `[] != 'x'` as true, but relational SQL generation historically diverged: null handling for equality was hand-rolled in **five separate places** (the router's `processEquals` FreeMarker selector branches, a projection-only OR-expansion, union bridge joins, an H2-only dialect-translation compensation, and the legacy renderer's `orNull` compensation), each with slightly different guards and gaps. The visible symptoms: projected boolean comparisons leaked SQL `NULL` instead of `false`, `!=` against optional parameters silently dropped null rows, and behavior differed store-to-store for identical Pure.

This PR centralizes null-safe equality behind two new generation-internal dynaFunctions — `nullSafeEqual` / `nullSafeNotEqual` — with per-dialect translations, and converges every hand-rolled site onto them. The equality path now emits **no FreeMarker operations at all** (optional parameters render as plain placeholders that substitute to a SQL `null` literal when absent — valid in static SQL); the selector machinery survives only for the enum-parameter path (scoped follow-up) and for previously serialized plans, which embed their own template functions. **Value:** null-safety that is expressible in deployable SQL (UDTFs, activators), one place defining the semantics (PCT-parity with Pure), each dialect declaring its best rendering once, and downstream code (union, dataquality recon) deleting its own null-handling entirely.

**Risk containment — `Feature.LEGACY_SQL_NULL_UNSAFE_EQUALS`.** Because this changes the router's `==` translation for every relational query, the PR ships a per-query escape hatch: a new execution-plan feature flag that routes `processEquals` back to the exact pre-PR behavior (both arms live side-by-side as private functions). The name says SQL deliberately: Pure has no null — the "unsafe" semantics only exist at the SQL level. Settable in-query via `->withFeatureFlags(...)` or programmatically via `addFlagToContext(...)`; default (flag absent) is the new null-safe behavior. **The SQL API is the first consumer by design, not as a fallback**: the SQL transpiler maps SQL `=` to Pure `==`, so null-safe equality would have broken SQL three-valued logic (`NULL = NULL` must yield NULL, not true). Both SQL plan-generation paths (Pure `getPlan` and the Java relation path in `SQLExecutor`) now set the flag, restoring genuine SQL semantics for SQL-API users while Pure queries keep Pure semantics — the three Postgres parity cases previously documented as accepted divergences (`null_equals_null`, `null_not_equals`, `case_simple_null_compare`) are back to PASS.

## How to review (suggested order)

1. **Tests first, by design** — commit 1 adds PCT tests + SQL-string locks capturing *current* behavior before any code change; later commits flip those locks, so each behavioral change is visible as an expectation diff. Reviewing commit-by-commit shows exactly what changed and what stayed byte-identical.
2. **The core** (`Introduce nullSafeEqual/nullSafeNotEqual…`): registration touchpoints (DynaFunctionRegistry, type inference, postgres-model bridge, boolean-op list), default portable renderings, and `processEquals`' branch collapse into a declarative multi-if. Key semantic notes below.
3. **The convergences**: union `nullSafeEquals` (30 lines → 1), the FreeMarker selector branch collapse, filter-context unification (`convertEqualFromFilter`/`processEqualFromFilter`/`processNotEqual` become thin redirects), dataquality recon.
4. **The feature flag** (`Add LEGACY_SQL_NULL_UNSAFE_EQUALS feature flag…`): enum value + `State` wiring mirroring `PUSH_DOWN_ENUM_TRANSFORM`; `processEquals` split into `nullSafeEqualsOperation` / `legacyNullUnsafeEqualsOperation` (the latter is the pre-PR code verbatim); the two SQL transpiler plan-gen call sites; tests proving both flag states (plain `=` + FreeMarker selector restored under the flag). Also folds a small dedup: the `VARIANT_TYPE_AS_INPUT` check in internalize plan generation now uses `contextHasFlag` instead of a hand-rolled context walk (which missed flags set directly on an `ExecutionOptionContext`).
5. **Dialect specifics**: native `IS [NOT] DISTINCT FROM` (H2, DuckDB, Postgres/Aurora, BigQuery, Databricks, Trino/Athena, Presto), `<=>` (MemSQL, SparkSQL), `equal_null()` (Snowflake — terse, and the UDTF target dialect), `coalesce`-based (Spanner — its PG interface rejects DistinctExpr; ClickHouse — version-inconsistent NULL boolean logic, and per-generation-state: join planner requires the plain expansion in ON clauses), full boolean-value emulation (SqlServer — T-SQL separates predicates from values; case-when wraps over Alias and bare boolean operations emit `cast(1 as bit)`/`cast(0 as bit)` so results round-trip through JDBC as real booleans, un-excluding 72 PCT tests beyond this PR’s own — truth tables, comparisons, `in`, `between`; the BIT-cast approach solves the `String→Boolean` result-mapping ceiling that the equivalent Oracle/Sybase case-when idiom still hits), portable OR-expansion default (everything else).

## Semantic decisions reviewers should validate

- **Filter (WHERE) column-vs-column**: byte-equivalent to the previous per-renderer compensations — locked by SQL-string tests on both stacks; no result-set changes.
- **One-sided comparisons keep plain `=` in filters** (a nullable column vs a bound value filters identically either way; preserves index-friendliness); null-safe form applies where the difference is observable (projections, both-nullable, optional parameters).
- **Literal-empty operands** (`[] == x`) now degenerate to `x IS NULL` (and `[] == []` to constant true) — matching Pure semantics; previously emitted `x = null` (never true).
- **One intentional behavior fix**: `!=` against a bound optional parameter now *includes* rows whose column is null (Pure: `[] != 'x'` is true); previously two-valued NOT semantics dropped them. Visible in the TestServiceRunner expectation diff.
- **SQL API keeps three-valued SQL semantics** via the flag (see above): raw SQL `=`/`<>` through the SQL API translate with legacy null-unsafe equality; known uncovered paths are the `/lambda` endpoint and the sql-expression compiler extension (the flag rides the ExecutionContext, not the generated lambda) — acceptable because neither generates a plan.
- **Defaults use only `not`/`=`/`is null`** — no `<>` vs `!=` token choice baked into dialect-agnostic code; `nullSafeNotEqual` needs three disjuncts because `NOT(<equal-expansion>)` yields NULL (not TRUE) for one-side-null.

## Risks and mitigations

- **Blast radius**: any consumer bitten by the new semantics can revert per query with `->withFeatureFlags(Feature.LEGACY_SQL_NULL_UNSAFE_EQUALS)` — no redeploy or global switch needed.
- **SQL text changes broadly** (filters/joins/projections on nullable operands render differently per dialect). Mitigated by: every adapter verified — locally against real engines (H2, DuckDB, ClickHouse, Spanner emulator, mssql container, python container, both server suites) and via this PR's own `pct-cloud-test` rounds (Postgres, Snowflake, Oracle, Trino, MemSQL, Databricks).
- **Join-plan performance**: null-safe forms on join keys can affect sargability on some optimizers; one-sided filter comparisons deliberately keep plain `=` to limit exposure.
- **Deliberately NOT changed** (no PCT to verify): SybaseASE/SybaseIQ/DB2 retain their existing (already-broken-for-projected-booleans) behavior; the enum optional-parameter FreeMarker path is untouched (its modernization is a scoped follow-up — and a prerequisite for enum parameters in activator SQL).

## Verification summary

All four new PCT tests pass natively on H2, DuckDB, Postgres, Snowflake, Oracle, Trino, MemSQL, Spanner, ClickHouse and SqlServer; the only remaining exclusions are genuine capability gaps (JAVA binding TDS translation, Deephaven `project`). Local: Test_Pure_Relational 2684, H2 PCT 490, DuckDB full 1210, Spanner 353, ClickHouse 353, SqlServer full module 1209, server suites 2554, TestServiceRunner 56, dataquality 87, python reverse PCT 327. Flag + SQL API round: focused flag tests 5, sql-http-api 66 (incl. SqlExecuteTest), sql-pure transpiler 155, Snowflake m2mudf 3, json-runtime 7, and **Postgres parity 2352/2352** — the three null-semantics cases confirmed fixed by the runner's own FIX-DETECTED check before their expectations were flipped to PASS.


rafaelbey:
### Heads-up for reviewers: intentional SQL null-semantics regression

This PR makes relational equality **null-safe** (`nullSafeEqual` / `nullSafeNotEqual`). A side effect surfaces through the SQL API, and I've **deliberately accepted it for now** rather than block the PR. Flagging it so the flipped parity expectations in `null_semantics.yaml` / `case_expressions.yaml` don't look like an oversight.

**What changed.** Via the Postgres SQL API, `NULL`-in-equality is now evaluated with Pure value semantics instead of SQL three-valued logic. Observed against real Postgres (`TestPostgresParity`, from `target/parity-report.json`):

| Query | Postgres | This branch (Relation) |
|-------|----------|------------------------|
| `SELECT (NULL = NULL)` | `null` | `true` |
| `SELECT (NULL <> 1)` | `null` | `true` |
| `CASE age WHEN NULL THEN 'null' ELSE 'not null'` | `not null` | `null` |

**Root cause.** The SQL binding lowers a SQL `NULL` literal to a Pure empty (`[]`) — `core_external_query_sql/binding/fromPure/fromPure.pure` (`createNormalizedCast(iv([]), String)`). The reworked `processEquals` (`core_relational/.../pureToSQLQuery/pureToSQLQuery.pure`) then routes `[]`-operand comparisons through the same null-aware handling this PR introduces, so `[] == []` is `true` and `col == NULL` degenerates to an `IS NULL` check.

**Why accept it here.** For `NULL = NULL` and `NULL <> 1` the **TDS** path already returned `true` before this PR (`expected_tds_status: FAIL` predates it); this change simply makes the Relation path consistent with TDS. The `CASE` case is a genuine new Relation-only divergence. Given the PR's scope is null-safe *relational* equality, I've marked the three cases `expected_rel_status: FAIL` (with inline comments) so parity stays honest, rather than silently masking or expanding scope.

**Follow-up (tracked separately).** Restore strict three-valued SQL `=`/`<>` in the SQL API by distinguishing a SQL `NULL` literal from a Pure empty at the binding layer — without breaking Pure's `[] == []` → true or `col == []` → `IS NULL`. That's the correct long-term fix; it's out of scope for this PR.

AFine-gs (approved):

PR #4901, opened 2026-07-05
Render equality-based CASE/if as DECODE where the dialect supports it

rafaelbey:
## What

Adds an `isDecodeSupported` capability to `DbExtension`. When a dialect declares it, an equality-based `case` — and Pure `if` / multi-`if` — switch on a single expression against non-null literal searches is rendered as the terser `DECODE(...)` instead of `CASE WHEN`.

- Convertible branch conditions: `equal`, `nullSafeEqual`, `in`, `isNull`, `isEmpty`.
- An `sqlNull` else is omitted (DECODE returns NULL on no-match); any other else becomes the DECODE default.
- A nested `if` / multi-`if` chain switching on one expression is flattened into a single `DECODE`.
- Non-convertible cases (match-dispatch, boolean `and`/`or`, non-literal or NULL searches) fall through to the existing CASE rendering unchanged.

Enabled on **Snowflake, MemSQL/SingleStore, Oracle, and Databricks** (all support `DECODE` with `NULL = NULL` matching).

## Why

`DECODE(col, a, x, b, y, else)` is materially terser than the repeated-column `CASE WHEN col = a THEN x ...` — most visible for enum push-down and multi-branch `if`. It's a pure readability/terseness change; the SQL is semantically equivalent.

## Notes

- **Stacked on #4879** (which introduces `nullSafeEqual`, referenced here). This branch will be rebased onto `master` once #4879 merges; until then the diff includes #4879's commits.
- Verified locally: Snowflake / MemSQL / Databricks / Oracle Pure suites green; RelationFunctions PCT green across H2 / DuckDB / compiled / interpreted / Java-binding / Python / SQL-reverse. Only two goldens flipped (`active='Y'` boolean case → `decode` on Snowflake + MemSQL), both regenerated.
- Opened primarily for **CI visibility** of the cloud-profile PCT (Snowflake/MemSQL/Oracle/Databricks), which exercises the DECODE rendering end-to-end.


finos-admin:

finos-admin:

PR #4950, opened 2026-07-17
Improve TDS csv parser to better align string processing to pure itself, quoting with single quote, and escaping with backslash.

rafaelbey:
Also, bugFix keys() and values() pure to sql to ensure we cast as map.


mimi-tam (approved):
lgtm

PR #5030, opened 2026-08-05
enclose values in quotes for snowflake csv upload

dramala:
#### What type of PR is this?

- Bug Fix

When streaming a result set into a snowflake table we now enclose values with ".

This solves a bug where if a value enda with \\, it will escape the , in the csv and interpret as a literal , instead of a delimeter


rafaelbey:
How we handle cases where data can be variant/json?  Do we need to escape the "?  Or this is not a problem we should fix on this PR?

rafaelbey (approved):

PR #5051, opened 2026-08-09
Fix release workflow to handle large artefacts

aziemchawdhary-gs:
# Fix `413 Payload Too Large` when publishing to Maven Central

## Problem

The `release` workflow has been failing at the upload step with a bare
`<center><h1>413 Payload Too Large</h1></center>`. The response is HTML rather
than the Publisher Portal API's usual JSON because the rejection happens at
nginx, before the API ever sees the request.

The Portal caps a single bundle at **1 GiB**. A full legend-engine release is
**~1.74 GB** across **600 coordinates / 5,610 files**, uploaded as one zip.
Presto hit the same wall at 4.3GB ([prestodb/presto#25856](https://github.com/prestodb/presto/issues/25856)).

## Approach

Split the upload into multiple bundles, and make the release recoverable first,
because splitting without that is more dangerous than the bug.

**The published artifact set is unchanged.** Dropping the obvious weight is
tempting and wrong. The 971MB `legend-engine-server-http-server:shaded` jar,
the `-tests.jar`s, sources and javadoc are all on Central today (verified
against `repo1.maven.org` for 4.138.0), and Central is immutable, so removing
any of them silently breaks consumers that resolve those coordinates. `.par`
and `-jar-with-dependencies.jar` artifacts — sometimes assumed to be the bulk
here — do not exist in the local repository at all; excluding them is a no-op.
Exclusions are therefore opt-in via `CENTRAL_EXCLUDE_GLOBS`, empty by default.

Splitting happens only on whole `<artifactId>/<version>` directories. A jar
must never be separated from its pom, signature and checksums, or the bundle
validates while leaving a permanently broken artifact on Central.

## Changes

Three commits, ordered so that all behaviour change is isolated in the last one.

### 1. `.github/scripts/central-publish.sh` — Portal API driver

The workflow drove the API inline with three bugs that only surface during a
release:

| | |
|---|---|
| `publishingType=AUTOMATIC` | With N bundles, if bundle 5 fails validation, bundles 1–4 are already irreversibly on Central. Now `USER_MANAGED`, so every bundle reaches `VALIDATED` before any is published. |
| `printf "$u:$p" \| base64` | `base64` wraps at 76 columns, putting a newline inside the `Authorization` header for a token over 57 bytes; and the credentials were used as a printf *format string*, so a `%` in the password corrupted them. Now `printf '%s:%s' "$u" "$p" \| base64 -w0`. |
| Unbounded status poll | No deadline, and a transient API error made `jq -r '.deploymentState'` yield the string `"null"`, so it spun against that until the 6-hour job limit. Now bounded by `CENTRAL_POLL_TIMEOUT`, with unreadable responses retried rather than parsed as a state. |

`FAILED` deployments are deliberately never dropped — Sonatype support needs
their files to diagnose the failure.

### 2. `.github/scripts/stage-and-split.sh` — staging, splitting, reporting

Bin-packs coordinates into bundles under a configurable target (1e9 bytes by
default), with a 1 GiB backstop check on each produced zip.

Because a coordinate cannot be split, a single one over the target is a **hard
error** rather than a silently oversized bundle, and one over 80% warns.

Also narrows the local-repo cleanup: it previously deleted every `*.xml`, which
would have swept up a genuine `.xml` artifact. It now targets the specific
bookkeeping filenames (`_remote.repositories`, `maven-metadata-*.xml`,
`*.lastUpdated`, `resolver-status.properties`).

### 3. `.github/workflows/release.yml` — wiring

Sequence is now: stage and split → upload every bundle → wait for **all** to
reach `VALIDATED` → only then publish any. A failed or cancelled run drops the
deployments still at `VALIDATED` and leaves `FAILED` ones in place.

Also adds a `dryRun` input, fixes the `DEVELOPMENT_ID` typo, and pins
`actions/checkout` to `v6` to match every other job.

## Observability

This failure was hard to diagnose because nothing recorded what was uploaded.
Every run now writes, **before** anything is uploaded:

- `bundle-report.md` — totals, per-bundle size as a percentage of the Portal
  limit, and the 20 largest coordinates, flagged at ≥80% of the target
- `bundles.tsv` — bundle / files / zip bytes / staged bytes / sha256
- `bundle-contents.tsv` — which coordinate landed in which bundle
- `coordinates.tsv` — every coordinate and its size

The upload step adds a **bundle → deployment-ID** table. That mapping was
previously impossible to reconstruct: when Sonatype says "deployment X failed",
you can now say which zip, which sha256, and which artifacts were in it.

All of it goes to the job summary and uploads as an artifact under
`if: always()` with 30-day retention, so a failed release keeps its evidence
after the runner is gone.

## Verification

Run against the real artifact tree (600 coordinates, 1.74GB):

| | |
|---|---|
| Bundles | 3 — 202.1 / **895.7** / 465.2 MB |
| Largest | **83.4%** of the 1 GiB limit |
| Coverage | 5,610 staged files = 5,610 bundled files, 0 duplicated, 0 coordinates spanning bundles |

- `test-stage-and-split.sh` — **29 assertions, ALL PASS.** Offline, synthetic
  repository, seconds. The two that matter most are that no file is duplicated
  across bundles and that the bundles cover every staged file exactly once — a
  split that drops a `.pom` or `.asc` produces bundles that validate
  individually but leave a permanently broken artifact on Central.
- `test-central-publish.sh` — **15 assertions, ALL PASS.** Drives the script
  against a local mock of the Portal API that decodes the `Authorization`
  header and asserts on `publishingType`, so the auth and polling fixes are
  verifiable without credentials or network. Mutation-checked: reintroducing
  the original `base64` line makes it fail with HTTP 401 on the first call.

Not done locally: `shellcheck` and `actionlint` were unavailable in the
environment. `bash -n` was run on all four scripts and the workflow YAML was
parsed. Worth running both in CI.

## Follow-ups, not in this PR

**The shaded jar is the real deadline.** `legend-engine-server-http-server` is
a single 968MB coordinate — **96.8% of the packing target**. A coordinate
cannot be split across bundles, so once it passes 1000MB the release hard-fails
and no amount of splitting helps. The fix is to shrink it, stop publishing it
after a downstream audit (legend-sdlc, legend-studio, legend-depot), or request
a bundle-size exemption from central-support@sonatype.com. The report now
prints this on every run rather than leaving it to be rediscovered.

**Stop bundling from `~/.m2` at all.** The `build` job runs `mvn install`, not
`deploy`, so the bundle is the whole local-repo subtree rather than the
deployable set. Deploying to a file-based staging repo with
`-DaltDeploymentRepository` would honour per-module `maven.deploy.skip` and
drop local-repo bookkeeping automatically. Worth noting that **no module
currently sets `maven.deploy.skip`**, so this is a smaller win than it sounds
until modules are audited.

**`central-publishing-maven-plugin` is dead config.** It is declared in the
root `pom.xml` with `autoPublish=true` and `waitMaxTime=3600`, but the workflow
never runs `deploy`, so it never executes. Left alone here to keep the diff
focused, but it should be removed or actually used.

**Release-count metric.** Many small deployments inflate the tracked "release
count"; Portal rate limiting begins 2026-10-01. Worth checking FINOS's standing
at https://central.sonatype.com/publishing/usage before a permanently split
release process becomes the norm — another argument for the follow-ups above.

## Reviewing this

```bash
bash .github/scripts/test-stage-and-split.sh   # expect ALL PASS, offline, seconds
bash .github/scripts/test-central-publish.sh   # expect ALL PASS, no network
```

To see real numbers without a fresh build, against any local `~/.m2` that has a
built version in it:

```bash
.github/scripts/stage-and-split.sh ~/.m2/repository/org/finos/legend/engine <version> /tmp/out
cat /tmp/out/central-publishing/bundle-report.md
```

Note that `dryRun` still runs the full `build` job, including
`release:prepare`, which tags and bumps versions. It rehearses the *publish*
path, not the release as a whole — it is not a no-op on `master`.

### DIFF
--- a/.github/scripts/central-publish.sh
+++ b/.github/scripts/central-publish.sh
@@ -1,0 +1,186 @@
+#!/usr/bin/env bash
+#
+# Copyright 2026 Goldman Sachs
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#      http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+#
+# Central Publisher Portal API driver.
+#
+#   central-publish.sh upload <bundle.zip> <deployment-name>  -> prints deployment id
+#   central-publish.sh await  <deployment-id> [target-state]  -> default VALIDATED
+#   central-publish.sh state  <deployment-id>                 -> prints current state
+#   central-publish.sh publish <deployment-id>
+#   central-publish.sh drop    <deployment-id>
+#
+# Uploads are always USER_MANAGED. With AUTOMATIC, a release split across N
+# bundles has no safe failure mode: if bundle 5 fails validation, bundles 1-4
+# are already irreversibly on Central. USER_MANAGED lets every bundle reach
+# VALIDATED before any of them is published, and lets a failed run drop the
+# validated-but-unpublished ones.
+#
+# Env:
+#   CI_DEPLOY_USERNAME, CI_DEPLOY_PASSWORD  Portal token (required)
+#   CENTRAL_API_BASE                        override for testing
+#   CENTRAL_POLL_INTERVAL                   seconds between status polls (60)
+#   CENTRAL_POLL_TIMEOUT                    seconds before giving up (3600)
+
+set -euo pipefail
+
+CENTRAL_API_BASE=${CENTRAL_API_BASE:-https://central.sonatype.com/api/v1/publisher}
+CENTRAL_POLL_INTERVAL=${CENTRAL_POLL_INTERVAL:-60}
+CENTRAL_POLL_TIMEOUT=${CENTRAL_POLL_TIMEOUT:-3600}
+
+: "${CI_DEPLOY_USERNAME:?CI_DEPLOY_USERNAME is not set}"
+: "${CI_DEPLOY_PASSWORD:?CI_DEPLOY_PASSWORD is not set}"
+
+# -w0 because base64 wraps at 76 columns and a token longer than that would put
+# a newline inside the Authorization header. '%s:%s' because the credentials
+# are data, not a printf format string -- a '%' in the password corrupts it.
+auth_token() {
+  printf '%s:%s' "$CI_DEPLOY_USERNAME" "$CI_DEPLOY_PASSWORD" | base64 -w0
+}
+
+urlencode() {
+  jq -rn --arg v "$1" '$v|@uri'
+}
+
+cmd_upload() {
+  local bundle=$1 name=$2 body code
+  [ -f "$bundle" ] || { echo "ERROR: no such bundle: $bundle" >&2; exit 1; }
+  body=$(mktemp)
+  code=$(curl -sS -o "$body" -w '%{http_code}' -X POST \
+    -H "Authorization: Bearer $(auth_token)" \
+    --form "bundle=@$bundle" \
+    "$CENTRAL_API_BASE/upload?publishingType=USER_MANAGED&name=$(urlencode "$name")")
+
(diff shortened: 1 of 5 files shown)


rafaelbey (approved):

finos-admin:
Successfully created backport PR for `release-legend-release-4.98`:
- #5052

PR #5074, opened 2026-08-13
Document every PCT function in legend-engine

rafaelbey:
Builds on finos/legend-pure#1305, which brought that repository's platform corpus to 309 of 309
documented signatures and added the `'''…'''` documentation form. legend-engine owns roughly as many
PCT functions and had almost none documented to the same standard: a signature published either
nothing or a single unstructured sentence, where the equivalent platform function published a full
entry.

These strings reach end users through `FUNCTIONS_<repository>.json`, so the gap was visible outside
the repository.

## What changes

**Coverage goes from 25 of 291 signatures to 291 of 291**, across all seven engine Pure
repositories. 150 signatures had no documentation at all and 116 carried a legacy `doc.doc` tagged
value; an element may not hold both forms, so each tagged value is deleted in the edit that replaces
it. `PCT.grammarDoc` and `PCT.grammarCharacters` are different tags and are retained.

| Repository | Signatures |
|---|---|
| `core_functions_relation` | 117 |
| `core_functions_standard` | 114 |
| `core_functions_unclassified` | 37 |
| `core_functions_variant` | 9 |
| `core_dataquality` | 7 |
| `core_scenario_quant` | 6 |
| `core` | 1 |

**101 `<<PCT.test>>` cases are documented too**, concentrated in the two `composition.pure` files
where a test's purpose is least obvious from its body.

**A written standard**, `docs/pct/pct-documentation.md`, adapted from the legend-pure one: the
template, the formatting rules, the overload rule, the cross-reference rule, and how to verify.
`conventions.md`, `purefunction-howto.md` and `native-howto.md` are repointed off the `doc.doc` form.

**A page on the composition tests**, `docs/pct/composition-tests.md`. Those tests compose two or more
functions and are the hardest to read; the page groups the relation suite into eight themes and
records the engine divergences the manifests pin.

**A browsable reference**, `PCT_Report_Functions.html` — a javadoc-style companion to the
compatibility matrix, rendered from the same `pct-docs.json` that `GeneratePCTFiles` already emits,
so there is no new build wiring. One page per function: compatibility per adapter, then every
overload with its documentation rendered, with `See also` references linking to the functions they
name. Screenshots are in the comment below. The last commit publishes the SQL parity report from a
docs job of its own.

## What the documentation says

Behaviour and examples are taken from the PCT assertions in each file rather than from the
signatures. Signatures say what the types are; the assertions say what actually happens:

- a `_range` window frame is decided by ordering value rather than position, so tied rows always
  share a frame — over one test partition both `o=1` rows read `20`, where the equivalent `rows`
  frame gives `10` and then `20`
- `pivot` returns `Relation<Any>` and its generated column names contain literal quote characters,
  so the cast that must follow doubles them
- `lpad` truncates as well as pads, and a multi-character fill discards the string being padded — a
  defect in the underlying `pad`, recorded as such
- `aggregate` keeps only the aggregated columns, unlike `extend`, which repeats the same value across
  every original row
- `lateral` drops a row whose function returns nothing, as an inner join does
- `in` cannot be negated against a relation on any relational store — SQL's `NOT IN` goes `UNKNOWN`
  on a null — so the entry carries the `exists` rewrite the error message suggests. The in-memory
  engines do answer it, so documenting it as simply working would have misled anyone running against
  a database.

## Two things reviewers should know

**Composition test documentation does not publish.** `FunctionsGeneration.addPCTTestToFunctionsDB`
attributes a test to a function by source id, and a `composition.pure` declares no
`<<PCT.function>>`, so the lookup misses and the test is dropped along with its documentation.
Confirmed against the generated report. The literals are still worth having for anyone reading the
source, and `docs/pct/composition-tests.md` carries what would otherwise be lost.

**Two modules emit no PCT report at all.** `legend-engine-xt-dataquality-pure` and
`legend-engine-pure-code-compiled-core` declare no `generate-pct-functions` execution, so their eight
signatures are documented but never published. Wiring dataquality up also needs
`dataquality_relation_helper.pure` split first, since `FunctionsGeneration` permits one PCT function
name per source file. Both are recorded in the standard rather than fixed here.

## Audit

After the first pass, the review threads on legend-pure#1305 were replayed against this corpus. That
found real defects, fixed in later commits:

- **25 relation signatures had never been converted** — every overload of `groupBy`, `pivot`, `rows`
  and `range`, plus `aggregate`, `reduce`, `lateral` and `flatten`. The gap survived verification
  because `Signature.documentation` is populated by the tagged value and the literal alike, so the
  report read *95 of 95 documented* either way. The standard now says a populated `documentation`
  field is not evidence of conversion, and gives the grep that is. This is the same failure the
  reviewer caught on `times`/`plus`/`minus`/`divide`.
- **Three files read back-to-front** — `least`, `stdDevPopulation` and `stdDevSample` each opened
  with the overload whose documentation pointed forward to text below it. The REPL shows the first
  signature that has any, so the reader met the delta first. Reordered, as legend-pure did for
  `greaterThanEqual` and `elementToPath`; those diffs are exact permutations, with no body changed.
- **18 signatures had no runnable example**, against the standard's own rule.

Also checked and clean: no singular prose on a non-`[1]` return (the `fold` class), no predicate
described as failing, no over-claimed direction or monotonicity, and every `See also` cross-reference
resolves against the union PCT index apart from the intra-module ones in `core_dataquality`, which
emits no report and were checked by hand.

## Rebase

Rebased onto master. Four collisions, all resolved in favour of master's newer work:

- **#5085** documented the six regex functions first, and corrected the "full POSIX ERE" claim this
  branch had introduced — the implementation is `java.util.regex`. Master's version is taken
  wholesale; this branch adds nothing to those files.
- **#5104** dropped `functionType.NormalizeRequiredFunction` from `relation::eval`; master's
  declaration is kept with this branch's literal.
- **#5136** collapsed `between`'s `StrictDate` and `DateTime` overloads into one `Date` overload. The
  file keeps master's shape, its `doc.doc` becomes a literal, and the summary's overload count and
  cross-reference are corrected to match.
- Master added composition tests where this branch added literals; both are kept.

Master also brought ten new PCT functions still on a legacy `doc.doc` — the four `joinStrings`
overloads and the `FuncColSpec` families on `groupBy` and `aggregate` — documented here so the
coverage claim stays true.

## Verification

Documentation is parse-time sugar over a tagged value, so no PCT result changes.

- `mvn clean install` green on the `-pure` modules, compiled-state integrity and PCT report tests
  passing
- `mvn checkstyle:check@verify` clean
- 291 of 291 signatures carry a `'''` literal; no legacy `doc.doc` remains on any PCT declaration

### DIFF
--- a/legend-engine-core/legend-engine-core-pure/legend-engine-pure-code-functions-standard/legend-engine-pure-functions-standard-pure/src/main/resources/core_functions_standard/boolean/operation/xor.pure
+++ b/legend-engine-core/legend-engine-core-pure/legend-engine-pure-code-functions-standard/legend-engine-pure-functions-standard-pure/src/main/resources/core_functions_standard/boolean/operation/xor.pure
@@ -14,6 +14,27 @@
 
 import meta::pure::test::pct::*;
 
+'''
+Exclusive or: true when exactly one of the two operands is true.
+
+Differs from `||` in rejecting the case where both are true, which makes it the test for "one or the
+other, but not both".
+
+**Parameters**
+- `first` — the first operand.
+- `second` — the second operand.
+
+**Returns** `true` when exactly one operand is `true`.
+
+**Examples**
+```pure
+xor(true, false)   // true
+xor(true, true)    // false, unlike ||
+xor(false, false)  // false
+```
+
+**See also** `meta::pure::functions::collection::or(Boolean[*]):Boolean[1]`
+'''
 function <<PCT.function>> meta::pure::functions::boolean::xor(first:Boolean[1], second:Boolean[1]):Boolean[1]
 {
    and(or($first, $second), or(!$first, !$second));
--- a/legend-engine-core/legend-engine-core-pure/legend-engine-pure-code-functions-standard/legend-engine-pure-functions-standard-pure/src/main/resources/core_functions_standard/boolean/operation/xor.pure
+++ b/legend-engine-core/legend-engine-core-pure/legend-engine-pure-code-functions-standard/legend-engine-pure-functions-standard-pure/src/main/resources/core_functions_standard/boolean/operation/xor.pure
@@ -29,6 +29,7 @@ other, but not both".
 **Examples**
 ```pure
 xor(true, false)   // true
+xor(false, true)   // true
 xor(true, true)    // false, unlike ||
 xor(false, false)  // false
 ```
--- a/legend-engine-core/legend-engine-core-pure/legend-engine-pure-code-functions-standard/legend-engine-pure-functions-standard-pure/src/main/resources/core_functions_standard/collection/or.pure
+++ b/legend-engine-core/legend-engine-core-pure/legend-engine-pure-code-functions-standard/legend-engine-pure-functions-standard-pure/src/main/resources/core_functions_standard/collection/or.pure
@@ -14,11 +14,38 @@
 
 import meta::pure::test::pct::*;
 
+'''
+Whether any value in a collection is true.
+
+The collection form of `||`. **An empty collection gives `false`**, the identity for disjunction and
+the mirror of `and(Boolean[*]):Boolean[1]` returning `true`.
+
+This is an aggregate: it is what a relational `groupBy` or a window aggregation calls to fold a
+boolean column.
+
+**Parameters**
+- `vals` — the values.
+
+**Returns** `true` when any value is `true`; `false` for an empty collection.
+
+**Examples**
+```pure
+or([false, true])    // true
+or([false, false])   // false
+```
+
+**See also** `and(Boolean[*]):Boolean[1]`,
+`meta::pure::functions::boolean::xor(Boolean[1], Boolean[1]):Boolean[1]`
+'''
 function <<PCT.function>> meta::pure::functions::collection::or(vals:Boolean[*]):Boolean[1]
 {
    $vals->fold({i,a|$i || $a}, false)
 }
 
+'''
+As `or(Boolean[*]):Boolean[1]`, for a collection known to hold at least one value, so the
+empty-collection case cannot arise.
+'''
 function <<PCT.function>> meta::pure::functions::collection::or(vals:Boolean[1..*]):Boolean[1]
 {
    $vals->fold({i,a|$i || $a}, false)
--- a/legend-engine-core/legend-engine-core-pure/legend-engine-pure-code-functions-standard/legend-engine-pure-functions-standard-pure/src/main/resources/core_functions_standard/collection/or.pure
+++ b/legend-engine-core/legend-engine-core-pure/legend-engine-pure-code-functions-standard/legend-engine-pure-functions-standard-pure/src/main/resources/core_functions_standard/collection/or.pure
@@ -32,6 +32,7 @@ boolean column.
 ```pure
 or([false, true])    // true
 or([false, false])   // false
+or([])               // false
 ```
 
 **See also** `and(Boolean[*]):Boolean[1]`,
--- a/legend-engine-core/legend-engine-core-pure/legend-engine-pure-code-functions-standard/legend-engine-pure-functions-standard-pure/src/main/resources/core_functions_standard/date/operation/timeBucket.pure
+++ b/legend-engine-core/legend-engine-core-pure/legend-engine-pure-code-functions-standard/legend-engine-pure-functions-standard-pure/src/main/resources/core_functions_standard/date/operation/timeBucket.pure
@@ -14,35 +14,65 @@
 
 import meta::pure::test::pct::*;
 
+'''
+Rounds a timestamp down to the start of the bucket it falls in.
+
+The way to group events into fixed intervals — five-minute bars, hourly counts, monthly totals —
+without computing boundaries by hand. The result is the **start** of the containing bucket, so all
+timestamps in the same interval collapse to the same value and can be grouped on directly.
+
+**Buckets are measured from the Unix epoch**, not from the earliest value in the data, so the same
+timestamp always lands in the same bucket regardless of what else is in the relation. That is what
+makes buckets comparable across queries, but it also means a bucket boundary need not align with a
+natural calendar start when `quantity` is greater than 1.
+
+**Parameters**
+- `date` — the timestamp to bucket.
+- `quantity` — how many units wide the bucket is.
+- `unit` — the unit, as a `DurationUnit`.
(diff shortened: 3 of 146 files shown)


rafaelbey:
## Preview — PCT function reference

The last commit adds `PCT_Report_Functions.html`, a javadoc-style companion to the compatibility matrix. It renders from the same `pct-docs.json` that `GeneratePCTFiles` already emits, so there is no new build wiring — the existing `pct-report` job copies all of `target/pct` to the Legend docs site.

Screenshots are from a local run, which sees 16 adapters; CI additionally has the cloud stores.

**Overview** — modules, and each adapter's overall pass rate.

Overview

**Browse by package** — module → package → function, with overload and test counts.

Package

**One page per function** — compatibility per adapter, then every overload with its documentation rendered. `See also` references link to the functions they name.

Function

**Test cases** — each test with its own documentation and qualifiers, one column per adapter.

Test cases

**Failures** — clicking any `×` shows that adapter's error message.

Failures

Also reachable on a running server at `/api/pct/functions`. Unlike the matrix, it counts every reported test — no qualifier, including `unsupportedFeature`, is treated specially.

legend-engine-core/legend-engine-core-pure/legend-engine-pure-code-functions-standard/legend-engine-pure-functions-standard-pure/src/main/resources/core_functions_standard/boolean/operation/xor.pure
```diff
@@ -14,6 +14,27 @@
...
+
+**Parameters**
+- `first` — the first operand.
+- `second` — the second operand.
+
+**Returns** `true` when exactly one operand is `true`.
+
+**Examples**
+```pure
+xor(true, false)   // true
```
kevin-m-knight-gs, line 31: Might as well write out all 4 cases, instead of just 3 of 4.
```suggestion
xor(true, false)   // true
xor(false, true)   // true
```

legend-engine-core/legend-engine-core-pure/legend-engine-pure-code-functions-standard/legend-engine-pure-functions-standard-pure/src/main/resources/core_functions_standard/collection/or.pure
```diff
@@ -14,11 +14,38 @@
...
+
+**Parameters**
+- `vals` — the values.
+
+**Returns** `true` when any value is `true`; `false` for an empty collection.
+
+**Examples**
+```pure
+or([false, true])    // true
+or([false, false])   // false
```
kevin-m-knight-gs, line 34:
```suggestion
or([false, false])   // false
or([])   // false
```

legend-engine-core/legend-engine-core-pure/legend-engine-pure-code-functions-standard/legend-engine-pure-functions-standard-pure/src/main/resources/core_functions_standard/date/operation/timeBucket.pure
```diff
@@ -14,35 +14,65 @@
...
 import meta::pure::test::pct::*;
 
+'''
+Rounds a timestamp down to the start of the bucket it falls in.
+
+The way to group events into fixed intervals — five-minute bars, hourly counts, monthly totals —
+without computing boundaries by hand. The result is the **start** of the containing bucket, so all
+timestamps in the same interval collapse to the same value and can be grouped on directly.
+
+**Buckets are measured from the Unix epoch**, not from the earliest value in the data, so the same
```
kevin-m-knight-gs, line 24: We should probably give the actual date, rather than saying "Unix epoch". A lot of people won't know when that is.

legend-engine-core/legend-engine-core-pure/legend-engine-pure-code-functions-standard/legend-engine-pure-functions-standard-pure/src/main/resources/core_functions_standard/math/aggregator/mode.pure
```diff
@@ -22,11 +22,30 @@ Class <<access.private>> meta::pure::functions::math::Accumulator
    prevCount: Integer[1];   
 }
 
+'''
+The most frequently occurring value in a collection.
+
+Unlike the mean and the median, the mode is always a value that actually appears. **Where several
+values tie for most frequent, one of them is returned rather than all** — which one is decided by the
+sort the implementation performs first, so treat it as unspecified.
```
kevin-m-knight-gs, line 30:
```suggestion
values tie for most frequent, one of them is returned rather than all.** Which one is returned is unspecified.
```

legend-engine-core/legend-engine-core-pure/legend-engine-pure-code-functions-standard/legend-engine-pure-functions-standard-pure/src/main/resources/core_functions_standard/math/aggregator/sum.pure
```diff
@@ -14,16 +14,47 @@
 
 import meta::pure::test::pct::*;
 
+'''
+The total of a collection of numbers.
```
kevin-m-knight-gs, line 18:
```suggestion
The sum of a collection of numbers.
```

legend-engine-core/legend-engine-core-pure/legend-engine-pure-code-functions-standard/legend-engine-pure-functions-standard-pure/src/main/resources/core_functions_standard/math/aggregator/sum.pure
```diff
@@ -14,16 +14,47 @@
...
+
+**An empty collection gives 0, not empty**, which is why the result is `[1]` — `sum` is defined in
+terms of `plus`, whose identity is zero. That means a sum of no rows is indistinguishable from a sum
+of rows that happen to cancel; test the collection separately if the difference matters.
```
kevin-m-knight-gs, line 22:
```suggestion
terms of `plus`, whose identity is zero. That means a sum of no values is indistinguishable from a sum
of values that happen to cancel; test the collection separately if the difference matters.
```

kevin-m-knight-gs (approved):

PR #5092, opened 2026-08-18
M2M: Keep owner-reachability diagnostic inert for constant reference data

aziemchawdhary-gs:
## What

Stops `calculateSourceTree`'s owner-reachability diagnostic from rejecting M2M transforms that navigate purely constant reference data, e.g.

```pure
*test::Tgt : Pure
{
  ~src test::SrcRoot
  label : referenceData()->map(w | $w.leaf)->map(l | $l.label)->joinStrings(',')
}
```

where `referenceData()` builds instances from literals and takes no source input. Since #5083 this shape failed plan generation with:

```
No intermediate-class property of test::SrcRoot returns test::RefWrap - cannot build path back to owner.
```

## Why

The navigated classes are genuinely unreachable from the source class, but no source data flows into them, so nothing needs fetching — the legacy fallback's empty source-tree contribution is the correct output. The diagnostic exists to catch the case where an unreachable chain *does* carry source-derived state (see `testUnreachableOwnerGivesClearError`) and silently dropping it would lose data.

The property tree cannot distinguish the two shapes: the scan drops source accesses inside constructor key expressions, so `$src->badWrap().foreign.y` and `referenceData()->map(...)` both reduce to the same unreachable chains by the time the precondition runs.

## How

The distinction is read off the transform expressions directly. A new private helper `referencesAnyVariable` walks the deactivated transform expression (function arguments, instance key expressions, nested lambdas, routed wrappers) looking for a reference to the transform lambda's source parameter. The precondition gains one disjunct: it only fires if at least one of the property mappings' transforms references its source parameter.

Transforms mixing a source reference with constant navigation (e.g. `if($src.flag, |referenceData(), |[])->...`) are conservatively treated as source-referencing, so the diagnostic stays as strict as before wherever source data is involved. `docs/engineering/architecture/source-tree-calculation.md` §7.4 documents the new gate.

## Tests

New regression tests in `testSourceTreeCalc.pure` (`constantRef` package):

- `testConstantReferenceDataNavigation` — property mapped purely from a constant chain; asserts the source tree contains only the root and none of the constant-derived properties.
- `testConstantReferenceDataAlongsideSourceProperty` — constant chain alongside a normal `$src.x` property; asserts `x` is kept and the constant-derived properties are absent.

Verified: full `legend-engine-pure-code-compiled-core` suite green (`Test_Pure_Core` 1141/1141, including the pinned error-message tests `testUnreachableOwnerGivesClearError` and `testMutualCycleGivesClearError`).

### DIFF
--- a/docs/engineering/architecture/source-tree-calculation.md
+++ b/docs/engineering/architecture/source-tree-calculation.md
@@ -459,6 +459,23 @@ subtree where `srcClass != owner` — that is the legitimate cross-branch
 shape where the legacy fallback's silent empty output is correct. The
 precondition stays inert there.
 
+A second gate (`transformsReferenceSource`) handles constant reference data:
+a transform such as `referenceData()->map(w | $w.leaf)->map(l | $l.label)`
+navigates classes that are genuinely unreachable from owner, but the data is
+built by a function from literals — no source data flows into it, so nothing
+needs fetching and the legacy fallback's empty contribution is correct. The
+property tree cannot make this distinction (`$src->badWrap().foreign.y`, the
+genuine-error shape, scans to the same unreachable chains because the scan
+drops the source access inside the constructor's key expressions), so the
+gate is read off the transform expressions directly: the diagnostic fires
+only if at least one of the property mappings' transforms references its
+source parameter (`referencesAnyVariable`, a recursive walk over the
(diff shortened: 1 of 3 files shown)


rafaelbey (approved):

finos-admin:
- #5100

PR #5097, opened 2026-08-18
Express JSON navigation and array operations in the relational store language

rafaelbey:
## What this enables

A modeller can express a multi-step JSON navigation **in the relational store language** — inside a `View` column, a `Filter`, a `Join` or a property mapping — and consumers see an ordinary typed property. The acceptance test is one property mapping:

```
firstLargeCity: extractFromSemiStructured(
    array_first(array_filter(
        extractFromSemiStructured(T.FIRM_DETAILS, 'firm.divisions', 'SEMISTRUCTURED[]'),
        d | extractFromSemiStructured($d, 'headcount', 'INTEGER') > 100)),
    'address.city', 'VARCHAR')
```

Navigate, filter a collection, take the first survivor, navigate further into it. `Variant` appears nowhere in the models or mappings — JSON is typed through an ExternalFormat `Binding`.

## Root cause worth highlighting

One 10-entry scalar whitelist in `dbExtension.pure` was blocking three separately-recorded failures at once: the whole `array_*` family (only `array_size` worked), `[*]` bindings that needed `parseJson` to recover array-ness, and a Snowflake `GET(VARCHAR…)` error whose recorded reason had misdiagnosed it as a union-router bug. Letting an extraction name an array element type (`'INTEGER[]'`) and a document (`'SEMISTRUCTURED'`) fixed all three and retired four expected-failure records.

## Lambdas reuse what already existed

`array_filter` / `array_transform` / `array_reduce` in the store language compile to the **existing** `FilterRelationalLambda` / `MapRelationalLambda` / `FoldRelationalLambda` nodes that the Pure path already produced. **No SQL-generation code was written** and no dyna-registry entries were added — only the authoring path (grammar, walker, protocol, compiler, composer). A parallel node class would also have been silently skipped by `reAliasQuery` and milestoning.

Syntax: `array_filter(<collection>, d | <body>)` with `$d` in the body; `array_reduce(<collection>, <seed>, (acc, x | <body>))`.

## Verification

| suite | result |
|---|---|
| DuckDB semi-structured | 154 / 154, **0 exclusions** |
| Snowflake semi-structured | 154 / 154, **0 exclusions** (was 2) † |
| Databricks semi-structured | 154 / 154, exclusions 8 → 4 † |
| H2 semi-structured | 154 / 154 (skips; H2 is on a demise path) |
| DuckDB PCT | 1469 |
| Relational EMIT | 389 |
| `Test_Pure_Relational` | 2734 |

† The two cloud figures were measured before the rebase onto 4.140.1-SNAPSHOT. Both suites are expected to be unchanged, but they have not been re-measured locally since; the cloud CI run is the confirmation.

Expected-failure entries now **fail when they stop reproducing**, which immediately caught two stale Databricks records — one of which was masking a wrong *answer* rather than an error.

## Known gaps, deliberately not addressed here

- **`explodeSemiStructured`**: Join-only reachability; at most one unique explode per join; the operand must be a bare `TableAliasColumn`, so nested explosion is impossible. Documented in the architecture doc; next piece of work.
- **No compile-time validation** of dyna name, arity, path or return type — all still fail at SQL generation. Section 10 of the doc proposes the seam and its warn-first rollout.
- **Explode's type argument** reads a different vocabulary than `extractFromSemiStructured`, and an unrecognised name silently yields a document. Left alone on purpose: unlike an unknown dyna name, an unknown type argument *succeeds* today, so rejecting it would break working models on upgrade. Belongs in the staged rollout.
- **Postgres / MemSQL / ClickHouse** have no array branch — they sit outside the parameterised suite, so the code would be unverifiable.
- **Engine-first**: the upstream `legend-pure-m2-store-relational-grammar` parser does not accept the lambda syntax, following the `TabularFunction` precedent.

## Docs

`docs/engineering/architecture/relational-dynafunctions.md` gains sections 11.8–11.12: the limitations found by building the models, array-typed extraction, store-language lambdas, which `array_*` functions actually work, and the explode type-argument divergence.

### DIFF
--- a/legend-engine-xts-relationalStore/legend-engine-xt-relationalStore-generation/legend-engine-xt-relationalStore-pure/legend-engine-xt-relationalStore-core-pure/src/main/resources/core_relational/relational/pureToSQLQuery/pureToSQLQuery_variant.pure
+++ b/legend-engine-xts-relationalStore/legend-engine-xt-relationalStore-generation/legend-engine-xt-relationalStore-pure/legend-engine-xt-relationalStore-core-pure/src/main/resources/core_relational/relational/pureToSQLQuery/pureToSQLQuery_variant.pure
@@ -463,7 +463,10 @@ function meta::relational::functions::pureToSqlQuery::processVariantInstanceOf(f
   let keyProp = $variantToMetadata.typeName->orElse('_type');
   let className = $variantToMetadata.typeLookup->filter(p | $p.second == $class)->head().first->orElse($class.name->toOne());
 
-  let typeProp = ^SemiStructuredPropertyAccess(operand = $variant, property = ^Literal(value = $keyProp), returnType = String);
+  // avoidCastIfPrimitive: the navigation is immediately wrapped in its own variantTo conversion
+  // below, so the renderer must not also cast. Dialects whose unquote is not idempotent (DuckDB
+  // renders variantTo on VARCHAR as ->>'$', which requires JSON input) fail on the double cast.
+  let typeProp = ^SemiStructuredPropertyAccess(operand = $variant, property = ^Literal(value = $keyProp), returnType = String, avoidCastIfPrimitive = true);
   let typePropCasted = newDynaFunction('variantTo', [$typeProp, ^DataTypeInfo(dataType = ^meta::relational::metamodel::datatype::Varchar(size=100))]);
   let instanceOf = newDynaFunction('equal', [$typePropCasted, ^Literal(value = $className)]);
 
--- a/legend-engine-xts-relationalStore/legend-engine-xt-relationalStore-generation/legend-engine-xt-relationalStore-pure/legend-engine-xt-relationalStore-core-pure/src/main/resources/core_relational/relational/pureToSQLQuery/pureToSQLQuery_variant.pure
+++ b/legend-engine-xts-relationalStore/legend-engine-xt-relationalStore-generation/legend-engine-xt-relationalStore-pure/legend-engine-xt-relationalStore-core-pure/src/main/resources/core_relational/relational/pureToSQLQuery/pureToSQLQuery_variant.pure
@@ -1048,6 +1048,23 @@ function meta::relational::functions::pureToSqlQuery::extractExplodeSemiStructur
     ]);
 }
 
+// Innermost first, so index 0 is the explode whose input must be a column and each later entry
+// fans out an array held inside the elements the previous one produced.
+function meta::relational::functions::pureToSqlQuery::explodeSemiStructuredChain(op: DynaFunction[1]):DynaFunction[*]
+{
+  let inner = $op.parameters->at(0);
+  if($inner->instanceOf(DynaFunction) && $inner->cast(@DynaFunction).name == meta::relational::functions::sqlQueryToString::DynaFunctionRegistry.explodeSemiStructured.name,
+     | $inner->cast(@DynaFunction)->explodeSemiStructuredChain()->concatenate($op),
+     | $op);
+}
+
+function meta::relational::functions::pureToSqlQuery::buildSemiStructuredNavigationForExplode(path: String[1], operand: RelationalOperationElement[1]):RelationalOperationElement[1]
+{
+  let pathNavigation = meta::relational::functions::sqlQueryToString::parseSemiStructuredPathNavigation($path);
+  let initialNavigation = if($pathNavigation->at(0)->isDigit(), | ^SemiStructuredArrayElementAccess(operand=$operand,index=^Literal(value = $pathNavigation->at(0))), |  ^SemiStructuredPropertyAccess(property=^Literal(value = $pathNavigation->at(0)->substring(1, $pathNavigation->at(0)->length()-1)), operand=$operand));
+  $pathNavigation->slice(1, $pathNavigation->size())->fold({property, navigation | if($property->isDigit(), | ^SemiStructuredArrayElementAccess(operand=$navigation,index=^Literal(value = $property)), |  ^SemiStructuredPropertyAccess(property=^Literal(value = $property->substring(1, $property->length()-1)), operand=$navigation)) }, $initialNavigation);
+}
+
 function meta::relational::functions::pureToSqlQuery::explodeReturnType(typeName:String[1]):Type[0..1]
 {
     let map = meta::relational::functions::database::sqlTextToRelationalDataTypeMap();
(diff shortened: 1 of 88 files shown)


legend-engine-xts-relationalStore/legend-engine-xt-relationalStore-generation/legend-engine-xt-relationalStore-pure/legend-engine-xt-relationalStore-core-pure/src/main/resources/core_relational/relational/tests/semistructured/model/semiStructuredUnionMapping.legend
```diff
@@ -83,7 +83,7 @@ Mapping test::mapping::testMapping
       [test::store::testDB]PERSON_SCHEMA.PERSON_TABLE_1.ID
     )
     ~mainTable [test::store::testDB]PERSON_SCHEMA.PERSON_TABLE_1
-    firm: Binding test::binding::FirmBinding : extractFromSemiStructured([test::store::testDB]PERSON_SCHEMA.PERSON_TABLE_1.FIRM_DETAILS, 'firm', 'VARCHAR')
+    firm: Binding test::binding::FirmBinding : extractFromSemiStructured([test::store::testDB]PERSON_SCHEMA.PERSON_TABLE_1.FIRM_DETAILS, 'firm', 'SEMISTRUCTURED')
```
rafaelbey, line 86: We should not change an existing test - if the platform works with varchar, we need to ensure it still is after this change.

legend-engine-xts-relationalStore/legend-engine-xt-relationalStore-generation/legend-engine-xt-relationalStore-pure/legend-engine-xt-relationalStore-core-pure/src/main/resources/core_relational/relational/pureToSQLQuery/pureToSQLQuery_variant.pure
```diff
@@ -298,17 +298,23 @@ function meta::relational::functions::pureToSqlQuery::processVariantConcatenate(
 
 function meta::relational::functions::pureToSqlQuery::processVariantDrop(f:FunctionExpression[1], currentPropertyMapping:PropertyMapping[*], operation:SelectWithCursor[1], vars:Map<VariableExpression, ValueSpecification>[1], state:State[1], joinType:JoinType[1], nodeId:String[1], aggFromMap:List<ColumnGroup>[1], context:DebugContext[1], extensions:Extension[*]):RelationalOperationElement[1]
 {
-  processDynaFunction('array_drop', {reo | $reo->init()->toVariantFriendlyValue($f.parametersValues->at(0).genericType->genericTypeToDataType($f.parametersValues->at(0).multiplicity))->concatenate($reo->last())}, $f, $currentPropertyMapping, $operation, $vars, $state, $joinType, $nodeId, $aggFromMap, $context, $extensions);
+  // Semi-structured guard: disable auto-flatten so the array stays as a scalar column
+  let effectiveState = if($f->isSemiStructuredArrayInput($vars, $state), |^$state(disableAutoFlatten = true), |$state);
```
rafaelbey: we should remove the changes to pure to sql, and its associated test.  This change should be about improving the relational grammar.  Extending to pure functions should be a separate chanegset.

rafaelbey: Done — removed the `disableAutoFlatten` guard added to the ten unconditionally-routed processors (`sort`, `reverse`, `removeDuplicates`, `contains`, `slice`, `take`, `drop`, `last`, `init`, `tail`), together with the two EMIT suites that asserted the new results (`orderingSuite`, `removeDuplicatesSuite`).

Tracked as #5114, with the measurements, so the follow-up has a home. §11.8 item 2 of the dyna-function doc now records it as an open defect rather than a shipped fix.

Two things in that file are staying, and I would rather flag them than have them read as an oversight:

- `explodeSemiStructuredChain` / `buildSemiStructuredNavigationForExplode` implement `explodeSemiStructured` written in a `Join` — store language, not a Pure function. They live in this file only because `extractExplodeSemiStructured` and `explodeReturnType` already do.
- `avoidCastIfPrimitive` on the `instanceOf` discriminator is part of wiring the shared semi-structured suite to DuckDB at all. Without it DuckDB double-casts — `variantTo` on a VARCHAR renders as `->>'$'`, which needs JSON input — and the suite fails.

Happy to pull either if you disagree.

legend-engine-xts-relationalStore/legend-engine-xt-relationalStore-generation/legend-engine-xt-relationalStore-pure/legend-engine-xt-relationalStore-core-pure/src/main/resources/core_relational/relational/tests/semistructured/model/semiStructuredUnionMapping.legend
rafaelbey: Pushing back on this one, and I think the entries that retired are the argument.

`'VARCHAR'` here tells the extraction to unquote a sub-document into text, and a `Binding` is then attached to that text. On any store that distinguishes a document from a string that is simply wrong — the property accesses the binding synthesises are handed a string where a document is required. That is exactly what the four expected-failure entries were recording:

- Snowflake, 2 entries: `Invalid argument types for function 'GET': (VARCHAR(134217728), VARCHAR(8))`
- Databricks, 2 entries: silently returns `firm_D` alone, and `firm_D/firm_E/firm_F` plus three nulls

It passed on H2 only because H2 has no semi-structured type at all — it holds the document in a `VARCHAR`, so the unquote is a no-op there and the defect is invisible. The model was encoding H2's representation as the contract, and the two cloud dialects were saying so. `'SEMISTRUCTURED'` is the correction, and all four entries retire.

Your point about not losing what worked still stands, though: H2 went from passing to skipped, and that was a real regression rather than an acceptable trade. It turned out not to be structural.

H2 renders `extractFromSemiStructured` in three places, and each ends its cast-type ladder with a bare `$returnType` fallthrough, so an unrecognised type name goes into the SQL verbatim as `cast(... as semistructured)`. H2 already renders the `SemiStructured` *datatype* as `VARCHAR(n)` in the same files, so mapping the type name the same way makes the cast the no-op it should be. The path this test actually takes is `h2SqlDialect.pure` (the `sqlDialectTranslation` one); the two legacy renderers get the identical arm for consistency, and only the first is covered by a test.

The two H2 skips are gone and H2 runs the corrected model: H2 semi-structured 156/156 with both union tests executing, DuckDB 156/156, relational EMIT 427/427, relational grammar 300 with no failures.

gs-rpant1729 (approved):

PR #5104, opened 2026-08-19
DataQuality: fix relation validation plan generation broken by ordered aggregates

rafaelbey:
## Problem

Plan generation for `DataQualityRelationValidation` assertions has been failing since #5091 with

```
Unable to resolve type param Z from parameter values
```

surfacing at `platform/pure/essential/tests/assert.pure:26`. Every `rowsWithEmptyColumn` /
`rowsWithNonEmptyColumn` validation is affected — including everything the DataQuality
rule-suggestion engine emits — so this is not limited to an exotic query shape.

Reported against:

```pure
$rel->filter(row|$row.col1->isNotEmpty())
    ->join(#>{db.table}#->select(~[col1])->rename(~col1, ~joined_col1), JoinKind.LEFT, {r1, r2 | $r1.col1 == $r2.joined_col1})
    ->rowsWithEmptyColumn(~joined_col1)
    ->assertRelationEmpty(~[col1, col2])
```

The plain `rowsWithEmptyColumn(~col)->assertRelationEmpty(...)` shape fails identically; the join is
not required to trigger it.

## Root cause — two stacked defects

**1. A stereotype that #5091 made vestigial.**

#5091 added `relation::eval` to `shouldStopFunctions` so the ColSpec aggregates would reach
`processRelationEval` rather than being folded into metamodel navigation. That also made
`<<functionType.NormalizeRequiredFunction>>` on `eval` dead weight: `shouldInline` consults
`shouldStop`, so eval's body is never inlined on any routed path. The stereotype's only remaining
effect is the normalization `prevalFunctionExpressionIfRequired` runs *before* the router stops —
`router_routing.pure:206` is that stereotype's sole consumer.

That normalization recomputes `resolveResolvedTypeParameters`, which is a no-op when the type
parameters already arrive concrete and a hard failure when they do not. `eval<Z,T>` declares `Z` only
inside `ColSpec<(?:Z)⊆T>` — a `GenericTypeOperation`, which has neither `typeParameter` nor `rawType`,
and `resolveResolvedTypeParameters` matches only a plain `typeParameter` or a function return type.

A literal colspec survives because the engine compiler stamps a concrete `Z` (`EvalInference`), which
is why `eval`'s own PCT test is green. A colspec **forwarded from an enclosing function's parameter**
— exactly what the DataQuality helpers do — does not.

**2. A latent DataQuality bug that #5091 exposed.**

`rowsWithEmptyColumn` / `rowsWithNonEmptyColumn` are declared `<T,Z>` over `ColSpec<(?:Z)⊆T>`, where
`Z` is the column's *value* type. Their handler stamped the ColSpec's own `RelationType` as `Z`.

Inert while eval was folded away — the folded `Column` carried the real type. Once eval is stopped,
the declared return type `Z[0..1]` is used as-is and a relation reaches `processEmpty`, which matches
only `DataType` or `Class`:

```
Match failure: RelationTypeObject instanceOf RelationType
```

## Fix

- Drop the vestigial stereotype from `relation::eval`. `shouldStopFunctions` is left alone —
  #5091's `reduce` / `joinStrings` push-down still depends on it.
- Extract the column's value type in the DataQuality handler, the way
  `Handlers.getGenericReturnTypeForEvalCol` already does for `relation::eval`. The four helpers whose
  ColSpec pins a concrete column type are declared `<T>` only and are untouched.
- `resolveResolvedTypeParameters` now names the offending function. The old message reported a bare
  `Z` shared by several functions in one expression; diagnosing this required patching that line
  first, so it is kept.

## Tests

Nothing in the repository took a `DataQualityRelationValidation` as far as an `ExecutionPlan`, and no
test compiled a `rowsWith*` helper inside an assertion, so both defects were invisible to CI. The
existing coverage stops one stage short in both of its forms: `doRelationTest` compares the generated
lambda, and the helper tests in `dataquality_relation_helper_test.pure` run over an in-memory `#TDS`.
Neither reaches the router. (Also note `DataQualityValidationArtifactGenerationExtension.generate`
swallows every exception, and `legend-engine-xts-dataquality` is absent from `modulesToTest.json`.)

Two layers are added, at the depth each is good for:

- **Execution**, in `TestDataQualityApi`, reusing the `ResourceTestRule` and result-parsing idiom that
  class already uses. The validations run against the seeded H2 database and the defect rows are
  asserted by exact set equality — `emailPresent` returns exactly `{Eve Davis}` (the one seeded NULL
  email), `personHasSampleRow` exactly `{Smith Failsson}` (the one person id absent from
  `SAMPLE_VAL_TBL`). A join that degraded to an inner join, or a null check applied to the wrong side,
  changes the answer rather than only the SQL text.
- **Plan generation**, in `TestDataQualityRelationValidationPlanGeneration`, asserting only that a plan
  is produced — which is precisely what regressed — over the plain shape, the reported
  filter/join/rename shape, and a concrete-`Z` helper.

### Known gap, not from this change

The `rowsWith*` helpers do not work on a **class mapped** query
(`Class.all()->project(...)->from(mapping, runtime)`); plan generation fails in the router with
`Cannot cast a collection of size 0 to multiplicity [1]` at `router_routing.pure:582`. This is not
caused by #5091: it reproduces with `eval` removed from `shouldStopFunctions` again, and
`rowsWithValueOutsideRange` — declared `<T>` only, so untouched by the type parameter fix here — fails
identically, while a plain `filter` validation on the same model executes fine. Pre-existing gap,
worth its own issue. The execution tests here therefore use a relation query, which is the shape this
PR fixes.

## Verification

| suite | result |
|---|---|
| H2 PCT — incl. `eval::testSimpleEval` and all 8 `joinStrings` cases | 1327 |
| DuckDB PCT | 1327 |
| javaPlatformBinding PCT | 1240 |
| `Test_Pure_Dataquality` | 93 |
| engine compiler (incl. `TestHandlersValidation`) | 461 |
| dataquality compiler / generation | 31 / 13 |
| checkstyle | clean |

The H2 run was checked to confirm `eval::testSimpleEval` and the eight `joinStrings` cases were
actually executed, not silently absent — they are the only plausible blast radius of removing the
stereotype.

## Follow-ups, deliberately not in this PR

- `resolveResolvedTypeParameters` remains blind to `GenericTypeOperation`. The eight quantified
  comparison functions from #5084 (`Relation<Z=(?:U)>`) sit on the identical trap and survive only
  because their type parameters happen to arrive concrete. Teaching the resolver to match a
  `GenericTypeOperation` structurally is the durable fix; `preeval.pure`'s `resolveGenericType` has
  the same blind spot. It is a change to a central inference function and deserves its own PR.
- #5091 left unreachable `parameters.size() == 2` arms in `MaxInference` / `MinInference` /
  `AverageInference` after those overloads were dropped before merge.
- The four `<T>`-only DataQuality helpers stamp two resolved type parameters against one declared
  type parameter; harmless via zip truncation, but worth trimming.

https://claude.ai/code/session_01C75aFdVbdmyp2cs5XJ4AvB


gs-jp1 (dismissed):

rafaelbey:
Filed #5106 for the class mapped query gap noted under "Known gap, not from this change" above.

gs-jp1 (approved):

PR #5105, opened 2026-08-19
Fan a to-many bound to a semi-structured array out to rows

### LINKED ISSUE #5099: Semi-structured: a [*] path mapped to a to-many property returns a stringified array instead of a collection
### What happens

A `[*]` path mapped to a to-many primitive property returns **one row holding the array's text**, rather than a collection.

```
Class wild::Row
{
  id: Integer[1];
  names: String[*];
}

names: extractFromSemiStructured([wild::store::DB]T.DOC, 'divisions[*].name', 'VARCHAR[]')
```

Against `{"divisions":[{"name":"Alpha"},{"name":"Beta"}]}`:

| query | result |
|---|---|
| `project(~[id, v: r \| $r.names])` | one row, `"[\"Alpha\",\"Beta\"]"` |
| `project(~[id, v: r \| $r.names->joinStrings(',')])` | one row, `"[Alpha, Beta]"` |

Expected: two rows, `Alpha` and `Beta`.

### Why it matters

No error is raised at compile or run time. A modeller declares `String[*]`, sees an array-shaped value in a spot check, and ships a property that is actually a single string. The rendering differs depending on how the property is queried, which makes it easier still to mistake for a working collection.

### Why it happens

Multiplicity in a relational mapping comes from **row cardinality**. `explodeSemiStructured` produces rows, so a to-many through a join works normally — `meta::relational::tests::semistructured::nested::testNestedExplode` maps `teams: Team[*]` that way and returns one row per (division, team) on DuckDB, Snowflake and Databricks.

A `[*]` path produces a **value**: one row with an array in one column. There are no rows for the to-many to range over, so the array is handed back as a single value.

### What would be reasonable

Either of these would be an improvement over the current silence:

1. **Reject it at compile time** — a to-many property mapped to an array-valued expression is not expressible today, and saying so is better than a wrong answer.
2. **Support it** — fan the array out to rows, so `[*]` bound to a to-many property behaves the way a modeller reading the mapping would expect. This is the outcome that was expected when the gap was found.

Option 2 is the useful one; option 1 is worth having in the meantime, since the failure is currently invisible.

### Where this came from

Found while adding `[*]` wildcard path support to the relational store language (#5097). The wildcard work itself is complete and tested — this is specifically about binding the result to a to-many property.

Reproduction is one property mapping; the models in `legend-engine-xt-relationalStore-emit/src/test/resources/relational-emit-models/relational-semistructured-wildcard` and `core_relational/relational/tests/semistructured/model/wildcardPathInStoreLanguage.legend` provide the surrounding setup.

rafaelbey:
> Built on #5097, which has since **merged** (`e1ca2210edb`). This branch is now rebased directly
> onto `master`, so the diff is only this PR's own work: **2 commits, 13 files**.

Fixes #5099.

## The problem

A `[*]` path bound to a to-many property returned one row holding the array's text, with no error
at compile or run time:

```
Class wild::Row { id: Integer[1]; names: String[*]; }

names: extractFromSemiStructured([wild::store::DB]T.DOC, 'divisions[*].name', 'VARCHAR[]')
```

Against `{"divisions":[{"name":"Alpha"},{"name":"Beta"}]}` that gave one row, `"[\"Alpha\",\"Beta\"]"`.
Relational multiplicity comes from row cardinality; a `[*]` path produces a *value*, so the to-many
had nothing to range over. A modeller declared `String[*]`, saw an array-shaped value in a spot
check, and shipped a property that was actually a single string.

## What it does now

Fans the array out to one row per element, so it behaves like a to-many reached through an exploding
join. The issue offered rejecting it at compile time as the cheaper alternative; that turned out
unnecessary, because the machinery already existed.

**The Binding-backed path already answers this question.** `processSemiStructuredRelationalPropertyMapping`
gates on `hasToOneUpperBound` and builds a `SemiStructuredArrayFlatten` reached through a lateral
`JoinTreeNode`, with no store-language join involved. That block is now `buildArrayFlattenLateral`,
called from both places. No new lateral-join machinery was written, and no dialect had to learn
anything new — `SemiStructuredArrayFlatten.navigation` is a plain `RelationalOperationElement`, which
Snowflake and Databricks already render generically.

**The gate** fires only when the property is to-many and the operation is none of the shapes that are
already single-valued or already fanned out:

| Excluded | Why |
|---|---|
| `RelationalOperationElementWithJoin` | the join already supplies row cardinality |
| a bare `TableAliasColumn` | a plain to-many primitive; flattening a scalar column breaks it |
| an enum transformer | read **pre**-pushdown; `buildPossibleEnumMappingPushDown` clears `transformer` and rewrites the op to a `case`, so reading the post-pushdown list would be a no-op |
| a union source | `processRelationalOperationElementOfPropertyMapping` returns a deliberately bogus column type there |
| `state.disableAutoFlatten` | the existing suppression flag |

No dyna-name list and no array-valued classification — the shape exclusions carry it, so nothing has
to be kept in sync as `array_*` grows.

The bare-column exclusion is not theoretical: **removing it makes
`testSimpleProjectWithJoinInMappingWithFunction` fail.** I checked by deleting the conjunct and
re-running the suite. The enum exclusion is defensive only — the transpose tests are `<<test.ToFix>>`
and do not run, so nothing exercises it.

**No left-join-back on primary keys** is needed, unlike `applyJoinWithExplodeInCondition`, whose
lateral chain is hard-coded `INNER`. Every dialect's flatten is outer by construction (`outer => true`
on Snowflake, `explode_outer` on Databricks, the `[NULL]` singleton on DuckDB), so a row whose array
is empty or absent survives with a null.

## One DuckDB dialect fix

`SemiStructuredArrayFlattenOutput` reads `VALUE` back through a cast that assumes JSON — what
Snowflake and Databricks hold in their arrays. A DuckDB list may instead hold already-typed values:
an array extraction casts to a typed list, and `array_transform` renders as `apply()` whose body
yields plain text. So `->>'$'` met `Alpha` and raised `Malformed JSON`.

The value is now wrapped in `to_json` for exactly the return types whose cast reads it as JSON
(`String`, `Enumeration`); `to_json` is idempotent, so the Binding path is unchanged. Wrapping the
whole *list* instead — the obvious fix, and the one I tried first — corrupts the untyped branch:
`map_keys` through a flatten comes back `"a"` rather than `a`, caught by
`testVariantMapColumn_keys_LateralFlatten`.

## Verification

Re-run in full after rebasing onto current `master` — 72 commits of drift, including the
CI-friendly-versions migration (#5177) and a legend-pure bump to 5.97.1:

| Suite | Result |
|---|---|
| **DuckDB** PCT module | **1498 / 1498** (incl. 163 semi-structured) |
| **H2** PCT module | **1498 / 1498** (incl. 163 semi-structured) |
| `Test_Pure_Relational` | **2749 / 2749** |
| Relational EMIT | **449 / 449**, 46 models |
| Checkstyle `check@verify` | clean |

Before the rebase, all 17 PCT modules were run against the old base. Fourteen green; three failed —
**deephaven** (8 F, 617 E), **postgres** (130 F, 3 E), **clickhouse** (4 F) — and all three are
**pre-existing**: re-running each with this change stashed gave byte-identical counts. Consistent
with the mechanism too — Postgres and MemSQL both `fail('Flatten is not supported')`, and MemSQL
passes because nothing in its suite reaches a flatten; deephaven is not a relational store.

**Snowflake and Databricks were not run** — no credentials locally, so their PCT modules execute zero
tests. The seven new tests are `paramTest`s and will run against both in CI. Their arrays hold
VARIANT and their flatten operators take the navigation generically, so the DuckDB cast problem
should not arise there, but that is reasoning rather than measurement.

### New tests

Seven, on the existing wildcard fixture (which already has a firm with an empty team list and one
with no divisions, so the empty and absent cases come free): the plain projection (the issue's
acceptance criterion), an integer element type, a non-wildcard array path, a two-level wildcard
(`array_flatten` over `array_transform`), `joinStrings`, `count`, and a filter on the fanned-out
property. Plus a new EMIT model, `relational-semistructured-array-to-many`, with five suites.

H2 skips all seven: it reaches the flatten through `findTableForColumnInAlias`, which cannot resolve
a single table alias column through a transform, and through the `sqlDialectTranslation` path, which
has no case for the extraction node. Both callers are H2-only.

## Known gaps, deliberately not addressed

- **`->size()` over a *primitive* to-many** emits `count(...)` with no `GROUP BY` and fails. This is
  **pre-existing, not introduced here**: adding `otherNames->size()` to the Binding path's model
  fails identically with this change reverted. `addresses->size()` works only because a `Class[*]`
  qualifies for `isSemiStructuredArrayExpression` and takes the `disableAutoFlatten` route
  (`array_size`), which a `String[*]` never does. `->count()` works and is tested.
- **Union-sourced array mappings** and **to-many `Enumeration` properties** — excluded by the gate.
- **A `disableAutoFlatten` equivalent for primitives**, so `names->size()` could render `array_size`
  instead of an aggregate over a lateral. A performance refinement on a correct result; it needs
  `isSemiStructuredArrayExpression`'s `isClassType` gate relaxed, which is riskier than this issue.
- **Nested implicit explosion** (an array property whose target is itself array-valued), mirroring
  the one-root-explode restriction in `applyJoinWithExplodeInCondition`.
- **`savedFilteringOperation` is not merged** on the new path, matching the Binding path, which has
  the same omission in production. The filter test covers the shape and passes.

## Second commit

`CLAUDE.md` only, and unrelated to #5099 — split out so it can be dropped independently.

Documents build isolation against the shared `~/.m2`. Since the reactor moved to CI-friendly
versions (#5177), stamping is a single `-Drevision=` flag rather than an edit to every pom, so that
is the documented method now, with Maven 3.9's chained local repository kept as the stronger option
for when `~/.m2` must not be written to or read from at all.

### DIFF
--- a/CLAUDE.md
+++ b/CLAUDE.md
@@ -23,24 +23,41 @@ installed last wins, and the other silently runs against the wrong jars. This is
 dangerous for Pure modules, whose `.pure` sources are compiled to bytecode inside the installed
 jar — a stale jar produces failures with no connection to anything you changed.
 
-**Before starting work on a branch, stamp the reactor with a unique version:**
+**Stamp the build with `-Drevision`.** The reactor uses Maven CI-friendly versions — a single
+`<revision>` property in the root `pom.xml`, which every module inherits — so overriding it on the
+command line gives the whole build its own coordinates without editing a single file:
 
 ```bash
-mvn versions:set -DnewVersion=$(git rev-parse --abbrev-ref HEAD)-SNAPSHOT \
-    -DprocessAllModules=true -DgenerateBackupPoms=false
+mvn clean install -DskipTests -Drevision=$(git rev-parse --abbrev-ref HEAD)-SNAPSHOT
+mvn clean test -pl <module>  -Drevision=$(git rev-parse --abbrev-ref HEAD)-SNAPSHOT
 ```
 
-Every install then lands under its own coordinates and cannot collide with another session.
-To undo before opening a PR (the version bump must **not** be committed):
+Installs land under those coordinates and cannot collide with another checkout. Pass the same
+`-Drevision` on **every** command in the session, tests included — one that omits it silently builds
+and resolves the shared `4.x.y-SNAPSHOT` coordinates again. Never commit a changed `<revision>`.
+
+**When you must not write to `~/.m2` at all** — someone else's build is running concurrently, or the
+shared repository already holds jars from another branch you do not want to read — add a chained
+local repository as well. Maven 3.9+ writes to the **head** and reads through to a read-only
+**tail**:
 
 ```bash
-mvn versions:set -DnewVersion=<original-version> -DprocessAllModules=true -DgenerateBackupPoms=false
-# or, if backup poms were kept:  mvn versions:revert
+HEAD=/tmp/m2-$(git rev-parse --abbrev-ref HEAD)
+mvn clean install -DskipTests -pl <changed-modules> \
+    -Dmaven.repo.local=$HEAD -Dmaven.repo.local.tail=$HOME/.m2/repository
 ```
 
-Symptoms of a collision, when you have skipped this: tests failing in modules you never touched,
-a test count that changes between identical runs, or a failure that disappears after rebuilding
-an unrelated module. Suspect the cache before you suspect your change.
+Nothing is written to `~/.m2`, and anything you did not build resolves from it as usual, so there is
+nothing to re-download and `-o` still works. Naming only the changed modules is enough when the
+tail's engine jars match your branch; when they may not, add `-am` so every engine-internal artifact
+is built from source and only third-party dependencies come from the tail. Delete `$HEAD` when done.
+
+Do **not** isolate with `versions:set` — it rewrites every `pom.xml` in the reactor, which buries the
+real diff and can be committed by accident. `-Drevision` achieves the same thing with no edits.
+
+Symptoms of a collision, when you have skipped isolation altogether: tests failing in modules you
+never touched, a test count that changes between identical runs, or a failure that disappears after
+rebuilding an unrelated module. Suspect the cache before you suspect your change.
 
 Run the server (main: `org.finos.legend.engine.server.Server`):
 ```
--- a/docs/engineering/architecture/relational-dynafunctions.md
+++ b/docs/engineering/architecture/relational-dynafunctions.md
@@ -1504,12 +1504,60 @@ Empty inner arrays, explicit nulls and absent keys each contribute nothing, so n
 needed. Genuinely ragged data — a level that is an array in one row and a scalar in the next —
 fails at the extraction cast, consistently across dialects now that §11.13 has landed.
 
-**Binding a `[*]` path to a to-many property does not give a collection.** It returns the array's
-text as a single value, with no error at compile or run time. Relational multiplicity comes from
-row cardinality: `explodeSemiStructured` produces rows, so `teams: Team[*]` through a join behaves
-normally, while a `[*]` path produces one row with an array in one column and has nothing to range
-over. Tracked as [#5099](https://github.com/finos/legend-engine/issues/5099), with an assessment of
-what a compile-time rejection would take recorded on the issue.
+**Binding a `[*]` path to a to-many property gives a collection**, as of
+[#5099](https://github.com/finos/legend-engine/issues/5099). Relational multiplicity comes from row
+cardinality, so the router fans the array out to rows: a to-many `DataType` property whose operation
+is neither a join nor a bare column is wrapped in a `SemiStructuredArrayFlatten` reached through a
+lateral `JoinTreeNode`, which is what the Binding-backed path has always done for a to-many
+(`pureToSQLQuery_variant.pure`, gated on `hasToOneUpperBound`). Both now call one
+`buildArrayFlattenLateral`. See §11.16.
 
 H2 skips wildcard tests: a `[*]` path becomes an `array_transform`, and array lambdas reach H2
 through the `sqlDialectTranslation` path, which has no case for the lambda nodes.
+
+### 11.16 A to-many bound to an array is exploded implicitly
+
+`teams: Team[*]` through an exploding join and `names: String[*]` bound to a `[*]` path are the same
+question — where do the rows come from — answered in two places. The join answers it in the store
+language; the property mapping had no answer at all and returned the array as one value.
+
+The gate sits in `processRelationalPropertyMapping`'s `DataType` branch, and fires when the property
+is to-many and the operation is **not** one of the shapes that is already single-valued or already
+fanned out. Each exclusion earns its place:
+
+| Excluded | Why |
+|---|---|
+| `RelationalOperationElementWithJoin` | the join already supplies row cardinality |
+| a bare `TableAliasColumn` | a plain to-many primitive mapped to a column; flattening a scalar column breaks it, and `testSimpleProjectWithJoinInMappingWithFunction` proves it by failing when the exclusion is removed |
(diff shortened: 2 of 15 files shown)


legend-engine-xts-relationalStore/legend-engine-xt-relationalStore-generation/legend-engine-xt-relationalStore-pure/legend-engine-xt-relationalStore-core-pure/src/main/resources/core_relational/relational/tests/semistructured/model/semiStructuredUnionMapping.legend
```diff
@@ -83,7 +83,7 @@ Mapping test::mapping::testMapping
       [test::store::testDB]PERSON_SCHEMA.PERSON_TABLE_1.ID
     )
     ~mainTable [test::store::testDB]PERSON_SCHEMA.PERSON_TABLE_1
-    firm: Binding test::binding::FirmBinding : extractFromSemiStructured([test::store::testDB]PERSON_SCHEMA.PERSON_TABLE_1.FIRM_DETAILS, 'firm', 'VARCHAR')
+    firm: Binding test::binding::FirmBinding : extractFromSemiStructured([test::store::testDB]PERSON_SCHEMA.PERSON_TABLE_1.FIRM_DETAILS, 'firm', 'SEMISTRUCTURED')
```
rafaelbey, line 86: This looks incorrect - we should still allow binding from varchars.

rafaelbey:
Pushed two fixes for the CI failures on this branch.

**`Test Module - server` — `testGraphFetchMultiPrimitiveOnInlineChild`.** The implicit-explosion gate was a list of shapes to skip, and the list could not be made complete. A to-many primitive mapped through a function over a join —

```
authorId : toString([db]@Book_Authorship | AuthorshipTable.author_child_id)
```

— is neither a bare `TableAliasColumn` nor a `RelationalOperationElementWithJoin`, so it slipped through and was given a flatten it must not have. It surfaced far away, as `Match failure: TypedCastExpressionObject instanceOf TypedCastExpression` from the one-arm match in `h2SqlDialect`. That match is untouched — once the gate stops firing, nothing reaches it.

The gate now asks what the operation *yields*. Only two shapes yield a semi-structured array, and both were read off an instrumented run rather than guessed: a `[*]` wildcard desugars to a `MapRelationalLambda`, and a path naming the array directly arrives as `extractFromSemiStructured` with a declared type ending in `[]`.

Not routed through `inferRelationalType`: it has no entry for `extractFromSemiStructured` and returns nothing for it, so a type-keyed gate would never fire for the case this feature exists to serve. Adding an entry would also stop other callers asserting on these dynas — worth doing, but separately.

**`Test Module - databricks` — `testToManyPropertyAggregated`.** Same root cause as the Postgres one: no `groupByCat` supplied to `processJoinStringsOperation`, so it asserted the database type was unsupported. Now `array_join(collect_list(...), sep)`, which is how the `joinStrings` dyna entry already spells it.

Note the gap is wider: Databricks was using `processJoinStringsOperationWithConcatCall`, the shared helper that passes no `groupByCat` at all. **BigQuery and ClickHouse use it too** and have the same hole — they stay green only because neither runs a semi-structured suite that reaches it.

Verified locally: emit 676/0/0 (the array-to-many golden still generates its flatten), DuckDB 1491/0/0 and H2 1491/0/0 including all seven wildcard tests, server `Test_Relational_UsingPureClientTestSuite` 2310/0/0 — covering both the fixed test and `testSimpleProjectWithJoinInMappingWithFunction`, the guard the original gate was written around — and `Test_Relational_Databricks_Semistructured` 163/0/0 against the cloud endpoint.

gs-rpant1729 (approved):

PR #5158, opened 2026-09-01
Legend SQL - Relation scalar comparison `<op>` against a single-row subquery

wendy-nweje-gs:
### Problem

SQL's scalar subquery comparison — `x = (SELECT max(y) FROM t)` — had no Pure equivalent. The  simpler pattern where the
subquery stands for a single value was inexpressible. This is the most common subquery comparison in SQL.

### Approach

Adds five scalar comparison functions (`equalTo`, `greaterThan`, `greaterThanEqual`, `lessThan`, `lessThanEqual`) that compare a value against the single value in a single-column, single-row relation — the Pure equivalent of SQL's `value = (SELECT ...)` scalar subquery pattern. The relation is the subquery; the cardinality constraint is enforced at runtime (the Pure body asserts, and every SQL store raises independently).

### New functions

| Function | SQL equivalent | Description |
|----------|---------------|-------------|
| `equalTo(U[0..1], Relation<Z=(?:U)>[1]):Boolean[1]` | `value = (SELECT ...)` | Equal to the scalar subquery result |
| `greaterThan(U[0..1], Relation<Z=(?:U)>[1]):Boolean[1]` | `value > (SELECT ...)` | Greater than the scalar subquery result |
| `greaterThanEqual(U[0..1], Relation<Z=(?:U)>[1]):Boolean[1]` | `value >= (SELECT ...)` | Greater than or equal |
| `lessThan(U[0..1], Relation<Z=(?:U)>[1]):Boolean[1]` | `value < (SELECT ...)` | Less than |
| `lessThanEqual(U[0..1], Relation<Z=(?:U)>[1]):Boolean[1]` | `value <= (SELECT ...)` | Less than or equal |

These complement the quantified forms (`equalAll`, `greaterThanAny`, etc.). Where the quantified forms use `= ALL (subquery)` / `> ANY (subquery)`, these scalar forms use the simpler `= (subquery)` pattern and enforce that the relation holds **at most one row** — more than one is an error, matching SQL scalar subquery semantics.

### Semantics

- The relation must carry exactly one column of the same type as the value (compiler-enforced via `Relation<Z=(?:U)>`).
- An **empty relation** answers `false` (no value to compare against).
- A **null in the column** answers `false` (absence, not a match).
- An **empty value** (`U[0..1]` with no element) answers `false`.
- **More than one row** raises an error, as it does on every SQL store.
- Named `equalTo` rather than `equal` because Pure's `FunctionDefinitionValidator` reserves `equal`/`eq`/`is`.

#### What type of PR is this?
- Improvement

Fixes #


rafaelbey (approved):

legend-engine-xts-relationalStore/legend-engine-xt-relationalStore-generation/legend-engine-xt-relationalStore-pure/legend-engine-xt-relationalStore-core-pure/src/main/resources/core_relational/relational/sqlQueryToString/extensionDefaults.pure
```diff
@@ -193,6 +193,7 @@ function meta::relational::functions::sqlQueryToString::default::getDynaFunction
     dynaFnToSql('acos',                   $allStates,            ^ToSql(format='acos(%s)')),
     dynaFnToSql('add',                    $allStates,            ^ToSql(format='%s',  transform=getTransformForAddPlus())),
     dynaFnToSql('and',                    $allStates,            ^ToSql(format='%s', transform={p:String[*]|$p->makeString(' and ')})),
+    dynaFnToSql('scalarSubQuery',         $allStates,            ^ToSql(format='%s', transform={p:String[1] | $p->meta::relational::functions::sqlQueryToString::default::bracketSubQuery()})),
```
rafaelbey, line 196: hmm dont sure this is right... like this is not a dynafunction and we need to be careful using this mechanism to encapsulate all sql conversion as the initial purpose then is lost.  

What lead to go on this direction?

rafaelbey (changes requested):

PR #5171, opened 2026-09-04
Review fix unqualified columns in subquery

FeruzGS:
## Summary

Two independent defects in `fromPure.pure` column resolution — both surfacing as `no column found named: 'X'` on valid SQL — plus one silent-bug correction where invalid SQL was previously accepted.

 **Fix 1** Bare correlated reference `WHERE t2.x = outerCol` fails with `no column found named: 'outerCol'` 
 - Resolves against the enclosing relation 

 **Fix 2** `SELECT root.flag FROM (…) AS root` fails when the derived table's FROM also aliases `root`  
 - Resolves against the outer derived table 

## Fix 1 — unqualified column in a correlated subquery

 

Bare references worked only when qualified, because `columnByNameParts` derives its context name from `parts->init()` — empty for a single-part name — so it never consulted enclosing scopes.

 

New `SqlTransformContext.correlatedOwner`: for a single-part name the current relation doesn't own itself, find the unique **named** enclosing relation that does. `processQualifiedNameReference` then pulls the row variable for that owner from `varMap`, so column metadata and row variable stay in agreement.

 

Deliberately narrow — returns empty (existing behaviour) when the name is qualified, when the relation owns it locally (guards ordinary correlated queries from being silently rerouted), when the owner is unnamed, when more than one enclosing relation owns it, or outside relation mode.

 

## Fix 2 — alias reused across nesting levels

 

`processSubRelation` appends child scopes to `contexts` and nothing clears them, so a stale same-named context stays visible one level out. Given:

 

```sql

SELECT root."flag"

FROM (SELECT CASE WHEN b."ID" IS NULL THEN 0 ELSE 1 END AS "flag"

      FROM service."/service/service1" AS root

      LEFT OUTER JOIN service."/service/service2" AS b

        ON root."Integer" = b."Integer") AS root

```

 

the qualified `root.flag` resolved into the *inner* `root`, which does not own `flag`.

 

`context(name)` and `columnByNameParts` now consult the relation's own name before scanning children (`$[this.name](http://this.name/) == $contextName` → return `$this`).

### DIFF
--- a/legend-engine-xts-sql/legend-engine-xt-sql-e2e-tests/src/test/resources/parity-tests/structural/subqueries.yaml
+++ b/legend-engine-xts-sql/legend-engine-xt-sql-e2e-tests/src/test/resources/parity-tests/structural/subqueries.yaml
@@ -327,3 +327,57 @@ tests:
     category: "subqueries"
     expected_tds_status: ERROR
     expected_rel_status: PASS
+
+  # Correlated EXISTS reading an UNQUALIFIED outer column. 'salary' only exists on
+  # the outer relation (persons), so Postgres resolves the bare name against the
+  # enclosing query.
+  - id: subquery_exists_correlated_unqualified_outer_column
+    sql: "SELECT p.name FROM persons p WHERE EXISTS (SELECT 1 FROM orders o WHERE o.person_id = p.id AND o.amount > salary / 1000) ORDER BY 1"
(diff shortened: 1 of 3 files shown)


aziemchawdhary-gs (approved):

finos-admin:
- #5259

finos-admin:
- #5260

PR #5181, opened 2026-09-08
Add Null ordering support to Relation/TDS API

gs-rpant1729:

Fixes #


aziemchawdhary-gs (approved):

rafaelbey:
Heads-up on a BigQuery interaction this introduced. It is invisible on master today because BigQuery has no PCT target yet; #5108 adds one, and that is how it surfaced.

**BigQuery rejects an explicit null-ordering clause inside `RANGE` analytic frames.** Both directions, verbatim from BigQuery:

```
NULLS FIRST not supported with descending sort order in RANGE clauses of analytic functions.
NULLS LAST not supported with ascending sort order in RANGE clauses of analytic functions.
```

BigQuery is natively nulls-low, and the Relation API orders nulls high, so `nullsLowWithClauseSupport()` has to emit an explicit clause on essentially every ordered window. In a `ROWS` frame that is fine, and all 34 `ROWS`-frame window tests still pass. In a `RANGE` frame BigQuery refuses it outright, including when the data contains no nulls at all and the clause cannot change the result: 71 of the 79 failing tests use fixtures with no null values.

**Impact measured against `legend-integration-testing`:** 79 PCT tests, of which 76 were passing before this change.

| Function group | Tests |
|---|---|
| `over` | 30 |
| `reduce` | 30 |
| `extend` | 15 |
| `nth`, `first`, `last` | 4 |

Split 44 on `DESC NULLS FIRST` and 35 on `ASC NULLS LAST`.

**Worth saying clearly: for 3 of the 79 this change is an improvement.** The `testRange_*_WithNullValues_*` cases previously returned silently wrong row ordering on BigQuery and were recorded as such. They now fail loudly instead, which is the better failure.

I have recorded all 79 as exclusions in #5108 rather than suppressing the clause on BigQuery. Suppressing it would trade 76 loud failures for 3 silently wrong results, and the comment above `processSortItemIgnoringNullOrder` already makes that call. Recording them keeps the regression visible instead of buried, but it does mean BigQuery's published coverage drops by 76 window tests.

If it is worth recovering that coverage, the narrowest fix looks like making the sort processor frame-aware, so the clause is emitted for `ROWS` frames and plain `ORDER BY` but skipped for `RANGE` on BigQuery. That reintroduces wrong ordering for `RANGE` frames over nullable columns, so it is a real trade rather than a clear win, and it is your call rather than mine. Happy to take it either way.

PR #5196, opened 2026-09-11
M2M: fix class-mapping id resolution for optional class-typed properties

wendy-nweje-gs:
#### What type of PR is this?
- Bug Fix

## Problem
An M2M mapping that includes sub-mappings and leaves an optional (`[0..1]`) class-typed property mapped to `[]` fails to compile:

```
Compilation failed: Can't find class mapping 'datamarts_ops_..._SingleStorePaymentConfirmation'
```

That id in the error message is not something the user wrote. It's an internal placeholder the compiler generated for itself. Legacy Pure accepts the same input without any issue — this is a regression in legend-engine.

## Why it happens

When a class-typed property mapping doesn't have an explicit `[id]`, the grammar layer auto-fills one by taking the property's class name and replacing `::` with `_`. That gives us a placeholder like `datamarts_ops_..._SingleStorePaymentConfirmation`.

The compiler then does two things that combine badly:

1. **Looks up that placeholder id verbatim in the include graph.** The included sub-mapping uses a real, user-chosen id (e.g. `[PAYMENTCONFIRMATION]`), so the lookup misses.
2. **The escape hatch can't rescue it.** The existing fallback only fires when the transform's return type matches the property class. For `optionalProp: []` the transform returns `Nil`, so the fallback doesn't fire either.

Both routes closed → hard error, and the internal placeholder id is echoed back at the user.

## The fix

All changes are in `ClassMappingThirdPassBuilder.visit(PureInstanceClassMapping)` plus one small helper. No changes to grammar, protocol, plan generation, or execution.

1. **Fall back to lookup-by-class when the id is a placeholder.** If the `_targetSetImplementationId` matches the synthetic default the grammar layer would have generated (class FQN with `_`), and the by-id lookup fails, retry using `classMappingByClass` recursively across includes — the same lookup the sibling `else`-branch already uses. When that finds a target, overwrite the placeholder on the property mapping with the resolved SetImplementation's real id. Downstream consumers (router, JSON serializer, plan generator) read that id at runtime; leaving the placeholder in place would cause them to re-resolve by class and pick the wrong SetImplementation when the class is mapped in more than one place.
2. **Accept `optionalProp: []` on class-typed properties.** If the transform is empty (upper multiplicity 0) and the property is optional (lower multiplicity 0), treat it as an explicit "leave empty" — no target class mapping required. Matches legacy Pure.
3. **Better error message on the residual failure path.** When resolution genuinely can't find anything, report _property name, class FQN, and parent mapping FQN_ instead of the internal placeholder id.

## Tests

New in `TestMappingCompilationFromGrammar`:

- Ticket shape: parent includes a child that maps `Target[T]`, parent has `opt: []` on a `Target[0..1]` property → compiles.
- No child mapping at all, `opt: []` on `Target[0..1]` → compiles.
- `required: []` on a mandatory `Target[1]` → still fails, with the new error message.

Fixes #

### DIFF
--- a/legend-engine-core/legend-engine-core-base/legend-engine-core-language-pure/legend-engine-language-pure-compiler/src/main/java/org/finos/legend/engine/language/pure/compiler/toPureGraph/ClassMappingThirdPassBuilder.java
+++ b/legend-engine-core/legend-engine-core-base/legend-engine-core-language-pure/legend-engine-language-pure-compiler/src/main/java/org/finos/legend/engine/language/pure/compiler/toPureGraph/ClassMappingThirdPassBuilder.java
@@ -97,22 +97,43 @@ public class ClassMappingThirdPassBuilder implements ClassMappingVisitor<SetImpl
             if (property._genericType()._rawType()._classifierGenericType()._rawType()._name().equals("Class"))
             {
                 SetImplementation setImplementation;
+                // `optionalProp: []` on a [0..1] class-typed property is a no-op — no target mapping needed.
+                boolean unmappedOptionalPassThrough = isEmptyTransformOnOptionalClassProperty(last, property);
                 if (p._targetSetImplementationId() != null && !p._targetSetImplementationId().equals(""))
                 {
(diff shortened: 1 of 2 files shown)


aziemchawdhary-gs (approved):

finos-admin:
- #5207

PR #5198, opened 2026-09-11
DB2 PCT framework

marsah-gs:
#### What type of PR is this?
Improvement

Adds DB2 as a container-backed PCT target, and the execution-side plumbing it needs.

DB2 was already a DatabaseType with a full SQL dialect in core_relational (sqlQueryToString/dbSpecific/db2), but nothing could execute against it: no driver wrapper, no connection extension, no authentication flow, and no test connection. So the dialect had never been run against a real DB2 server, it did not appear in the published coverage report, and any regression in that generated SQL was invisible.

Four new modules under legend-engine-xt-relationalStore-dbExtension/legend-engine-xt-relationalStore-db2, modelled on ClickHouse:

-execution: DB2Manager (jdbc:db2://host:port/database), DB2Driver wrapping com.ibm.db2.jcc.DB2Driver, DB2Commands, and DB2ConnectionExtension registering all of it. Authentication is DB2StaticWithUserPasswordFlow: username and password come from two vault references under a configured base reference, so no credential is held in a connection definition. IBM now publishes com.ibm.db2:jcc to Maven, so the driver is a plain runtime dependency pinned in the root POM rather than a jar each deployment sources itself.

-PC: core_relational_db2_pct repository with its own adapter function (testAdapterForRelationalWithDB2Execution), kept in-module so the shared pct_relational.pure that every other adapter compiles against is untouched. Seven scope suites: essential, standard, grammar, relation, unclassified, variant, scenario-quant, plus the report provider that puts DB2 on the coverage page.
DB2TestConnectionIntegration starts DB2 in Testcontainers and registers its credentials into the test vault, so this runs in the default profile with no secrets unlike the cloud targets. CustomDB2Container carries what the DB2 image actually requires and Testcontainers does not give it by default: privileged mode, private IPC and UTS namespaces, 2 GB shared memory, a fixed hostname, and license acceptance. Startup is retried 3x with a 240 s timeout, and the registry is overridable via legend.engine.testcontainer.registry as the other container-backed adapters do. Added to modulesToTest.json as a new db2 CI group.

-pure: core_relational_db2 repository holding the dialect's sqlQueryToString tests, and the compiled suite that runs them.

-connection: the RelationalDatabaseManager service registration, matching the shape of the Postgres, ClickHouse and MemSQL connection modules.

Fixes #

This resolves no PCT tests. It is the framework only. Every test in all seven scopes is recorded as an exclusion carrying the error the runner reported, so the baseline is the honest current state of the dialect and any subsequent change to it shows up as a diff in these manifests rather than as a silent pass or failure.

No. DB2 execution is reachable only from the PCT module; the new modules are not registered in the extension collections or the server distribution, so no deployed engine gains a DB2 connection from this change.


legend-engine-xts-relationalStore/legend-engine-xt-relationalStore-dbExtension/legend-engine-xt-relationalStore-db2/legend-engine-xt-relationalStore-db2-PCT/src/main/java/org/finos/legend/engine/plan/execution/stores/relational/test/db2/integration/CustomDB2Container.java
```diff
@@ -0,0 +1,44 @@
...
+    public CustomDB2Container()
+    {
+        super(DockerImageName.parse(System.getProperty("legend.engine.testcontainer.registry", DEFAULT_REGISTRY) + IMAGE).asCompatibleSubstituteFor(CANONICAL_IMAGE_NAME));
+        this.acceptLicense();
+        // the image remounts /database and starts the instance at boot, both of which need full privileges
+        this.withPrivilegedMode(true);
+        this.withCreateContainerCmdModifier(cmd ->
+        {
+            // a host UTS namespace, which testing defaulted to, rejects a container host name outright
+            cmd.withHostConfig((cmd.getHostConfig() == null ? new HostConfig() : cmd.getHostConfig()).withUtSMode("private").withIpcMode("private").withShmSize(SHM_SIZE));
```
rafaelbey: we might need to do this only when the container env is podmand rather than docker.  For example

```
  private static boolean isPodman()                                                                                                                                                                              
  {                                                                                                                                                                                                              
      List<VersionComponent> components = DockerClientFactory.instance().client().versionCmd().exec().getComponents();                                                                                           
      return components != null && components.stream().anyMatch(c -> c.getName() != null && c.getName().startsWith("Podman"));                                                                                   
  }
```

PR #5204, opened 2026-09-14
Memoize milestoning alias walk to avoid exponential blowup on shared TableAlias subtrees

aziemchawdhary-gs:
## What

  Rewrites `meta::relational::milestoning::getQueryTableRelationalOperationElements` (the walk behind `getQueryRelationAliases` / `getQueryTableAliases`) so that each distinct `TableAlias` is expanded at most once per traversal.
  
  The walk now threads a `visited: TableAlias[*]` set through the recursion via a new private `AliasWalkState` class (`visited` + `found`). When a `TableAlias` that has already been visited is reached again by a different path, the walk short-circuits instead of re-expanding its subtree. The public entry points
  keep their signatures and simply return `.found`.

  ## Why

  A `SelectSQLQuery`'s `.data` and `.filteringOperation` subtrees routinely share the same `TableAlias` instances. Milestoning filters in particular embed `TableAliasColumn(alias=$tableAlias, ...)` nodes that reference the very alias already reachable through `.data`. The old walk had no notion of "already seen",
  so every path to a shared alias re-expanded its full subtree from scratch.

  For queries with many nested milestoned joins this made the walk exponential in nesting depth rather than linear in the number of distinct aliases, and plan generation would hang or take minutes on queries that should be fast. The walk runs on every relational plan (`generateSQLExecutionNode` calls it to
  populate `RelationMetadata.relations` on the `SQLExecutionNode`), so any sufficiently nested milestoned query hit this.

  ## Notes

  - Behaviourally, the only difference is that a shared alias now appears once in the result rather than once per path. The sole caller already de-duplicates via `removeDuplicatesBy(uniqueRelationName)`, so the emitted plan is unchanged.
  - `AliasWalkState` exists because `Pair<U,V>` type parameters can't carry a `[*]` multiplicity, so a `Pair<TableAlias[*], ...>` isn't expressible.
  - The visited check uses `==`, which is identity comparison for `TableAlias`, so it is cheap.
  - No new tests: existing milestoning and relational plan tests cover the traversal; this change is purely a complexity fix with identical output after de-duplication.


gs-jp1 (approved):

finos-admin:
- #5205

finos-admin:
- #5206

PR #5211, opened 2026-09-14
Add enumeration-mapping introspection helpers with Java-gen support

Rakan1837:
### Adds five Pure helpers under meta::pure::mapping for working with EnumerationMappings from Pure code:

  - getSourceEnumValues(Mapping[1], Enumeration<T>[1]): T[*]
  - getSourceEnumValuesByName(Mapping[1], Enumeration<T>[1], String[1]): T[*]
  - getEnumerationMappingValueByName(Mapping[1], Enumeration<T>[1], String[1], Any[1]): T[0..1]
  - mapToEnumValue(Enum[1], Enumeration<T>[1], Mapping[1]): T[1]
  - mapToEnumValueOptional(Enum[0..1], Enumeration<T>[1], Mapping[1]): T[0..1]

 All five walk the Mapping.includes DAG transitively (unbounded depth), so a child mapping sees EnumerationMappings contributed by any ancestor.

  ### Design

  A plain Pure implementation only works in interpreted mode. In compiled mode, Mapping can't cross the processInstanceValue boundary (Unhandled value type: meta::pure::mapping::Mapping), so each helper has two mirrored
  implementations:

  - Interpreted: single-expression bodies in enumerationHelpers.pure delegating to allEnumerationMappings.
  - Compiled: inline function coders in metaLibrary.pure that resolve the include-DAG at code-gen time (the Mapping and target Enumeration must be compile-time literals) and emit a literal Java expression at the call site —
    Arrays.asList(EnumType.getEnumFromPureName("A"), …) for the list-returning helpers, or a chained equals(...) ? … : … lookup for the by-value/by-name helpers. mapToEnumValue/mapToEnumValueOptional fall back to
    EnumType.getEnumFromPureName(sourceEnum.getName()) when no mapping entry matches.

  ### Tests

  testEnumerationHelpers.pure — 12 tests: direct and included mappings, a 6-level-deep include chain (regression guard for a walk that previously stopped at depth 4), unknown name/unknown source lookups, both mapToEnumValue
  fallback paths, and the mapToEnumValueOptional empty/non-empty cases.

  ### Notes
  - Also fixes a pre-existing gap in the v1_33_0 valueSpecification transform: a missing NoSetRoutedValueSpecification match arm, which surfaced because enumValues() calls now route through NoSetRoutedValueSpecification.


erwin-jeanlouis-gs (dismissed):
Code is clean and have appropriate test coverage

legend-engine-core/legend-engine-core-pure/legend-engine-pure-code-compiled-core/src/main/resources/core/pure/protocol/v1_33_0/transfers/valueSpecification.pure
```diff
@@ -55,6 +55,7 @@ function meta::protocols::pure::v1_33_0::transformation::fromPureGraph::valueSpe
       [
          f:FunctionRoutedValueSpecification[1]|$f.value->meta::protocols::pure::v1_33_0::transformation::fromPureGraph::valueSpecification::transformValueSpecification($inScope, $open, false, $extensions),
          e:ExtendedRoutedValueSpecification[1]|$e.value->meta::protocols::pure::v1_33_0::transformation::fromPureGraph::valueSpecification::transformValueSpecification($inScope, $open, false, $extensions),
+         nsrvs:NoSetRoutedValueSpecification[1]|$nsrvs.value->meta::protocols::pure::v1_33_0::transformation::fromPureGraph::valueSpecification::transformValueSpecification($inScope, $open, false, $extensions),
```
aziemchawdhary-gs, line 58: if we are adding this to v1_33 protocol then I think we need to add this to vX_X_X protocol as well

aziemchawdhary-gs (changes requested):
Can we add some M2M tests as well to exercise these new functions? Similar to tests we have in simpleObject.pure

aziemchawdhary-gs (approved):

PR #5232, opened 2026-09-18
Support Lakehouse Snowflake connections in SnowflakeApp/UDTF

ahallgs:
#### What type of PR is this?
Improvement

Support Lakehouse Snowflake connections in SnowflakeApp/UDTF, resolving classCast error

Fixes #


rafaelbey:
Im not sure why the identity is required to unblock the new behavior.  It concerns me that artifact generation, which is an SDLC activity now expect to use it to get different behaviors.

